Issue #18, opened 2018-10-02
Simplify differential expression analysis


jaclyn-taroni:
As noted in https://github.com/AlexsLemonade/refinebio-examples/pull/6#discussion_r219489521, the experiment selected for differential expression analysis (same used one in `compare-processing`) has two factors: time point and genotype. These two factors are not taken into account in the differential expression example. We should pick a "simpler" experiment (e.g., single factor like treatment) for our differential expression analysis example.

cansav09:
Closing issue. This was addressed by PR #22

PR #22, opened 2018-11-21
Used simpler experiment for differential expression

cansav09:
As a response to #18, here I've replaced the old experimental file, E-GEOD-39842, with a new experimental file, GSE71270. GSE71270 that is a CREB vs Control experiment, n = 10. I think this addresses what we were looking to do. 

I accidentally sent a pull request that was messed up because I didn't reset the local branch to the master.


differential-expression/gene_DE.Rmd
```diff
@@ -35,10 +35,10 @@ add them as data.frames to your environment.
 
 ```{r Import data from .tsv files}
 # Read in metadata tsv file
-metadata <- read.delim("../data/E-GEOD-39842_metadata.tsv", header = TRUE)
+metadata <- read.delim("../data/metadata_GSE71270.tsv", header = TRUE)
 
 # Read in data tsv file
-df <- read.delim("../data/E-GEOD-39842.tsv", header = TRUE, row.names = 1,
+df <- read.delim("../data/GSE71270.tsv", header = TRUE, row.names = 1,
```
jaclyn-taroni: Would probably change both of these to use `file.path`

cansav09: Will do.

cansav09:
Yeah, that's a good point. I will change it to `model.matrix` Better to be consistent with what limma recommends. We don't expect this linear model to run through the origin either, so intercept is probably needed.

cansav09:
Okay. Lemme know if this is good now.

cansav09:
In Rmd, the `;` is needed when you are trying to add to an existing plot. Otherwise it will give an error that the plot doesn't exist.

jaclyn-taroni:
Ah, TIL!

PR #30, opened 2019-04-09
Batch correction example

cansav09:
Here's the ComBat example. We can figure out if this dataset will be good for the PLIER examples or others as well. 
For this example I searched for "HEK 293" search in refine.bio


batch_correction/README.md
```diff
@@ -0,0 +1,8 @@
+# refine.bio Examples: Batch Correction
```
jaclyn-taroni: I don't think this README is as intended. The text under both headers is pretty similar and the H3 header is from another example.

batch_correction/batch_correct_combat.Rmd
```diff
@@ -0,0 +1,155 @@
...
+Prep metadata to be in proper format for batch info to be used by ComBat
+
+```{r Prep metadata}
+# Metadata file name
+metadata.filename <- file.path("..", "data", "metadata_HOMO_SAPIENS.tsv")
+
+# Read in data tsv file
+metadata <- readr::read_tsv(metadata.filename)
+
+# dataset <- as.factor(stringr::word(colnames(df), sep = ":", 1))
```
jaclyn-taroni: Remnant from development?

batch_correction/batch_correct_combat.Rmd
```diff
@@ -0,0 +1,155 @@
...
+
+```{r Import refine.bio data}
+# Put name of data file
+data.filename <- file.path("..", "data", "HOMO_SAPIENS.tsv")
+
+# Unzip the file and direct it to the data folder
+unzip(paste0(data.filename, ".zip"), exdir = file.path("..", "data"))
+
+# Read in data tsv file
+df <- readr::read_tsv(data.filename) %>%
```
jaclyn-taroni: Can we set `progress = FALSE` here? Since it's such a big file, showing progress in the notebook is a lot of lines.

batch_correction/batch_correct_combat.Rmd
```diff
@@ -0,0 +1,155 @@
...
+# Only keep these in the data and metadata
+metadata <- metadata[hek293, ] %>% 
+  readr::write_tsv("filtered_metadata_HOMO_SAPIENS.tsv")
+
+# Only keep these in the data and metadata
+df <- df[, hek293] %>% 
+  readr::write_tsv("filtered_HOMO_SAPIENS.tsv")
+
+# Take a look at the information we have for these data
+metadata
```
jaclyn-taroni: In the notebook setting, preference for using `dplyr::select()` here so we drop things like age which are not informative.

jaclyn-taroni: May also be useful to include something like `table(refinebio_platform)` so folks get an idea that multiple platforms  (chips) are present in this dataset.

PR #44, opened 2019-04-17
Get KEGG pathway GMT for mouse symbols

jaclyn-taroni:
Uses [`hcop`](https://github.com/stephenturner/hcop) to do the mapping.


pathway-analysis/kegg_ortholog_mapping.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+# we need devtools in order to install the hcop package we will use to
+# do the ortholog mapping
+if (!("devtools" %in% installed.packages())) {
+  install.packages("devtools")
+}
+
+# this installs a specific version of hcop
+# we pass the commit hash to the ref argument
+devtools::install_github("stephenturner/hcop",
```
cansav09: Should there be an `if` statement for this too? I realize that this package isn't as ubiquitous as `devtools` but if they ran the `ortholog-mapping` notebook, they might have already installed it.

pathway-analysis/kegg_ortholog_mapping.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+### Conversion of KEGG pathways
+
+For each KEGG pathway we have (currently populated with human Entrez IDs),
+we need a new gene set that is comprised of mouse gene symbols.
+
+```{r}
+kegg_mouse_list <- 
+  lapply(kegg_human_list, 
+         function(x) {
```
cansav09: Is there a more descriptive variable name you could use other than `x`?

pathway-analysis/kegg_ortholog_mapping.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+Briefly, the GMT format has one pathway per line and it follows this pattern:
+
+```
+<PATHWAY_NAME>\t<PATHWAY_DESCRIPTION>\t<gene_1_in_pathway>...<gene_n_in_pathway>
+```
+
+We've lost the description information because it's removed by 
+`qusage::read.gmt`.
+The description in `r kegg_file` follows this pattern:
```
cansav09: I don't exactly understand this mini section here (Lines 200 - 208). Perhaps if I ran it I would understand what you mean?

pathway-analysis/kegg_ortholog_mapping.Rmd
```diff
@@ -0,0 +1,252 @@
...
+mouse_ortholog_df <- entrez_in_pathways_df %>% 
+  dplyr::inner_join(mouse, by = "human_entrez")
+```
+
+### 1:many mapping
+
+For 1:many mappings, we'll pick the one with the most support.
+Here, we'll only consider the _number_ of sources without any regard for _what_
+those resources are.
+Essentially, we weigh all sources equally though they likely have more or less
```
cansav09: Is there a way that someone could use the permissiveness of the criteria for sorting or nah? Or what I'm guessing realistically the answer is probably: not easily here?

pathway-analysis/kegg_ortholog_mapping.Rmd
jaclyn-taroni: I did not use an `if` here because we're installing a _specific commit_. If I didn't pass a commit hash to `ref`, I would have used an `if` statement. Running this when you have the same specific version install just tells you that it has not been updated since the reference you're passing it.

cansav09: Ah. Okay. Cool. Sounds good.

pathway-analysis/kegg_ortholog_mapping.Rmd
jaclyn-taroni: I think you would have to make a choice about _weighting_ sources which I would probably encode in a column for each specific source. So the answer is yes you could do it, but it is beyond the scope of this example.

pathway-analysis/kegg_ortholog_mapping.Rmd
jaclyn-taroni: I think the key point may be at line 232. As far as adding more context goes - how familiar would you say you are with the GMT file format?

cansav09: I see. 232 feels like the punchline here. I've only briefly worked with gmt files. Not much.

jaclyn-taroni: Do you think this would make sense if you followed the link to GMT file formats?

cansav09: I looked at those. I'll just run the thing so I can see what the object looks like.

jaclyn-taroni: You can also open the `.gmt` added here. I think that would clear things up. I would think most users would go that route? I'm trying to discern if there's additional context that would help, but I'm not sure there is.

cansav09: What does `r kegg_file`refer to? The website or an object in R?

jaclyn-taroni: Discussed in person: In the HTML file, this would render as the full path assigned to `kegg_file`. This allows me to avoid hardcoding this path in the text.

Issue #54, opened 2019-04-23
Apply general recommendations for each module


jaclyn-taroni:
From @dvenprasad:

> General Thoughts/Recommendations:
>
> - Include install for dplyr and other packages?
> - Link to GenePattern Notebook instead of the preview (am I supposed to create my own notebook and try replicate the image or is that notebook available for me to plug and play?)
> - Link notebook links to rendered notebooks
> - Add blurb recommending moving their own data in the data folder of the example
> - Add a list of required packages or expected setups in README for each examples.

And from #40:

> Add a list of packages/softwares they need to have set up to run the example notebooks.

A statement about downloading in the README (https://github.com/AlexsLemonade/refinebio-examples/pull/62#discussion_r279555058):

> To run the examples yourself, you will need to clone or download this repository < https://github.com/AlexsLemonade/refinebio-examples >

Apply these recommendations to the following modules
- [x] Clustering (@cansav09)
- [x] Differential expression (@cansav09)
- [x] Dimension reduction (@cansav09)
- [x] Ensembl ID conversion (@cansav09)
- [x] QN your own data (@jaclyn-taroni)
- [x] Ortholog mapping (@jaclyn-taroni)
- [x] Pathway analysis (@jaclyn-taroni)
- [x] Validate differential expression (@cansav09)

PR #62, opened 2019-04-29
Apply general recommendations to ortholog mapping

jaclyn-taroni:
## Purpose:

Addresses #54 for this module. I also corrected an issue with a bulleted list not rendering properly in the notebook itself.

## Pull Request Check List:

* [x] Double check your paths


ortholog-mapping/README.md
```diff
@@ -1,11 +1,41 @@
...
+You can do so by clicking `Tools` > `Global Options` > `Packages` and selecting a CRAN mirror near you with the `Change` button.
+
+#### Interacting with R Notebooks
+
+You can open the R Notebook by opening the `.Rmd` file in RStudio.
+Note that working with R Notebooks requires certain R packages, but RStudio should prompt you to download them the first time you open one.
+This will allow you to modify and run the R code chunks.
+Chunks that have already been included in an example can be run by clicking the green play button in the top right corner of the chunk or by using **Ctrl + Shift + Enter** (**Cmd + Shift + Enter** on a Mac).
+See [this guide using to R Notebooks](https://bookdown.org/yihui/rmarkdown/notebook.html#using-notebooks) for more information about inserting and executing code chunks.
+
```
dvenprasad: Just realized that we will be linking directly to a lot of the example directories. We don't mention that they would need to clone/download the repo to start using it anywhere but the main README. It is worth mentioning that. 
Perhaps something brief like: 
>To run the examples yourself, you will need to clone or download this repository < https://github.com/AlexsLemonade/refinebio-examples >

jaclyn-taroni: Adding this to #54

jaclyn-taroni: If you like the changes in https://github.com/AlexsLemonade/refinebio-examples/pull/62/commits/39ecdb2acf13550e8f69881d0c83a58a69e5cdff, we can use that pattern elsewhere. The rationale for linking to that StackOverflow answer is because I think a bit more context for folks who are less familiar with GitHub is helpful but don't want to overload this with detail.

dvenprasad: Yes, that addition looks good to me.

PR #67, opened 2019-04-30
Updating Ensembl ID convert README

cansav09:
This PR applies the general recommendations of #54 to the Ensembl ID conversion README


jaclyn-taroni:
There is only one notebook and there are no plots for this example workflow. In #62, I tailored the language to that module to avoid any potential confusion.

cansav09:
I have changed this to get rid of mentions of the plots folder so that people don't get confused since this has no plots output.

jaclyn-taroni:
There are still references to multiple examples. You can use the text from ortholog mapping.

PR #70, opened 2019-04-30
Ortholog mapping: remove last few references to multiple experiments

jaclyn-taroni:
Found a couple things while looking at #67


cansav09:
LGTM!

PR #86, opened 2019-06-07
Add multi license to cover notebook materials

cansav09:
After discussing the [license for the training-modules repository](https://github.com/AlexsLemonade/training-modules/issues/65) with @cgreene we realized that the BSD-3 license doesn't exactly cover the content in refine.bio-examples. After polling twitter about how to license notebooks, Twitter came back with a multi-license. 

This was the example we followed for this conclusion: 
https://github.com/UF-Carpentry/website/blob/master/LICENSE.md

The CC-BY section is copy-paste the same license used for training-modules which was taken from the Greene lab's CC-BY.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -2,7 +2,7 @@ Copyright (c) 2019 Alex's Lemonade Stand Foundation
 
 * The code blocks contained within any computational notebooks are available for re-use under the BSD 3-Clause License ([see section below](#bsd-3-clause-license)).
 
-* All other content is available for re-use under the CC-BY 4.0 license ([see section](#creative-commons-attribution-4-0-international) below).
+* All other content is available for re-use under the CC-BY 4.0 license ([see section](#creative-commons-attribution-40-international) below).
 
 # BSD 3-Clause License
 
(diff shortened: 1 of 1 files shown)


cgreene:
Writing a brief blurb here, so we can reference it in other repos where we make the change. After realizing that computational notebooks contain both prose and source code, I posted [a twitter poll](https://twitter.com/GreeneScientist/status/1136706935535886336?s=20). Though the poll is not yet concluded, greater than 50% of 131 votes are in favor of CC-BY, with the rest of the votes split between BSD 3-Clause and Other / See Votes. Thus, the poll leads to CC-BY.

Victoria Stodden, an expert in the topic, [added a compelling paper to the discussion](https://twitter.com/victoriastodden/status/1136721715579326469?s=20). One was behind a paywall and, somewhat ironically, unavailable to be used by me at the time. However, [this paper](https://web.stanford.edu/~vcs/papers/ijclp-STODDEN-2009.pdf), which I could read led me to conclude that if we were going to update the licensing we should dual-license.

The example noted by @cansav09 above was how we implemented the dual license.

dhimmel:
More discussion on the conundrum at https://github.com/cognoma/cancer-data/issues/5.

Personally, I would release the entire repository under CC BY, and then also release just the code blocks under BSD. This way there is a single license for the entire repository which can simplify reuse for some users that just want to deal with a single license.

Issue #99, opened 2020-05-29
Add inline comments in notebooks where users need to replace code


dvenprasad:
Users didn't always feel confident about what they needed to replace. Users tended to skim through the comments before the code blocks and often missed what they need to do. 

Add inline comments like
`Replace path with path to your data file`
`Replace with your organism` 
So it explicit what they are expected to do. 

<img width="838" alt="Screen Shot 2020-05-29 at 9 19 26 AM" src="https://user-images.githubusercontent.com/15315514/83264039-88743800-a18d-11ea-96de-31f31e05efa7.png">

cbethell:
I can get started with this ticket 👍 I spoke to Candace and we both believe this is probably a good start for me with this repo.

PR #104, opened 2020-07-01
Add inline comments where users need to replace code - PR 1 of n

cbethell:
## Purpose:

This PR addresses issue #99 by adding inline comments in the notebooks in this repository where users would need to replace code to tailor the notebooks for use with different data files.

This PR adds inline comments to the notebooks in the `clustering`, `differential-expression`, and `ensembl-id-convert` directories. 

I looked for areas where users may need to replace code followed the format of the example on issue #99 to add inline comments. 

Please let me know if I missed any sections that should have an inline comment added or if the wording I added is not clear.

## Results
If applicable, briefly report your findings.

This PR includes updates to the results files in the mentioned directories, which upon some investigation, I believe reflect updates made to the input files (so this should be fine but correct me if I am wrong). 

## Pull Request Check List:
* [x] Run a linter
* [x] Comments and/or documentation up to date
* [x] Double check your paths


cbethell:
@dvenprasad, I am going to request your review here since you filed issue #99. 

Let me know if the changes made in this PR are in line with what you expected when you filed that issue. (I went over the notebooks a good number of times so hopefully I didn't miss a section that would require an inline comment but please let me know if I have!)

clustering/clustering_example.Rmd
```diff
@@ -96,8 +100,10 @@ at this [link](https://bioconductor.org/packages/devel/bioc/vignettes/ComplexHea
 browseVignettes("ComplexHeatmap")
 
 # Create an annotation that labels samples' groups by color
-annot <- HeatmapAnnotation(df = data.frame(Groups = rep(c("SP", "MP"), each = 2)),
-    col = list(Groups = c("SP" = "green", "MP" = "orange")))
+annot <-
+  HeatmapAnnotation(df = data.frame(Groups = rep(c("SP", "MP"), # Replace with group names relevant to your data
+                                                 each = 2)),
+                    col = list(Groups = c("SP" = "green", "MP" = "orange"))) # Be sure to change the group names here as well
```
cansavvy, line 106: May want to add a note about the colors too. Perhaps a link to color options.

clustering/clustering_example.Rmd
```diff
@@ -40,8 +40,8 @@ Create output folder.
 
 ```{r}
 # Create the plots folder if it doesn't exist
-if (!dir.exists("plots")) {
-  dir.create("plots")
+if (!dir.exists("plots")) { # Replace with path to desired output plots directory
```
cansavvy, line 43: For these kinds of things we might want declare the file paths once and assign them, then supply the assignment. I could see people forgetting to change this in both places. We probably should have done this the first time around.

differential-expression/microarray_DGE.Rmd
```diff
@@ -41,12 +41,12 @@ Create output folders.
 
 ```{r}
 # Create the results folder if it doesn't exist
-if (!dir.exists("results")) {
-  dir.create("results")
+if (!dir.exists("results")) { # Replace with path to desired output results directory
```
cansavvy, line 44: Same thing here I said above.

clustering/clustering_example.Rmd
```diff
@@ -52,10 +52,14 @@ add them as data.frames to your environment.
 
 ```{r Import data from .tsv files}
 # Read in metadata tsv file
-metadata <- readr::read_tsv(file.path("data", "metadata_GSE12955.tsv"))
+metadata <- readr::read_tsv(file.path("data", # Replace with path to your metadata file
```
cansavvy, line 55: Also debating about if we want to declare these file paths else where, but I am less certain about this one.

differential-expression/rnaseq_DGE.Rmd
```diff
@@ -70,26 +70,30 @@ Create output folders.
 
 ```{r}
 # Create the results folder if it doesn't exist
-if (!dir.exists("results")) {
-  dir.create("results")
+if (!dir.exists("results")) { # Replace with path to desired output results directory
```
cansavvy, line 73: Same thing here.

clustering/clustering_example.Rmd
cbethell: This is the only comment I did not address in the last commit because it is not clear that declaring these paths else where will be beneficial or would make a difference here as it did with the file paths to the plots/results directories. Was this also your thinking?

cbethell: Also, is your suggestion here to declare the metadata and data file paths else where or just the path to the data folder? I believe the latter could be useful but that's also assuming that users would save their metadata and other data files in the same directory.

cansavvy: I think either strategy is fine. But perhaps just declaring the folder and building the metadata and data paths from that? As long as we have a comment at each place the person would need to change it, I'm not dead set on either way.

cbethell:
> I'm still good with this, @cbethell , it's up to you if you want to declare the file paths as an object. If we want to be consistent, we can/should do that. But also maybe variety is okay?

I had the same thought re variety possibly being okay in this case so thank you for the second opinion! 

I believe since these are example notebooks, allowing them to declare file paths in the two different ways can be useful. If it becomes confusing to our users, we can make the notebooks in this PR consistent with the declare file paths as an object model, but for now I'll merge this.

Issue #118, opened 2020-07-08
Set up a navigation bar on github pages


cansavvy:
Related: https://github.com/AlexsLemonade/refinebio-examples/issues/107#issuecomment-655688144

I think the header on the top of the page would include the main technologies/ larger sections (see below).
I think we'd also include TOC on the side of the web page to allow for easy jumping around to different analyses modules for *within* that technology. 

#### Rough Outline
```
refinebio-examples
├── Getting_Started 
│   ├── What is refinebio?
│   ├── About how this tutorial book is structured
│   ├── What you need to install to run the examples
│   ├── How to follow these examples
│   ├── How to get the data
│   ├── How to use/get Rmds
│   └──A quick primer on file paths and R
├── RNA-seq
│   ├── Introduction to RNA-seq
│   ├── Gene ID conversion
│   ├── Clustering
│   └── ... rest of the analyses for RNA-seq
├── Microarray
│   ├── Introduction to Microarray
│   ├── analyses examples for microarray
│   └── ...
├── Advanced Topics
│   ├── Introduction to advanced topics
└── Compendium (eventually)
    └── ...
```

For some of these things, we might be able to borrow what Jackie figured out with Jekyll `training-specific-template` repo?
Once we get the basics done we could even prettify it so it matches refinebio aesthetics.

cansavvy:
@dvenprasad What do you think about this idea?

jaclyn-taroni:
(Josh not Jackie figured that out.)

jashapiro:
I'll just chime in that getting the two navbars is not impossible (top and side), but will probably require dev time to make some template html. I couldn't find a prebuilt jekyll theme that did it, but it should be doable, and even possible to have the side navbar reflect the index as with `nb.html` files.

jashapiro:
Oh, I should also caution that I seem to remember jekyll and .Rmd files not playing together as well as I liked. I did some odd hacking to make things go on my old website. https://github.com/jashapiro/jashapiro.github.io/blob/2f6391cd28f6eef3a79cb41450159fd104421d65/Rakefile#L63
Maybe things are better now, but another option worth looking at is `blogdown`/hugo, which can still put the rendered site on github pages as static files (and should be able to do all the templating).

jaclyn-taroni:
[...]

Issue #102, opened 2020-06-29
Organize examples by technology/use case, rather than analysis type


jaclyn-taroni:
Related to #101 and #92 

If we present things as a follow along tutorial, it makes sense to me to organize material by technology or use case rather than analysis type. Specifically, duplicating analysis _types_ like gene identifier conversion across separate microarray and RNA-seq "tracks" allows for some repetition of concepts and more opportunities to vary things in a way that supports learning (e.g., use different species [#98]). 

Here are some initial thoughts about how to organize things, where the listing of individual types of analyses are incomplete and for illustrative purposes:

* Introduction
  * Setting up RStudio
  * Obtaining files locally, etc.
  * Skip to "advanced use cases" (more on that below)
* Microarray 
  * Background on how we process microarray data (similar information to what's in the docs)
  * Clustering
  * Gene ID conversion
  * Differential gene expression analysis
  * Pathway analysis
* RNA-seq
  * Background on how we process RNA-seq data (similar information to what's in the docs)
  * Clustering
  * Gene ID conversion
  * Differential gene expression analysis
  * Pathway analysis
* Larger cross-experiment, cross-platform experiments & compendium
  * How we process these datasets in general and in the specific compendium case (similar information to what's in the docs)
  * Machine learning examples
* "Advanced usage"
  * Normalizing your own data
  * API usage examples
  * Eventually: Python Client
  * Eventually: Command line
  * Eventually: R Client

Where for the clients this may be more like a place where we link out to other documentation specific to those clients, but include some info about installation, for example.

cansavvy:
The heart of this issue has been addressed in #120

cansavvy:
Although the heart of the issue is addressed, we should keep this open for reference as well as the "Advanced topics" section part of this outline which is not yet added.

PR #119, opened 2020-07-08
Testing a restructuring by tech and jekyll things

cansavvy:
## Purpose:
My main question here is 1) will this restructure of folders generally work and 2) will it work with jekyll navi things and Github pages? 

Related to #118 and #102 

I will *not* be planning on merging this branch but instead if it seems like it will work then I will go back and make this into smaller PRs we can more carefully comb over. 
But I don't know if this will work yet/I don't know what I'm doing. ¯\_(ツ)_/¯ So I'm placing it in a more visible spot that maybe others can comment on. 

There's no TOC bar on the side. I haven't even attempted that.


jashapiro:
Suggestion: separate the restructuring from the Jekyll stuff. If you restructure first, it will make Jekyll changes much easier to isolate.

cansavvy:
We're going to go with @jashapiro 's suggestion above. Closing this now that we kinda? know its maybe reasonable.

Issue #127, opened 2020-07-09
Create clustering example with RNA-seq data module


cansavvy:
Related to #110 

Currently we have a microarray example of clustering with ComplexHeatmap. We should be able to largely copy the microarray example but switch out the dataset and the accompanying wording. 

When switching to an RNA-seq dataset we should also keep in mind whether the strategy of keeping genes with the largest variances is still appropriate (I think the answer should be yes). But perhaps in addition we'll want a minimum TPMs cutoff?

This can be done separately from the "getting started" additions to each analysis #116

cansavvy:
Here's a DESeq2 example that might be good to borrow from: http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#heatmap-of-the-count-matrix

However this would involve switching from `ComplexHeatmap` -> `pheatmap`. I'm not opposed to this because I do think `ComplexHeatmap` could be overwhelming for many users. 

We may consider making a "basic heatmap" and "advanced heatmap" where we use `pheatmap` and `ComplexHeatmap` respectively.

cansavvy:
I also think we should reconsider renaming this module to be more reflective of the central goal which is a heatmap not really clustering. We could rename it to `clustered-heatmap` so the word `heatmap` is actually in the title. Additionally we should tell people how to shut off the clustering in case the heatmap part is all they really wanted.

cansavvy:
Here we should go with the Non-QN'ed data -> DESeq2 normalization -> use `assay()` to extract the normalized data.

cbethell:
Based on @cansavvy's comments above, my plan is to:

1. Copy the microarray clustering notebook and rename the module and notebook to include the word `heatmap`.
2. Find a suitable RNA-Seq dataset (with non-QN'ed data), place it into the `data` directory, and read that into the notebook.
3. Perform the DESeq2 normalization (using http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html#heatmap-of-the-count-matrix as inspiration to borrow from) and use `assay()` to further analyze the now normalized data.
3. Keep the section that chooses genes of interest using variances, but add a step to define a minimum TPMs cutoff. 
4. Use the `pheatmap` package to create a basic clustered heatmap (perhaps add an additional heatmap to show that the clustering can be turned off, although comments may be suffice here).

Does this plan appear to fall in line with your above thoughts @cansavvy?

[...]

Issue #137, opened 2020-07-10
Analysis Updates: RNA-seq differential-expression module 


cansavvy:
Since we are trying to integrate more DESeq2 functionality into our RNA-seq section, we should make differential expression reflect that as well since that is its main utility. 

We can borrow from training-modules steps: https://github.com/AlexsLemonade/training-modules/blob/master/RNA-seq/05-nb_cell_line_DESeq2.Rmd and maybe useful tidbits from here: https://github.com/AlexsLemonade/training-modules/blob/master/RNA-seq/03-gastric_cancer_exploratory.Rmd

<del>I'm debating about whether we should still maintain a QN'ed from refinebio version of differential-expression? One strategy is we could keep the current example but reframe it as a "pre-normalized DE" example and make the new notebook a "starting at counts DE" example. Thoughts? </del>

jaclyn-taroni:
The current example isn't QN'd data.

cansavvy:
Oh I see that now. Hmm.. I guess I was thrown off by DGEList thing. Why did we choose limma-voom here?

jaclyn-taroni:
I do not have a good reason.

cansavvy:
We should look at the examples listed here and decide what we would like to do (or maybe a few of them): https://bioconductor.org/packages/release/bioc/vignettes/tximport/inst/doc/tximport.html#Downstream_DGE_in_Bioconductor

I do not understand the rationale of limma-voom vs the various DESEq2 models and I suspect this will also be an area of confusion for our users. 

So although we don't want to get too far into the weeds on dge models, we should evaluate: 
- what steps we are recommending
- why we are recommending them (when might they not be a good idea)
- and how this might fit in with the new RNA-seq modules being added (#110)

cansavvy:
I think I remember reading this before, but it's a great little summary to keep in mind for this: https://www.biostars.org/p/284775/

cansavvy:
Also this by Mike Love: https://mikelove.wordpress.com/2016/09/28/deseq2-or-edger/

cansavvy:
I don't feel like its super necessary or helpful to keep an example of each edgeR, limma-voom, and DESEq2. So what I think might be a good way forward is to change this example to be DESEq2-centric, but briefly acknowledge the others and link out to examples of those as well as the nice summary articles. 

Our rationale for choosing DESeq2 would basically be 
- Other methods are all about as good for normalization purposes
- It has a lot of nice docs and examples
[...]

Issue #125, opened 2020-07-09
Write introduction to RNA-seq section header Rmd


cansavvy:
Related to #123 

This `intro_to_rna-seq.Rmd`  should include:

1. A brief introduction to rna-seq technology
2. A primer on DESeq2 and include a rationale about why we use it. 
    - What is DESeq2 and why it is handy? 
        - Plays nicely with tximport which refinebio uses
        - It includes a lot of functions that can perform all your basic differential expression needs. 
    - DESeq2 normalization vs refinebio QN
          - How do we get QN'ed vs non-QN'ed from refinebio
          - Compare and contrast the normalization methods (QN'ed by us vs VST vs rlog vs ???)
          - How did we choose which normalization method to use in each tutorial example? 
3. Further reading about RNA-seq technologies and DESeq2
4. TOC

I figure we can post good articles and ideas for this introduction here.

cansavvy:
Related to: https://github.com/AlexsLemonade/refinebio-examples/issues/103#issuecomment-656737302

Some sources we should include: 
- [`DESeq2`](https://bioconductor.org/packages/release/bioc/html/DESeq2.html)
- [Love, Anders, and Huber. _Genome Biology_. 2014.](https://doi.org/10.1186/s13059-014-0550-8)
- [excellent vignette](https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html)

cansavvy:
We should consider making a nice diagram or two akin to the [training-module workflow diagrams](https://github.com/AlexsLemonade/training-modules/blob/master/RNA-seq/diagrams/rna-seq_1.png). 
Or perhaps other visuals.

cansavvy:
This RNA-seq header section will be linked out to individual analyses in instances where a user may want more information, but we don't want to get too far off the rails of the specific analysis. 
Related: https://github.com/AlexsLemonade/refinebio-examples/pull/140/files#r456047768

Two instances in particular when we will want to "link out" to this header document would be: 
- whenever we pick a normalization method - link to the about normalization section 
- whenever we create a DESEq2DataSet - link out to about "DESEq2"

For now, while this refinebio-examples is not live and still under construction, we will put some TODOS like: `TODO: link to RNA-seq header` tags where these links will need to end up being.

cansavvy:
[...]

PR #122, opened 2020-07-09
WIP: Testing out intro to an example analysis: Clustering

cansavvy:
## Purpose:

Toward #116 this is a rough draft of testing what this looks like. 

For context, this is a tester to show what an individual analysis' "getting started" section would look like. In the event that a user cherry picks an analysis for following, we want to make sure that all the information is there (https://github.com/AlexsLemonade/refinebio-examples/issues/114#issuecomment-655626702) This would be separate from the larger header of "getting started" described on #109. 

### Concerns/questions for reviewers: 
- I don't like that 95% this would need to be copy and pasted to every single example, so if we can find a way to make this less cumbersome when we want to make changes, that'd be good.
- What sections here should we potentially link out to the overall `Getting Started` section #109 
- The headers are all so big the words get lost! I do like what they are doing with the TOC on the side. But how do we make this more visually appealing and easy to follow? I might try playing with aesthetics. 
- III want a header that more clearly signifies the start of the analysis steps. Right now it's just `Example clustering analysis`.

Here's a zipped up version of the rendered notebook if you don't want to checkout the branch. 
[clustering_example.nb.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/4898307/clustering_example.nb.html.zip)

Here's a preview of the section headers: 
```
# Purpose of the analysis:
## Further learning resources about this analysis
***
# How to run this example
## About dataset we are using for this example
## Obtaining the dataset from refinebio
### Obtaining this example R notebook
### Set up your analysis folder
### Running the example analysis
### Using your own data
***
```

PR #140, opened 2020-07-15
Add RNA-Seq clustering heatmap example notebook

cbethell:
## Purpose:

This PR addresses the issue creating a clustering example with RNA-seq data (#127).

As noted in one of my [comments on that issue](https://github.com/AlexsLemonade/refinebio-examples/issues/127#issuecomment-656948685), I followed a plan in the order as follows:

1. Copy the microarray clustering notebook and rename the module and notebook to include the word `heatmap`.
2. Find a suitable RNA-Seq dataset (with non-QN'ed data) and placing it into the `data`.
3. Create the DESeqDataSet.
4. Add a step to define a minimum counts cutoff. 
5. Perform the DESeq2 normalization and use `assay()` to further analyze the now normalized data.
6. Keep the section that chooses genes of interest using variances.
7. Use the `pheatmap` package to create a basic clustered heatmap and add an additional heatmap to show that the clustering can be turned off.

## Questions for reviewers

- Does the dataset seem like an appropriate dataset to use for this example (specifically worried about the heatmap)?
- Are there any suggestions for a less precarious way of defining the `Groups` variable for the heatmap annotation? (Still looking into this myself)
- Are the comments clear/suffice?

## Pull Request Check List:
* [x] Run a linter
* [x] Comments and/or documentation up to date
* [x] Double check your paths


03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+# Subset the data choosing only genes whose variances are in the upper quartile
+df.by.var <- df[which(variances > sum.stats.var[5]), ]
+```
+
+## 4) Create a DESeq dataset
+A `DESeqDataSet` object is required to carry out the normalization step that
+you will see later using a function from the `DESeq2` package. This chunk of
+code will turn the data.frame (or matrix) into a `DESeqDataSet` object.
+
+```{r Create a DESeq dataset}
```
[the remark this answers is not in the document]
cansavvy: This likely going to be a thing we do many times in the whole RNA-seq section's many modules so we may want to have it pretty well honed down. Can also add an explanation about `DESeqDataSet`'s and etc in the RNA-seq header section (#125) which we can link to here.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+  colData = metadata,
+  design = ~refinebio_title
+)
+
+# Place the colData in a separate data.frame object for annotation
+col_df <- as.data.frame(colData(dds))
+
+# Define a `Groups` column variable
+col_df <- col_df %>%
+  dplyr::mutate(Groups = rep(c("SP", "MP"), # Replace with group names relevant to your data
```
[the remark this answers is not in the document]
cansavvy: In other words, can you explain why you used this `colData(dds)` strategy you are using here? It isn't clear to me, but this could because I haven't tried this myself.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+```
+
+## 2) Import and set up data
+Data downloaded from refine.bio include a metadata tab separated values ("tsv")
+file and a data tsv file. This chunk of code will read the both tsv files and 
+add them as data.frame objects to your environment.
+
+The data that we downloaded from refine.bio for this analysis is a dataset, with
+a total of four samples, containing RNA-sequencing results for mouse primary
+natural killer (NK) cells and NK cells treated with DKK2 for 24 hours.
```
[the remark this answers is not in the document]
cansavvy: Ignore this comment until we get a new, bigger dataset

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+plots_dir <- "plots" # Replace with path to desired output plots directory
+
+# Create the plots folder if it doesn't exist
+if (!dir.exists(plots_dir)) {
+  dir.create(plots_dir)
+}
+```
+
+## 2) Import and set up data
+Data downloaded from refine.bio include a metadata tab separated values ("tsv")
```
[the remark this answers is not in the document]
cbethell: Good point!

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+alternatively, the heatmap could be created using only genes of interest. 
+For this example, we will sort genes by variance, but there are many alternative
+criterion by which you may want to sort your genes <i>eg</i> fold change,
+t-statistic, membership to a particular gene ontology, so on. 
+
+```{r Choose genes}
+# Calculate the variance for each gene
+variances <- apply(df, 1, var)
+
+# Determine summary statistics for gene variances
```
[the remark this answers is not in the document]
cbethell: Sounds good and no worries!

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+sum.stats.var <- summary(variances)
+
+# Subset the data choosing only genes whose variances are in the upper quartile
+df.by.var <- df[which(variances > sum.stats.var[5]), ]
+```
+
+## 4) Create a DESeq dataset
```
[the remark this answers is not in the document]
cbethell: Another good point 👍

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+## 3) Choose genes of interest
+Although you may want to create a heatmap including all of the genes in the set,
+alternatively, the heatmap could be created using only genes of interest. 
+For this example, we will sort genes by variance, but there are many alternative
+criterion by which you may want to sort your genes <i>eg</i> fold change,
+t-statistic, membership to a particular gene ontology, so on. 
+
+```{r Choose genes}
+# Calculate the variance for each gene
+variances <- apply(df, 1, var)
```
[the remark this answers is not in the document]
cbethell: I think it's a good idea 👍 (we do use the `apply` function elsewhere)

03-rnaseq/clustering/clustering_heatmap_example.Rmd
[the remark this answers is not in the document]
cbethell: I will give your first suggestion a try and let you know from there. I did try something similar before and that did not work, but I believe it was due to the way I generated the `Groups` column in that case.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,220 @@
...
+```
+
+
+## 5) Define a minimum counts cutoff
+We are going to do some pre-filtering to keep only rows with 10 or more reads
+total.
+
+```{r Define minimum counts cutoff}
+# Define a minimum counts cutoff
+genes_to_keep <- rowSums(counts(dds)) >= 10
```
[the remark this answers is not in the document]
cbethell: Sounds good to me 👍

cbethell:
> I think this is great start to get the ball rolling! Honestly this PR brought up a lot of things for us to think about. With a lot of them related to #125 - I'm going to put these notes down on that issue.
> 
> I have a lot of comments, but tbh, a lot of them are to 2018 Candace's choices (I probably should have made it more clear to you that you are free to burn this whole notebook down and start fresh if you so want). But while we are revamping these notebooks, we should probably take the opportunity to change anything we don't think is ideal. (No time like the present I suppose).
> 
> But now that we have the ball rolling, I think the bigger change we may want to make is to use a dataset with more samples since a clustering of n = 4 is not super helpful (Thanks a LOT 2018 Candace for the bad example).
> 
> Here's what I think the plan should be moving forward:
> 
> 1. Find a bigger RNA-seq dataset with metadata labels we can use - feel free to post your ideas on the issue #127 for feedback.
> 2. Make this notebook fit that dataset. I would try it out a "quick and dirty" clustering on it before revamping the rest so we don't get to attached - feel free to post the plot you get. We won't want a dataset that clusters to nonsensically or too perfectly - something that's a good, realistic example but has sensible enough metadata labels we can use for trying to ground this in some semblance of truth.
> 3. Then we can go through addressing comments listed here (except for the RNA-seq header links parts).
> 4. For the time being, you can ignore the items where I said we will need to link it to the RNA-seq header section (#125) and just put a `TODO: link to RNA-seq header` I'll write the things we've found here on #125 and when that gets addressed we can come back put the links.
> 
> Let me know if what I'm saying here makes sense, feel free to send me your ideas.

It all makes sense, I'll implement the changes/steps you suggested here update the PR along the way. I will also take you up on that suggestion of posting on the issue for feedback on the dataset and plot 👍 Thank you!

03-rnaseq/clustering/clustering_heatmap_example.Rmd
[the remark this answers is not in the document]
cbethell: This is the only comment I have yet to address because I have not yet thought of an appropriate tidyverse strategy to replace what the apply function is doing here.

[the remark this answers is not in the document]
jashapiro: Apply is fine here. This is essentially a matrix, and you are doing a row-wise calculation, which tidyverse is less adept at. 
A suggestion you could make is to use `matrixStats::rowVars()`, which is much faster than `apply()`, but I think apply is fine.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,296 @@
...
+you will see later using a function from the `DESeq2` package. This chunk of
+code will turn the data.frame (or matrix) into a `DESeqDataSet` object.
+
+```{r Create a DESeqDataset}
+# The `DESeqDataSetFromMatrix` function needs the values to be converted to
+# integers
+## TODO: Link to explanation about `DESeqDataSet` in the RNA-seq header section
+df_by_var <- df_by_var %>%
+  # Mutate numeric variables to be integers
+  dplyr::mutate_if(is.numeric, as.integer)
```
jashapiro, line 140: `as.integer()` always rounds down, so it should be avoided.
```suggestion
  dplyr::mutate_if(is.numeric, round)
```

cbethell: Gotcha 👍

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,296 @@
...
+## 6) Perform DESeq2 normalization
+We are going to use a function from the `DESeq2` package to normalize the data. 
+
+```{r Perform DESeq2 normalization}
+# Normalize the data in the `DESeqDataSet` object using the
+# `normTransform` function from the `DESEq2` R package
+## TODO: Add a note/link to section in documentation on why we use
+# `normTransform` vs `vst` and `rlog`
+# Let's normalize the data for the collapsed counts first 
+ntd_collapsed <- normTransform(dds_collapsed)
```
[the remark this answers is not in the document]
cbethell: I left this TO DO because I am not sure myself. My decision came from the original issue #127 where you mentioned `DESeq2` normalization, so I figured that the `normTransform`, being a function from the `DESeq2` package, was the best to implement here and it was also implemented in the vignette you put on the original issue.

cbethell:
Thank you for the feedback @cansavvy! I _believe_ I rearranged the notebook as you described in your [comment above](https://github.com/AlexsLemonade/refinebio-examples/pull/140#pullrequestreview-452804156) and addressed your smaller review comments.

Please let me know if I misinterpreted any of your comments! Also, the comment that I added at the end on the results represented on the heatmap could probably use more detail, but I wanted to ensure that this is the heatmap that we want (however, feel free to comment on my lack of detail there if you agree and possibly what you would like to see added there).

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+}
+```
+
+## 2) Import and set up data
+Data downloaded from refine.bio include a metadata [tab separated values ("tsv")]
+(https://en.wikipedia.org/wiki/Tab-separated_values) file and a data tsv file.
+This chunk of code will read the both tsv files and add them as data.frame
+objects to your environment.
+
+The data that we downloaded from refine.bio for this analysis is a dataset, with
```
cansavvy, line 69:
```suggestion
The data that we downloaded from refine.bio for this analysis has
```

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+a total of 24 samples (obtained from 6 mice with 4 technical replicates for each
```
cansavvy, line 70:
```suggestion
24 samples (obtained from 6 mice with 4 technical replicates for each
```

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+mouse), containing RNA-sequencing results for Group 3 and 4 medulloblastoma
+orthotopic xenograft mice with and without digoxin treatment.
+
+You can download this dataset by going to the following link and clicking 
+`Download Now`: https://www.refine.bio/experiments/SRP150101/transcriptome-analysis-of-group-3-and-4-medulloblastoma-orthotopic-xenograft-mice-with-digoxin-treatment
```
cansavvy, line 75: This part is actually not necessary here, because it will be covered when we add in the getting started section for this module (being workshopped over on #122 ).

cansavvy: But you can keep it here for now I guess.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+
+```{r Import data from .tsv files}
+# Read in metadata tsv file
+metadata <- readr::read_tsv(file.path(
+  "data", # Replace with path to your metadata file
+  "metadata_SRP150101.tsv" # Replace with the name of your metadata file
+)) %>%
+  dplyr::select(refinebio_accession_code, refinebio_title, refinebio_treatment) %>%
+  replace(is.na(.), "None") %>%
+  dplyr::mutate(
```
cansavvy, line 85: Can we put these `mutate` steps in their own chunk? This way the user has a chance to see what `metadata` looked like before. Would also be good to add a chunk in between where you print out a preview of `metadata`.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+# Determine the upper quartile variance cutoff value
+upper_var <- quantile(variances, 0.75)
+
+# Subset the data choosing only genes whose variances are in the upper quartile
+df_by_var <- df %>%
+  dplyr::filter(variances > upper_var)
+```
+
+## How are our technical replicates clustering together?
+
```
cansavvy, line 141: This is great to start out with the question like we discussed. Can you try to add some words here to explain why we are doing this and give a brief summary of what we will be doing?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+  design = ~ refinebio_treatment
+)
+```
+
+### 5) Prepare the metadata
+We are going to format the metadata to be used for annotation on the heatmap.
+
+```{r}
+# Let's prepare the metadata for the uncollapsed `DESeqData` set object
+metadata_uncollapsed <- metadata %>%
```
cansavvy, line 168: This metadata will only be used for annotation correct? 
If yes, can we instead call this something like "annotation_df" or something like that? 

This is in part because I'd like to move away from using `collapsed` `uncollapsed` terminology and instead try to use terms that are more reflective of the samples and the question at hand. I think this will be helpful for our users to keep focused on the question/purpose of what we are doing.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+  dplyr::select(-c("sample_info_summary", "refinebio_title")) %>%
```
cansavvy, line 169: Can you add a comment to explain what you are doing here? aka why are we dropping these two columns?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+  tibble::column_to_rownames("refinebio_accession_code")
```
cansavvy, line 170: Also comment here for why we are storing these as rownames (cuz pheatmap I assume).

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+```
+
+### 6) Define a minimum counts cutoff
+We are going to do some pre-filtering to keep only rows with 10 or more reads
```
cansavvy, line 174: Can you add a sentence or two about why we are doing this? Users may not know why.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+total.
+
+```{r Define minimum counts cutoff}
+# Let's take a look at the distribution of counts first
+
+# Sum up the counts for each row
```
cansavvy, line 180: We might be a little more comfortable with rows = genes and columns = samples. But our users are probably not as comfortable with this. Can you bring these steps back to what they mean in terms of biology?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+                       "yellow"))(25
+                       )
+  )
+```
+
+"Now that we've seen our replicates are fairly identical... let's take a look at our experimental question". 
+
+## How are our biological replicates clustering together?
+
+### 4) Create a DESeqDataset
```
cansavvy, line 235: We may want to get rid of the numbering? What do you think?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,338 @@
...
+A `DESeqDataSet` object is required to carry out the normalization step that
```
cansavvy, line 236: I'm debating about whether we need to explain this or not again in the same notebook. Unsure. My default thought is to keep it.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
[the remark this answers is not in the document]
cansavvy: When you have time (so maybe not before we get training stuff wrapped up), I would look at these docs for: 
- [`normTransform`](https://www.rdocumentation.org/packages/DESeq2/versions/1.12.3/topics/normTransform) 
and see if you can decipher how this is different than 
- [`rlog`](https://www.rdocumentation.org/packages/DESeq2/versions/1.12.3/topics/rlog) or 
- [`vst`](https://www.rdocumentation.org/packages/DESeq2/versions/1.12.3/topics/varianceStabilizingTransformation)

We can chat about it sometime. The paper might also be helpful to you: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4302049/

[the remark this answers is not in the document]
cbethell: Gotcha! :thankyou:

03-rnaseq/clustering/clustering_heatmap_example.Rmd
cbethell: I had the same though as it could be a bit confusing, although I also do think that it can be helpful in allowing users to keep track of the steps sequentially if they go back to the notebook at some point and want to use it to answer just one of the questions.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
cbethell: In your [previous review comment](https://github.com/AlexsLemonade/refinebio-examples/pull/140#discussion_r456024079), you mentioned that there is a likelihood of someone skipping over this in the getting started section so it is good to have here as well.

03-rnaseq/clustering/clustering_heatmap_example.Rmd
cansavvy: Can we go back and make it 4a and 4b or is that more confusing?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,387 @@
...
+dds <- dds[genes_to_keep, ]
+```
+
+### 7) Perform DESeq2 normalization
+We are going to use a function from the `DESeq2` package to normalize the data. 
+
+```{r Perform DESeq2 normalization}
+# Normalize the data in the `DESeqDataSet` object using the
+# `normTransform` function from the `DESEq2` R package
+## TODO: Add a note/link to section in documentation on why we use
```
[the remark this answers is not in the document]
cbethell: Gotcha.

cansavvy:
> we used much of what we have here (e.g., the dataset and collapsing steps) as a DGE example?

I like this idea. So if we move this to DGE (probably delete the cluster and heatmap steps). Find a simpler/lower complexity dataset for this clustering example so we can remove the collapse/uncollapse steps from here?

03-rnaseq/clustering/clustering_heatmap_example.Rmd
```diff
@@ -0,0 +1,435 @@
...
+                                   color = "red",
+                                   linetype = "dashed")
+```
+
+```{r Define a minimum counts cutoff for all replicates object}
+# Define a minimum counts cutoff and filter `DESeqDataSet` object to include
+# only rows that have counts above the cutoff
+genes_to_keep_all_reps <- rowSums(counts(dds_all_reps)) >= 10
+dds_all_reps <- dds_all_reps[genes_to_keep_all_reps, ]
+```
```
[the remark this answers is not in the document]
cbethell: This was included per @cansavvy's review suggestion [here](https://github.com/AlexsLemonade/refinebio-examples/pull/140#discussion_r456042011), but a summary of the reasoning here is to plot the distribution of counts before defining a cutoff.

cbethell:
> > we used much of what we have here (e.g., the dataset and collapsing steps) as a DGE example?
> 
> I like this idea. So if we move this to DGE (probably delete the cluster and heatmap steps). Find a simpler/lower complexity dataset for this clustering example so we can remove the collapse/uncollapse steps from here?

I also have the same question here @jaclyn-taroni. I started to look for a simpler dataset and can post one or two once I find them to get feedback on whether it's a dataset that would be better suited for the notebook.

jaclyn-taroni:
Yes I agree with:

> Find a simpler/lower complexity dataset for this clustering example so we can remove the collapse/uncollapse steps from here

Keep an eye on how much metadata smoothing is required, too, as you're picking a dataset. My understanding is that a lot of the metadata "stuff" that has to happen here is also because of the replicates.

cbethell:
Any thoughts on whether or not the following dataset would be better suited for this notebook?

[A role for p53 in the adaptation to glutamine starvation](https://www.refine.bio/experiments/SRP151009/a-role-for-p53-in-the-adaptation-to-glutamine-starvation-through-the-expression-of-slc1a3#samples)

Screen Shot 2020-08-04 at 5 24 22 PM

I am still in search for a dataset with a larger number of individual samples and lower complexity, but wanted to put this here on the ticket in the meantime in case someone thinks we should go ahead with this one.

jaclyn-taroni:
Now that the plan for using this notebook to get started on a DGE example is detailed on #137 and we have a plan to move forward with clustering specifically (https://github.com/AlexsLemonade/refinebio-examples/issues/127#issuecomment-669347715), I am going to close this pull request, retain the `cbethell/add-clustering-rna-seq-example` and disconnect this PR from #127.

PR #148, opened 2020-08-05
Contributing guidelines

cansavvy:
## Purpose:
Closes #144 

This is a draft of style and contributing guidelines. 

I tested out some proposed changes to the header on the clustering notebook for microarray.

Note that the snakemake instructions won't work unless you move to that branch (or it is merged in).

PR #150, opened 2020-08-05
Custom rendering script 

### LINKED ISSUE #149: Create custom rendering script
To make citations render correctly, `pandoc` needs a `bibliography:` specification. However, if the user downloads the `Rmd` on their own to run and does not have the file specified by `bibliography:` , `pandoc` fails to render. See the error and more discussion over here: https://github.com/AlexsLemonade/refinebio-examples/pull/148#issuecomment-669170681

The work around we've come to as of now is to make all refinebio-examples notebooks be called by a custom rendering script that will add on the `bibliography:` script upon rendering. This will leave the `.Rmd` files ready for download and easy use without a `.bib` file. 

Talking to @jashapiro to complete this add on to the `.Rmd` header chunk, we can borrow aspects from `exrcise` package code: https://github.com/AlexsLemonade/exrcise/tree/master/R
and ultimately use the `render` command: https://rmarkdown.rstudio.com/docs/reference/render.html

The script would look something like: https://github.com/AlexsLemonade/training-modules/blob/master/scripts/make-live.R

The usage of this will also need to be documented in the (currently WIP #148) CONTRIBUTING.md

cansavvy:
## Purpose:
Closes #149 

This script is modeled after the `make-live.R` script and `exrcise` package functions. 
It works for adding the bibliography specification. But really the function I made could be used in a much broader capacity for adding anything to the rendered version that we don't want in the regular version. (But one step at a time!). 

- Currently I have only added the getting_started.Rmd file to the script because its the only one "ready for prime time".  We will need to add appropriate references to each module as its revamped. (This will all be added to #148 doc and the appropriate issues). 
- I will add usage instructions for this script on the #148 PR CONTRIBUTING.md doc. But basically just run `scripts/render-notebooks.R` 
- There's a `--bib_file` option where we can supply the bibliography file in whatever pandoc allows. It needs to be supplied in reference to the root git repository so all the files are able to find it. Hopefully this is not too confusing. I have the default option 
- I haven't been able to make the navbar thing work yet. but we should be able to easily add it in once we figure it out. I will add this note on the #118 issue ("make this easily navigable"). One thing at a time.


01-getting-started/getting-started.Rmd
```diff
@@ -5,6 +5,15 @@ output:
   html_notebook: 
     toc: true
     toc_float: true
+navbar:
```
cansavvy, line 8: This isn't working. I think its more complicated than what I tried.

01-getting-started/getting-started.Rmd
```diff
@@ -37,10 +46,10 @@ Each analysis is self-contained and provides information with how to obtain the
 We encourage you to download the [`.Rmd`](http://rmarkdown.rstudio.com) and follow the "getting started" section in the example before diving into our analysis examples.
 
 **Each analysis contains:**   
-- A README that introduces you to the analyses, concepts, requirements, and workflows for that module.
-- An R Notebook which consists of:
-     - An R markdown (`.Rmd`) file(s) that you can use in RStudio to run the analysis and contains it's own "getting started" section which describes how to download the example dataset from refine.bio. 
-     - An nb.html file that is the resulting output of the Rmd file rendered as an HTML file.
+- A README that introduces you to the analyses, concepts, requirements, and workflows for that module.  
+- An R Notebook which consists of:  
+    - An R markdown (`.Rmd`) file(s) that you can use in RStudio to run the analysis and contains it's own "getting started" section which describes how to download the example dataset from refine.bio.
```
cansavvy, line 51: These bullets are not indenting properly.

scripts/render-notebooks.R
```diff
@@ -0,0 +1,71 @@
...
+                            outfile = file.path(dirname(infile), "tmp-header-changed.Rmd")) {
+
+  # Read in as lines
+  lines <- readr::read_lines(infile)
+
+  # Find which lines are the beginning and end of the header chunk
+  header_range <- which(stringr::str_detect(lines, "---"))
+
+  # Add the bibliography specification line at the beginning of the chunk
+  new_lines <- append(lines, bib_specify_line, header_range[1])
```
jashapiro, line 54:
```suggestion
  new_lines <- append(lines, header_line, header_range[1])
```

scripts/render-notebooks.R
```diff
@@ -0,0 +1,71 @@
...
+    # Save to original html output file name
+    output_file = stringr::str_replace(infile, ".Rmd", ".nb.html")
+  )
+
+  # Remove the temporary header change .Rmd output file
+  file.remove(outfile)
+}
+
+# Render all the notebooks but with added header line
+purrr::map(infiles, add_header_line)
```
jashapiro, line 71: If you have made a list of tempfiles, and one of the outfiles then you can use `map2()` as below.

```suggestion
html_output = rmarkdown::html_document(toc = TRUE, toc_float = TRUE, number_sections = TRUE),
purrr::map2(infiles, tempfiles, add_header_line, header_line = bib_specify_line)
purrr::map2(tempfiles, html_files, ~ rmarkdown::render(.x, output_file = .y, output_format = html_output))
purrr::map(tempfiles, file.remove)
```

Alternatively, you could have `add_header_line()` return the outfile name, then use it something like this

scripts/render-notebooks.R
```diff
@@ -0,0 +1,71 @@
...
+# Make specialized function for adding a header specification
+add_header_line <- function(infile,
+                            header_line = bib_specify_line,
+                            outfile = file.path(dirname(infile), "tmp-header-changed.Rmd")) {
+
+  # Read in as lines
+  lines <- readr::read_lines(infile)
+
+  # Find which lines are the beginning and end of the header chunk
+  header_range <- which(stringr::str_detect(lines, "---"))
```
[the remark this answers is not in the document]
cansavvy: Ah. Good point. It's also shorter too!

scripts/render-notebooks.R
```diff
@@ -0,0 +1,71 @@
...
+
+# Load library:
+library(optparse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-b", "--bib_file"), type = "character",
+    default = "references.bib", # Default is this file, but it can be changed
+    help = "File path relative to root directory (where .git is)
```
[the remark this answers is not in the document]
cansavvy: I don't mind having the files being processed one at a time, but it does concern me that a file might be changed and accidentally not re-rendered with this script. 

Is there a non-janky way to optionally re-render only files that have had edits done to them?

scripts/render-notebooks.R
cansavvy: The only problem I have with this (unless I'm wrong) is that if something fails, i'm going to have to go through and delete any temporary files that were left. Is that right?

scripts/render-notebooks.R
cansavvy: Good catch

scripts/render-notebooks.R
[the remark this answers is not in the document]
jashapiro: snakemake? (ducks)

scripts/render-notebooks.R
jashapiro: Yes, but I think that is better in general. Then you have the files for debugging. And the way things are it will overwrite files by default, so they would be deleted on the next working run anyway.

scripts/render-notebooks.R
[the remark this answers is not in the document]
jashapiro: To be honest, this is actually a good argument for a non-monolithic script (one that works on one file at a time). And really it is a pretty good place to start writing `snakemake`! This would be a one step workflow: you would have a `main` rule that defines the output files, and a `render` rule that looks for the appropriate input file to generate those output files then runs the appropriate shell script.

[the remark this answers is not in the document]
cansavvy: You've convinced me.

scripts/render-notebooks.R
cansavvy: So can I ignore this part then? We will move to a script that processes on file at a time and I'll use snakemake to have it go through all the files?

jashapiro: Yup. Ignore it. Sorry 🐈 , we are about to get a 🐍

scripts/render-notebooks.R
```diff
@@ -27,48 +31,43 @@ option_list <- list(
...
+}
 
-# Check that the file exists in the root directory
-if (!file.exists(normalized_bib_file)) {
-  message(normalized_bib_file)
+# Check that the file exists
+if (!file.exists(opt$bib_file)) {
   stop("File specified for --bib_file option is not at the specified file path.")
 } else {
   bib_specify_line <- paste("bibliography:", normalized_bib_file)
```
jashapiro, line 47: You probably want to get the full path to avoid `.Rmd` path weirdness here.

```suggestion
  bib_specify_line <- paste("bibliography:", normalizePath(opt$bib_file))
```

scripts/render-notebooks.R
```diff
@@ -31,11 +31,7 @@ option_list <- list(
 # Parse options
 opt <- parse_args(OptionParser(option_list = option_list))
 
-opt$bib_file <- "references.bib"
-opt$rmd <- "01-getting-started/getting-started.Rmd"
-
 # Make these file paths
-opt$bib_file <- normalizePath(opt$bib_file)
 opt$rmd <- file.path(opt$rmd)
```
jashapiro, line 35: I don't think this is doing anything, so you can delete it.

scripts/render-notebooks.R
```diff
@@ -0,0 +1,80 @@
...
+}
+
+# Add the bibliography specification line at the beginning of the chunk
+new_lines <- append(lines, header_line, header_range[1])
+
+# Write to an outfile
+readr::write_lines(new_lines, tmp_file)
+
+# Specify final output file
+output_file <- stringr::str_replace(basename(opt$rmd), "\\.Rmd$", ".nb.html")
```
jashapiro, line 67: I'm going to suggest you make the final output file an option too. Maybe use this if it isn't specified, but flexibility is good. (And move this up with the other option parsing.)

Don't use `basename()`, or things might not go where expected.
```suggestion
if (is.null(opt$html)){
  output_file <- stringr::str_replace(opt$rmd, "\\.Rmd$", ".html")
} else {
  output_file <- opt$html
}
```

scripts/render-notebooks.R
```diff
@@ -0,0 +1,80 @@
...
+}
+
+# Check that the file exists
+if (!file.exists(opt$bib_file)) {
+  stop("File specified for --bib_file option is not at the specified file path.")
+} else {
+  header_line <- paste("bibliography:", normalizePath(opt$bib_file))
+}
+
+# Specify the output file
```
jashapiro, line 46:
```suggestion
# Specify the temp file
```

scripts/render-notebooks.R
```diff
@@ -0,0 +1,80 @@
...
+
+# Stop if no chunk found
+if (length(header_range) == 0) {
+  stop("Not finding the `---` which is at the beginning of the header.")
+}
+
+# Add the bibliography specification line at the beginning of the chunk
+new_lines <- append(lines, header_line, header_range[1])
+
+# Write to an outfile
```
jashapiro, line 63:
```suggestion
# Write to tempfile
```

01-getting-started/getting-started.Rmd
jashapiro: I think this needs to go in a separate file.

scripts/render-notebooks.R
cansavvy: It would not let me specify the whole file path, only the file name. It seems like its assuming it will go in the same directory as the file I give it. 🤷‍♀️

jashapiro: https://rmarkdown.rstudio.com/docs/reference/render.html says a path should work? maybe `normalizePath(opt$rmd)` here for safety?

cansavvy: It seems it wants the file path relative to the `.Rmd` provided in input. Unless there's a way to not make it this way.

cansavvy: normalizePath() works with this.

jashapiro: That's why I think a full path may work. but don't stress about it if not. this is not a case we need to care about.

cansavvy: It thinks my directory doesn't exist even though dir.exist() says it does.

jashapiro: Updating my suggestion to include normalizePaths (requires the addition of an `--html` option)
```
if (is.null(opt$html)){
  output_file <- stringr::str_replace(normalizePath(opt$rmd), "\\.Rmd$", ".html")
} else {
  output_file <- normalizePath(opt$html)
}
```

cansavvy: Implemented this option in [9a25c9e](https://github.com/AlexsLemonade/refinebio-examples/pull/150/commits/9a25c9e43203017bf3bbe67bb2eea76147d41d02)

01-getting-started/getting-started.Rmd
jashapiro: I think this may be because you have extra spaces at the ends of the lines? Unless that was you trying to fix it?

cansavvy: Yeah, I've added and deleted spaces.

Snakefile
```diff
@@ -1,3 +1,7 @@
+rule target:
+    input: 
+        "01-getting-started/getting-started.html" 
+
 rule render_citations:
     input:
         "01-getting-started/getting-started.Rmd"
```
jashapiro, line 7: This is where the `output:` comes in. It allows snakemake to trace backward to figure out which rules need to be run. Right now, you probably already have `getting-started.html` so when you run `snakemake` it says there is nothing to do. What you want to have happen is that when you delete that file (or modify the .Rmd) the steps to generate the file are rerun. To do that, you need things linked up so every input in the target is the same as a an output from some rule (and all the inputs from that are outputs from somewhere, or separately provided files).

```suggestion
        "01-getting-started/getting-started.Rmd"
    output:
        ""01-getting-started/getting-started.html"
```

Snakefile
```diff
@@ -0,0 +1,15 @@
...
+    input:
+        rmd = "{basename}.Rmd",
+        bib = "references.bib"
+    output: 
+       "{basename}.html"
+    shell:
+        "Rscript scripts/render-notebooks.R "
+        "--rmd {input.rmd} "
+        "--bib_file {input.bib} "
+        "--html {output}"
```
jashapiro, line 15: Very minor style preference here (plus an endline):
```suggestion
        "Rscript scripts/render-notebooks.R"
        " --rmd {input.rmd}"
        " --bib_file {input.bib}"
        " --html {output}"
        
```

Snakefile
cansavvy: Implemented in https://github.com/AlexsLemonade/refinebio-examples/pull/150/commits/096f52610d31757e65513c3d6af8852506fad524

.gitignore
```diff
@@ -4,6 +4,8 @@ _site
 .DS_Store
 .RData
 .Rhistory
+.snakemake/*
```
jashapiro, line 7: Shouldn't matter
```suggestion
.snakemake
```

01-getting-started/getting-started.Rmd
```diff
@@ -93,7 +93,7 @@ R Markdown documents have the `current directory` always set as wherever the `.R
 This means all file paths specified in the `.Rmd` must be specified _relative_ to the location of the Rmd. 
 
 For more practice with setting file paths in `.Rmd` files see these:  
-- [This handy course chapter from Baumer and Crouser](https://beanumber.github.io/sds192/lab-import.html#paths)  
+- [This handy course chapter from Baumer and Crouser](https://beanumber.github.io/sds192/lab-import.html#paths)
```
jashapiro, line 96: This fixes rendering. 
```suggestion

- [This handy course chapter from Baumer and Crouser](https://beanumber.github.io/sds192/lab-import.html#paths)
```

01-getting-started/getting-started.Rmd
```diff
@@ -49,20 +49,20 @@ These requirements can be installed by following the instructions at these links
 We recommend installing `devtools` from CRAN (e.g. running `install.packages("devtools")` in R). 
 
 ### Required software:
-* [**R**](https://cran.r-project.org/)  
-* [**RStudio**](https://www.rstudio.com/products/RStudio/) an [integrated development environment](https://en.wikipedia.org/wiki/Integrated_development_environment) for working with R and [R Notebooks](https://bookdown.org/yihui/rmarkdown/notebook.html).  
-* [**Bioconductor**](https://bioconductor.org/install/) for installing packages from this package repository. 
-* [**tidyverse**](https://www.tidyverse.org/) - we opt for using tidyverse packages for handling and cleaning the data. 
+* [**R**](https://cran.r-project.org/)  [@R-base]
```
jashapiro, line 52: Can we settle on a style for bullets? You are using `*`, `-` _and_ `+` in different bullet lists in this doc!
```suggestion

* [**R**](https://cran.r-project.org/)  [@R-base]  
```

01-getting-started/getting-started.Rmd
```diff
@@ -37,10 +37,10 @@ Each analysis is self-contained and provides information with how to obtain the
 We encourage you to download the [`.Rmd`](http://rmarkdown.rstudio.com) and follow the "getting started" section in the example before diving into our analysis examples.
 
 **Each analysis contains:**   
-- A README that introduces you to the analyses, concepts, requirements, and workflows for that module.
-- An R Notebook which consists of:
-     - An R markdown (`.Rmd`) file(s) that you can use in RStudio to run the analysis and contains it's own "getting started" section which describes how to download the example dataset from refine.bio. 
-     - An nb.html file that is the resulting output of the Rmd file rendered as an HTML file.
+- A README that introduces you to the analyses, concepts, requirements, and workflows for that module.
```
jashapiro, line 40: This will fix the rendering. You need a blank line above bullet lists.
```suggestion

- A README that introduces you to the analyses, concepts, requirements, and workflows for that module.  
```

scripts/render-notebooks.R
```diff
@@ -0,0 +1,92 @@
...
+# Write to an tmp file
+readr::write_lines(new_lines, tmp_file)
+
+# If no output html filename specification, make one from the original filename
+if (is.null(opt$html)){
+  output_file <- stringr::str_replace(normalizePath(opt$rmd), "\\.Rmd$", ".html")
+} else {
+  # Render is weird about relative file paths, so we have to do this  
+  output_file <- file.path(base_dir, opt$html)
+}
```
jashapiro, line 82: Move this block to line 55 or 58ish, since it is path-setting and that is where you have most of that.

01-getting-started/getting-started.Rmd
cansavvy: This is because I borrowed these lists from different places!

cansavvy:
I still like the doctoc for the intro to the section Rmds because I envision people might be scanning for the info they are looking for. We can delete this later if we want.

jashapiro:
> I still like the doctoc for the intro to the section Rmds because I envision people might be scanning for the info they are looking for. We can delete this later if we want. And sometimes people ignore the toc on the side.

Seems reasonable. But then I would definitely suggest adding a doctoc refresh step to the `render_citations` rule to the make sure it stays up to date.

PR #160, opened 2020-08-10
Add an RProject file

jashapiro:
## Purpose:
I like having an Rproject file for ease of development

## Strategy
I added an Rproject file and a .gitignore line

Do you like my Rproject file? 
We could change the defaults, but this should cover the main settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ _site
 .DS_Store
 .RData
 .Rhistory
+.Rproj.user
 .snakemake
 *.nb.html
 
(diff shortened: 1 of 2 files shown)


jashapiro:
> Is this something we should have written about somewhere for the users? Perhaps in the Getting Started section? Changes here seem good.

I don't think there is much to say about it, except maybe noting that it exists in the repo. It is a convenience thing that can be ignored if people don't want to use it. (In fact, most features *should* be ignored, as we will want users to be able to download bare RMD files, so `rprojroot`-type functions should not be used.)

cansavvy:
Okay. Seems fine. 👍

PR #161, opened 2020-08-10
Add navbar example

jashapiro:
## Purpose:
Add a navbar to the top of pages

## Strategy
Using the features of `rmarkdown::render()` I added a navbar that can be (partially?) reused by multiple pages. 

I started with a `.yml` file but quickly ran into its limitations, so I switched to `.html`. The nice thing about that switch is we have lots of room for customization, including dropdown menus, which are demonstrated in this version. 

More details on possible further customization and the tags used are here: https://getbootstrap.com/docs/3.3/components/#navbar 

This is just a draft to explore whether we want to do this and to add comments. There may be some challenges implementing this with the current directory structure, as nested folders can get complicated in terms of maintaining relative links. If we want to pursue this, we would probably want to flatten the directory structure somewhat. 

To actually implement this, I would probably want to make a single master `_navbar.html` file, and then link to it from the subdirectories. Alternatively, we could have separate navbars for each subdirectory, with different components expanded. For example, when working with microarray data, those components would be visible, not buried in the submenu.


cansavvy:
It looks nice!! We would add this to the the `render-notebooks.R` script? (It's unclear to me currently how you got the getting_started.html to turn out)

jashapiro:
> It looks nice!! We would add this to the the `render-notebooks.R` script? (It's unclear to me currently how you got the getting_started.html to turn out)

I didn't do anything but rerender! Just having the file there will make `rmarkdown::render()` use it automatically. Technically, we should probably add it as an input to the Snakefile, but I didn't do that step yet (just a draft!).

jashapiro:
I added the `_navbar.html` file to the Snakefile (which required separating the `basedir` from the `basefile`), so changes in the navbar will trigger rerenders. What I still don't like about this is that this requires a `_navbar.html` per directory with notebooks in it, which is not a tragedy if we have a single, flat directory for Microarray & RNA-seq, but subdirectories are a pain.

jashapiro:
I have now added the navbar to all sections, so I am marking this ready for review. The main `_navbar.html` file lives in `components/_navbar.html` but it is symlinked to all sections. 

There are comments indicating where to add elements for each new page, but instructions should probably be added to the contribution guidelines as well. I didn't do that yet partly because I couldn't quite figure out where to put it!

jashapiro:
> This looks great! Thanks for figuring this out. My only nitpick comment is it's been bothering me that `Examples` is capitalized in `refine.bio Examples`. Can we switch it to `refine.bio examples` or do people not like that?

I have no preference. Happy to do whatever makes people's eyes happiest.

jaclyn-taroni:
My preference is `refine.bio examples` based on not much to be perfectly honest.

cansavvy:
I added tenative instructions for adding to this navbar file in #187 while I was making changes to that doc anyway.

PR #151, opened 2020-08-06
Add rearranged version of RNA-seq clustering example notebook

cbethell:
## Purpose:
What issue(s) does your PR address?

This PR addresses the issue creating a clustering example with RNA-seq data (#127).

As noted in my [last comment on that issue](https://github.com/AlexsLemonade/refinebio-examples/issues/127#issuecomment-669347715), I followed a plan to reorder/shorten the clustering notebook in closed PR #140:

1. Choose a simpler/less complex dataset (removing the need for the collapse replicates steps).
2. Start by plotting the heat map.
3. Prepare metadata and add annotation bars.
4. Update sentence structure/context after draft PR is filed.

## Questions for reviewers

- Does the dataset seem like to be better suited for use in this example?
- Is the execution of steps still logical? 
- Should I be copying the introduction from the [microarray clustering notebook](https://github.com/AlexsLemonade/refinebio-examples/pull/122/files) into the introduction for this PR?

Note that I removed the numbering strategy and followed the other measures outlined with the contributing guidelines (_I believe_). However, the Snakemake function `snakemake --cores 1` won’t run when I add the notebook to the `Snakefile`, so I may be adding it wrong. I got it to work for the notebook in this PR by replacing what was already there and I didn’t want to commit and overwrite what’s in the repo.

## Pull Request Check List:
* [x] Run a linter
* [x] Set the seed (if applicable)
* [x] Comments and/or documentation up to date
* [x] Double check your paths
* [x] Check headline numbering
* [x] Spell check any Rmd file or md file
* [x] Restart R and run all notebooks fresh and save

PR #165, opened 2020-08-11
Refining the template notebook

cansavvy:
## Purpose:

Trying out a manner of commenting what material is part of the template and what is new. (I'm finding while reviewing #151 that this will be annoying for reviewers So I have some `<!-- comments-->` to try to mark where material is added and these are mean to be left.)

Also we noted some places where the template notebook was not adhering to the style guide (thanks @cbethell )

Also noticed the screenshot `download-analysis-folder` actually wasn't cropped at the informative part so I redid it and updated the src links to that commit. 
## Strategy

There are also two types of comments `<!-- -->` are meant to be left (but I hope they don't confuse the users?) And `{{}}` should be replaced with the proper information.


template/template_example.Rmd
```diff
@@ -62,42 +66,46 @@ if (!dir.exists(results_dir)) {
...
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/download-email.png" width=500>  
 
 It make take a few minutes for the dataset to process.
 You will get an email when it is ready. 
 
 ## About the dataset we are using for this example
+<!-- ADD PATH AND SUMMARY -->
 
-For this example analysis, we will use this [<<BRIEF DATASET NAME>>](https://alexslemonade.github.io/refinebio-examples/ <<RELATIVE PATH TO THIS ANALYSIS FROM REFINEBIO-EXAMPLES>> )..
+For this example analysis, we will use this [<!--BRIEF DATASET NAME-->](https://alexslemonade.github.io/refinebio-examples/ <!--RELATIVE PATH TO THIS ANALYSIS FROM REFINEBIO-EXAMPLES--> )..
```
jashapiro, line 89: Did you test this? I am not sure if you can put a comment inside a markdown url... 

```suggestion
For this example analysis, we will use this [<<BRIEF DATASET NAME>>](https://alexslemonade.github.io/refinebio-examples/ <!--RELATIVE PATH TO THIS ANALYSIS FROM REFINEBIO-EXAMPLES--> )..
```

template/template_example.Rmd
```diff
@@ -108,32 +116,33 @@ Your new analysis folder should contain:
...
-<img src="screenshots/analysis-folder-structure.png" width=400>
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/analysis-folder-structure.png" width=400>
 
 In order for our example here to run without a hitch, we need these files to be in these locations so we've constructed a test to check before we get started with the analysis. 
 Run this chunk to double check that your files are in the right place. 
 
 ```{r}
 # Check if the gene expression matrix file is in the data directory stored as `data_dir`
-file.exists(file.path(data_dir, <<DATA_ACCESSION FILENAME>>))
+file.exists(file.path(data_dir, <!--DATA_ACCESSION FILENAME-->))
```
jashapiro, line 126: I think you want to leave this (and the next one) as it was (pretty sure html comments don't work in code blocks)
```suggestion
file.exists(file.path(data_dir, <<DATA_ACCESSION FILENAME-->))
```

cansavvy: I think i missed some of these because I switch strategies.

jashapiro:
When reviewing #166, I noticed that the use of `<<TEMPLATE>>` sometimes results in rendering oddities as the angle brackets can be interpreted as (broken) HTML tags. I suggest we switch to something that html does not use; perhaps `{{TEMPLATE}}`?

template/template_example.Rmd
```diff
@@ -11,7 +11,7 @@ output:
 
 # Purpose of the analysis:
 
-<<WHAT DOES THIS ANALSIS DO AND WHEN WOULD YOU USE IT?>>
+<!--WHAT DOES THIS ANALYSIS DO AND WHEN WOULD YOU USE IT?-->
```
cbethell, line 14: So the idea here would be to add context based on this prompt but also leave this prompt in the notebook above the context once added?

cansavvy: Yes. That's the idea. That way when we review things we know where to expect added content.

cbethell: Gotcha 👍

Issue #163, opened 2020-08-11
[Discussion] Get rid of subdirectory folders? 


cansavvy:
Some discussion here about READMEs: https://github.com/AlexsLemonade/refinebio-examples/issues/141#issuecomment-671995996 and here about `_navbar.html` files: https://github.com/AlexsLemonade/refinebio-examples/issues/118#issuecomment-671982097 suggests we should have a larger discussion about the idea of getting rid of subdirectories (aka 02-microarray contains all notebooks, no `clustering` directory. And one data folder per tech section). 

The goal of "each notebook can be self-contained" and implementation (#146) of also kinda means the subdirectories are less useful than they were when we began. 

This is probably a good time to decide this type of thing before we start adding more analyses.

Originally we kept each notebook in its own folder in case we went with the download zip strategy (See #100).
But now that we have instructions for users to set up their own directory (with some chunks to do it for them and check #146) I think the subdirectories are less necessary.

cansavvy:
The `plots/` and `results/` folders would also have to be split up by dataset subdirectory folders which means we might have the current folder set up strategy in (#146) would be good for offline use, but if someone does end up downloading the whole repository (or when we are working with the whole repository for developing material).

cansavvy:
A plus side is we would only need one "screenshots" folder: https://github.com/AlexsLemonade/refinebio-examples/pull/151#issuecomment-672036546

jashapiro:
> A plus side is we would only need one "screenshots" folder: [#151 (comment)](https://github.com/AlexsLemonade/refinebio-examples/pull/151#issuecomment-672036546)

I am worried about the screenshots folder, as it does violate our "just download the RMD" idea. Can we use an abbreviated set of instructions in the templates that don't include screenshots?

cansavvy:
> > A plus side is we would only need one "screenshots" folder: [#151 (comment)](https://github.com/AlexsLemonade/refinebio-examples/pull/151#issuecomment-672036546)
> 
> I am worried about the screenshots folder, as it does violate our "just download the RMD" idea. Can we use an abbreviated set of instructions in the templates that don't include screenshots?

You can't see the screenshots in the unrendered version anyway which is what a user would be working from anyway.

cansavvy:
>  Can we use an abbreviated set of instructions in the templates that don't include screenshots?

Can you tell me more about what this would look like? The screenshots show users how to download data so this affects (#100)

cansavvy:
[...]

PR #166, opened 2020-08-11
Use URLs for screenshots instead!

cansavvy:
## Purpose:
This PR makes it so the template.Rmd uses URLs for the screenshots. See discussion on #163 

It was in #165 but Josh rightly told me I should split it up. 
https://github.com/AlexsLemonade/refinebio-examples/pull/165#pullrequestreview-465280662

### DIFF
--- a/template/template_example.Rmd
+++ b/template/template_example.Rmd
@@ -69,11 +69,11 @@ Go to this [dataset's page on refine.bio](<<REFINEBIO DATASET PAGE LINK>>).
 
 Click the "Download Now" button on the right side of this screen.  
 
-<img src="screenshots/download-now.png" width=200>  
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/download-now.png" width=200>  
 
 Fill out the pop up window with your email and our Terms and Conditions:  
 
-<img src="screenshots/download-email.png" width=500>  
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/download-email.png" width=500>  
 
 It make take a few minutes for the dataset to process.
 You will get an email when it is ready. 
@@ -90,7 +90,7 @@ Refine.bio will send you a download button in the email when it is ready.
 Follow the prompt to download a zip file that has a name with a series of letters and numbers and ends in `.zip`. 
 Double clicking should unzip this for you and create a folder of the same name.  
 
-<img src="screenshots/download-folder-structure.png" width=400> 
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/download-folder-structure.png" width=400> 
 
 For more details on the contents of this folder see [these docs on refine.bio](http://docs.refine.bio/en/latest/main_text.html#rna-seq-sample-compendium-download-folder).  
 
@@ -108,7 +108,7 @@ Your new analysis folder should contain:
 
 Your example analysis folder should now look something like this (except with respective experiment accession id and analysis notebook name you are using): 
 
-<img src="screenshots/analysis-folder-structure.png" width=400>
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/analysis-folder-structure.png" width=400> 
 
 In order for our example here to run without a hitch, we need these files to be in these locations so we've constructed a test to check before we get started with the analysis. 
 Run this chunk to double check that your files are in the right place. 
--- a/template/template_example.Rmd
+++ b/template/template_example.Rmd
@@ -80,7 +80,7 @@ You will get an email when it is ready.
 
 ## About the dataset we are using for this example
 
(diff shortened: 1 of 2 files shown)


template/template_example.Rmd
```diff
@@ -69,11 +69,11 @@ Go to this [dataset's page on refine.bio](<<REFINEBIO DATASET PAGE LINK>>).
 
 Click the "Download Now" button on the right side of this screen.  
 
-<img src="screenshots/download-now.png" width=200>  
+<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/master/template/screenshots/download-now.png" width=200>
```
jashapiro, line 72: It's a bit ugly, but I think this is a good place to use permalinks to protect against broken images. The same hash should work for all links here (replacing `master`).
```suggestion
<img src="https://raw.githubusercontent.com/AlexsLemonade/refinebio-examples/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-now.png" width=200>  
```

cansavvy:
I think I'm misunderstanding. If we use permalinks and we updated the screenshots, that wouldn't leave them not updated? 

Why would the master branch links not work?

jashapiro:
If we update the screenshots, yes, we will have to update the links to them. But if some future revision deletes screenshots, everybody who downloaded the old RMD file would find a bunch of broken links, which seems suboptimal to me.

cansavvy:
I agree that that is suboptimal but these links will be in every analysis so any screenshot change will also require 100 more changes in every analysis. I don't like either option.

But I don't see users keeping the getting started section around for a really long time once they've gone through the material and set things up (might very much be wrong about this).

cansavvy:
Which leads us to the larger question of how do we want to deal with any updates to template.Rmd? Can we use Jekyll magic or a special script to update getting started material in all notebooks?

jashapiro:
I think we want to avoid jekyll here because it makes things much less portable (RMD files will not "just work"). We could write some infrastructure to change things automatically, but I don't think it is worth the time. 

Changing permalinks is fairly straightforward; a find and replace across the project files will work, as the hash should be unique. We can leave with the master link for now, but I do think they should be made permalinks before publishing, maybe to a release tag.

cansavvy:
> Changing permalinks is fairly straightforward; a find and replace across the project files will work

Gotcha. That's a good point. Okay. Will change to permalinks.

template/template_example.Rmd
```diff
@@ -69,18 +69,18 @@ Go to this [dataset's page on refine.bio](<<REFINEBIO DATASET PAGE LINK>>).
 
 Click the "Download Now" button on the right side of this screen.  
 
-<img src="screenshots/download-now.png" width=200>  
+<img src="https://github.com/AlexsLemonade/refinebio-examples/blob/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-now.png" width=200>
```
jashapiro, line 72:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-now.png" width=200>  
```

template/template_example.Rmd
```diff
@@ -69,18 +69,18 @@ Go to this [dataset's page on refine.bio](<<REFINEBIO DATASET PAGE LINK>>).
 
 Click the "Download Now" button on the right side of this screen.  
 
-<img src="screenshots/download-now.png" width=200>  
+<img src="https://github.com/AlexsLemonade/refinebio-examples/blob/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-now.png" width=200>  
 
 Fill out the pop up window with your email and our Terms and Conditions:  
 
-<img src="screenshots/download-email.png" width=500>  
+<img src="https://github.com/AlexsLemonade/refinebio-examples/blob/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-email.png" width=500>
```
jashapiro, line 76:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-email.png" width=500>  
```

template/template_example.Rmd
```diff
@@ -90,7 +90,7 @@ Refine.bio will send you a download button in the email when it is ready.
 Follow the prompt to download a zip file that has a name with a series of letters and numbers and ends in `.zip`. 
 Double clicking should unzip this for you and create a folder of the same name.  
 
-<img src="screenshots/download-folder-structure.png" width=400> 
+<img src="https://github.com/AlexsLemonade/refinebio-examples/blob/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-folder-structure.png" width=400>
```
jashapiro, line 93:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/fe91830c24b5d3c289efd9fc2ff3a59611e9b001/template/screenshots/download-folder-structure.png" width=400> 
```

Issue #129, opened 2020-07-09
Create PCA dimension reduction with RNA-seq data


cansavvy:
Related to #110 

Currently we have a microarray example of PCA. We should be able to largely copy the microarray example but switch out the dataset and the accompanying wording. 

Ideally we will use the same dataset as we will use for UMAP RNA-seq example (#128) and ideally the dataset will not have too weird looking of a structure.

This can be done separately from the "getting started" additions to each analysis #116

cansavvy:
I think the dimension reduction module would be a good opportunity to show DESeq2 strategy and non DESeq2 strategy side by side (See #103). Because [`PCAplot`](https://rdrr.io/bioc/DESeq2/man/plotPCA.html) is a handy wrapper function, I think we could do this example as Non-Qn'ed download -> DESEq2 object creation and then in parallel we could do the UMAP example (#128) as a QN'ed example. 

I *think* sticking with the same dataset for both would still be nice.

cansavvy:
I've changed my mind. I think keeping both as the same DESeq2 normalized dataset where we show here how to use a DESeq2 function and how to use a non-DESeq2 function would be a nicely complementary set of examples.  So here, we would use the `DESeq2::PCAplot` with a DESeq2 object and in the UMAP example (#128) we show how to extract the `assay()` data and use a non-DESeq2 function.

cansavvy:
We should also include a comment about the base R PCA functions as an alternative (`prcomp` and/or `princomp`).

cbethell:
As discussed with @cansavvy, the most logical plan for tackling this issue at this time seems to be as follows:

1. Set up - read in `DESeq2` package
2. Import and set up data - find a large non-Qn’ed dataset that can be used for the `PCA` and `UMAP` RNA-seq example notebooks and read that in with its metadata (will post tentative datasets once found, keeping in mind that we want a dataset that has adequate accompanying metadata and that it may be more efficient to borrow a dataset that we have used in one of our previous analyses if it becomes time consuming to search for a new suitable dataset)
3. Create a DESeqDataset object
4. Define minimum counts cutoff
5. Normalize data - Note `vst()` is faster than `rlog()` (so we may switch to using `vst()` for this module is necessary and provide a link and brief explanation for why we would have made this decisionl)
4. Use `plotPCA()` function from DESeq2 to plot the PCA values

[...]

PR #176, opened 2020-08-13
WIP: Add RNA-seq PCA example notebook

cbethell:
## Purpose:

This PR addresses the issue on creating a PCA example notebook with RNA-seq data (#129).

## Strategy

As noted in one of my [comments on that issue](https://github.com/AlexsLemonade/refinebio-examples/issues/129#issuecomment-671506700), I followed a plan in the order as follows:

1. Set up - read in `DESeq2` package
2. Import and set up data - find a large non-Qn’ed dataset that can be used for the `PCA` and `UMAP` RNA-seq example notebooks and read that in with its metadata (will post tentative datasets once found, keeping in mind that we want a dataset that has adequate accompanying metadata and that it may be more efficient to borrow a dataset that we have used in one of our previous analyses if it becomes time consuming to search for a new suitable dataset)
3. Create a DESeqDataset object
4. Define minimum counts cutoff
5. Normalize data - Note `vst()` is faster than `rlog()` (so we may switch to using `vst()` for this module is necessary and provide a link and brief explanation for why we would have made this decisionl)
4. Use `plotPCA()` function from DESeq2 to plot the PCA values

This PR is still a work in progress (context and comments still need to be updated and some analysis steps added) so there are no particular questions for reviewers as yet other than:
- Does the dataset seem like an appropriate dataset to use for this example?
- If it is an appropriate dataset, are there any suggestions on how we should move forward with handling the variables?

Also note that I had to change the formatting of the `Snakefile` in order to get it to successfully run locally (_not completely sure what's going on there_).

### Content checks

### Formatting Checks
* [ ] Spell checked any Rmd file or md file.

### Docker/Snakemake
* [x] In the Docker container, run [snakemake for rendering](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#how-to-re-render-the-notebooks)


03-rnaseq/dimension-reduction/pca_example.Rmd
```diff
@@ -0,0 +1,312 @@
...
+plotPCA(dds_norm,
+  intgroup = c("refinebio_treatment", "refinebio_disease")
+)
+
+# pcaData <- plotPCA(dds_norm, intgroup = c("refinebio_treatment", "refinebio_disease"), returnData = TRUE)
+#
+# ggplot(pcaData, aes(
+#   x = PC1,
+#   y = PC2,
+#   color = factor(refinebio_treatment),
```
cansavvy, line 294: Seems good! But as we recently decided, you can add each of these variables as labels one at a time. 

Side question: does DESeq2 turn these variables into non factors? If no, can we change these to factors before we get to plotting? (Ideally before we get to the DESeq2 object if that is possible).

cbethell: > Side question: does DESeq2 turn these variables into non factors? If no, can we change these to factors before we get to plotting? (Ideally before we get to the DESeq2 object if that is possible).

Good question! I'll look into that and make the appropriate changes.

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+We are going to do some pre-filtering to keep only genes with 10 or more reads total.
+Note that rows represent gene data and the columns represent sample data in our dataset.
+
+```{r}
+# Let's take a look at the distribution of counts first
+
+# Sum up the counts for each gene -- which means we want the sum of counts for each row as the rows represent gene data and the columns represent sample data in our dataset
+count_sums <- data.frame(count = rowSums(counts(dds)))
+
+# Create a density plot using the sums
```
cansavvy, line 259: I know I said I thought this was a good idea to plot this out, but with how peak-y this plot is, I don't think it tells us much. Can we remove it from here (and maybe from clustering heatmap too). Sorry to change my mind on it. I usually do like plotting things before making cutoffs, but in execution I don't think it actually helps us. Sorry to double back!

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+  y = PC2,
+  color = refinebio_treatment,
+  shape = refinebio_disease
+)) +
+  geom_point(size = 3)
+
+# Close the PNG file:
+dev.off()
+```
+
```
cansavvy, line 359: I think we lost the "further resources" section and there's some super cool articles we should put here from training-modules (skip the ones that are scRNA-seq specific though) https://github.com/AlexsLemonade/training-modules/blob/3dbc6f3f53c680ec6aa2f513851c1cd4635cc31c/scRNA-seq/05-dimension_reduction_scRNA-seq.Rmd#L422

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+```{r}
+plotPCA(dds_norm,
+  intgroup = "refinebio_treatment"
+)
+```
+
+In this chunk, we are going to add a variable to our plot for annotation.
+
+The variables `refinebio_treatment` and `refinebio_disease` are given to the `plotPCA()`
+due to the hypothesis of the experiment stating "that differences in tumor transcriptional programs may reflect differential responses to androgen deprivation therapy (ADT) and subsequent metastasis".
```
cansavvy, line 305: Can you paraphrase instead of quote directly?

cansavvy: And probably put the citation again.

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+
+```{r}
+plotPCA(dds_norm,
+  intgroup = c("refinebio_treatment", "refinebio_disease")
```
cansavvy, line 309: I would put a comment to highlight the difference here.

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+)
+```
+
+It's hard to distinguish the different `refinebio_treatment` versus `refinebio_disease` in the plot above.
+
+Let's use the `ggplot2` package functionality to customize our plot further.
+
+```{r}
+# We first have to save the results of the `plotPCA` function for use with `ggplot2`
+pcaData <-
```
cansavvy, line 319: Can we make these into their own two chunks? And we can add some preview words before each step?

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+    intgroup = c("refinebio_treatment", "refinebio_disease"),
+    returnData = TRUE # This argument tells R to return the PCA values
+  )
+
+# Now let's plot using `ggplot2` functionality
+ggplot(pcaData, aes(
+  x = PC1,
+  y = PC2,
+  color = refinebio_treatment,
+  shape = refinebio_disease
```
cansavvy, line 331: Can you comment here to, to highlight this is the key thing that makes the difference?

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+  geom_point(size = 3)
+```
+
+## Save annotated PCA plot as a PNG
+
+You can easily switch this to save to a JPEG or TIFF by changing the function and file name within the function to the respective file suffix.
+
+```{r}
+# Open a PNG file
+png(file.path(
```
cansavvy, line 342: We should use `ggsave()` here. (It can do PNGs if that was the concern?)

cansavvy: Also may want to specifically name the plot (aka don't depend on default of `last_plot()`) since there's multiple plots and people may get confused which one is being save).

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
```diff
@@ -0,0 +1,368 @@
...
+# PCA Visualization - RNA-seq
+
+## Install libraries
+
+See our getting started with [package installation](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#what-you-need-to-install) for a list of the other software you will need, as well as more tips and resources. 
+
+In this analysis, we will be using the R package [`DESeq2`](http://bioconductor.org/packages/devel/bioc/vignettes/DESeq2/inst/doc/DESeq2.html) [@DESeq2] for normalization and production of PCA values and the R package [`ggplot2`](http://r-statistics.co/Complete-Ggplot2-Tutorial-Part1-With-R-Code.html) [@ggplot2] for plotting the PCA values.
+
+```{r}
+if (!("ggplot2" %in% installed.packages())) {
```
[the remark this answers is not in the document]
cbethell: Gotcha, no I believe your original idea makes sense here.

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
cbethell: I believe I implemented this in the last commit.

03-rnaseq/dimension_reduction_rnaseq_01_pca.Rmd
cbethell: _Believe_ I added such comment here in the second to last commit.

Issue #173, opened 2020-08-13
Update microarray analyses notebooks


cansavvy:
Although the content itself exists, our microarray analyses will need updating. 

Each analysis notebook will: 
- Have the getting started template material added to it(# 146)
- Be updated to adhere to the style guide (#148)
- Have its dataset added to the S3 bucket. (See https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-s3-bucket)
- Be added to the Snakefile and rendered with snakemake on the Docker container. 
- Be evaluated for any content changes that we may have updated thoughts on (this one we will have to determine analysis by analysis)

As of now here are the microarray notebooks that will need to be updated. We can break out each of these into their own issue to be discussed specifically. 

- [x] clustering_microarray_01_heatmap.Rmd #181 
- [x] differential_expression_microarray_01.Rmd #182 
- [x] dimension_reduction_microarray_01_pca.Rmd #196 
- [x] dimension_reduction_microarray_02_umap.Rmd #195 
- [ ] ortholog_mapping_microarray_01.Rmd #202 
<del> - [ ] pathway_analysis_microarray_01_ortholog_mapping_kegg.Rmd <\del> We likely don't need a separate orthlog mapping example now. Particularly if we use `msigdbr`
- [x] pathway_analysis_microarray_02_ora_with_webgestaltr.Rmd #206 
- [ ] pathway_analysis_microarray_03_qusage_meta_analysis.Rmd #215 
- [ ] pathway_analysis_microarray_04_qusage_replicate_vignette.Rmd #215
- [ ] pathway_analysis_microarray_05_qusage_single_dataset.Rmd #215 
- [ ] pathway_analysis_microarray_06_ssgsea.Rmd

cansavvy:
As we revamp the notebooks, we should also keep in mind what are recurring questions/concepts that should be covered in the intro to microarray document #124. Questions that are tangential to the main analyses are the kind we'd link out to the intro document.  (Same as discussed for the intro to RNA-seq document).

Issue #162, opened 2020-08-11
Set up dataset storage for development purposes


cansavvy:
From: https://github.com/AlexsLemonade/refinebio-examples/issues/154#issuecomment-670519342

Where do we want to store data for development purposes? 

> One general question before I disappear for a few days - could the scope then be run the `Snakefile` upon commit and push back to GitHub? That would probably be pretty handy, but some problems with that are run time, environment, and all the data files. (These may be prohibitive problems!) 

> To the last point -- data files -- solving that issue such that there is a centralized place _we_ store the datasets for development could be pretty useful. My impression is that the workflow now if you are reviewing someone's PR or you have to work on a module you haven't worked on before, you have to go to refine.bio yourself and wait to get the dataset download. (Please tell me if that impression is wrong.) This seems like this could be a significant point of friction during development once we have the number/scope of modules that I think we will have by the time all is said and done. We also expect some development in the materials before we have the intro stuff cemented, so I think considering how can we make doubling back easier if you're not the person who wrote the analytical code for a notebook is worth a think. (S3?)

cansavvy:
My oversight on whether the RNA-seq data was QN'ed 
or not https://github.com/AlexsLemonade/refinebio-examples/issues/136#issuecomment-673489252 also makes me realize that we should have a storage system for the data so we can keep track of it and know exactly what everybody is using. 

I think we should also stop including the datasets in GitHub so we make sure we are always using the up to date dataset.

cansavvy:
I think having all the data in less folders will also help
I think I'm going to break this discussion out into actionable 3 issues: 

- Set up an S3 bucket to store the datasets on 
- Make a script that downloads the data
- Remove all metadata and datasets stored on GitHub. 

And each should contain bits about adding instructions about all of this to CONTRIBUTING.md

cansavvy:
This is all wrapped up!

Issue #112, opened 2020-07-07
Restructure validate-differential-expression


cansavvy:
Right now, validate-differential expression is made up of two notebooks. One that sets up a practice dataset to compare to, and the second that does the actual comparison. We have to think about how to restructure this module and it's set up notebook with the impending restructuring of the refinebio-examples #106

cansavvy:
At this point, the validate-differential-expression structures is a bit confusing because it involves two notebooks and one of which isn't actually needed from the users point of view: 

validate_differential_expression_adv_topics_00_author_de.Rmd - sets up what kind of results a user might want to compare to refinebio data. 
validate_differential_expression_adv_topics_01.Rmd - Does the actual comparison. 

A user who is actually interested in using this notebook, would only need the files from the "author_de" notebook to run the examples but then they would really just replace our author_de files with their own files. So I'm not sure how to make this more clear as well as set up a better file format so perhaps the "author_de" notebook is a bit hidden since its not meant to be an example on it own.

cansavvy:
I was thinking about this, I think the easiest solution is to turn `validate_differential_expression_adv_topics_00_author_de.Rmd ` into a script that is `source()` called by `validate_differential_expression_adv_topics_01.Rmd`. We make it clear to users that this step can be deleted when they are importing their own dataset for comparison. 

Related to this, I think the `Using this example with your own data` section will need to be much more extensive in this situation.

PR #178, opened 2020-08-14
Remove analysis subdirectories and adjust accordingly

### LINKED ISSUE #170: Remove all GitHub stored datasets
Currently, every example dataset is stored on GitHub.

But for users, we've added instructions for how to download the data from refine.bio 
And for our own development purposes we will download from S3 so we can keep track of what datasets we are all using #162 

We will need to add to the gitignore. 

This step might want to be done at the same time as we remove the subdirectories, if we are deciding to do that.

### LINKED ISSUE #171: Remove subdirectory folders in each section
From #163 

I think we should move forward with this idea. But here's how I think we should do it: 
We need to a notebook organization method that is sans folders. 

### Notebook naming conventions:
I have two ideas that can be used together or we just choose one over the other. 

1) Number the notebooks in a way that those of the same module are kept together
Because we are making our notebooks self-contained, the order is not 100% relevant but still matters. So the order of the notebooks could be from 'most foundational knowledge' and/or 'least amount of background knowledge required' is first.

2) Each notebook has a prefix if it comes in a group of notebooks. So all `differential_expression_`, `pathway_analysis_` Currently most modules contain one notebook, some two, except pathway analysis which has 6. I do find this quite ugly however and it will make notebook names quite long. 

### Data organization 
This is more straightforward and is the perk of this new folder structure because duplicate datasets (for modules that may both use the same dataset) will not need to be kept. (not sure how much this is an issue currently)

Example: `02-microarray/data/experiment_accession/`
- This is certainly more organized but also won't allow us to run notebooks as the users would be, because we don't instruct them to keep an `experiment_accession` folder. 

     OR 

Example:  `02-microarray/data/`
- the admittedly less aesthetically pleasing idea (but it won't affect our users because they won't see it)

### Intro to "module"s

Any pertinent background information in the current READMEs would go in new "intro_to_pathway_analysis.Rmd" type files. (noting that all installation info and usage info can be dropped because its all covered in the "getting started per analysis" section.) Old READMEs are deleted. 

### Plots and results folders 

This might look like a bit of a mess unless there's some kind of symlink magic we can use to keep plots/results in their own folders. But we don't want to change the code in the Rmd itself. But we would like the files to be organized.

cansavvy:
## Purpose:
Closes #171 
Closes #170 

Implementing the strategy planned out in 171 where all subdirectories are deleted, and notebooks all subscribe to this format: `<analysis_module>_<section/tech>_<notebook#>_<name_of_analysis>.Rmd`

Related to this, plots and results are now saved with the experiment accession id first. (this is described in the CONTRIBUTING.md file). 

## Strategy
- Moved `.Rmd` files from out of their subdirectories and deleted them. (We can go to the v1 release to see the READMEs for ref if needed later). 
- Added `data` `plots` and `results` to the gitignore. Data is not stored here now. (It will shortly enough be available for easy download from S3 #162)
- Renamed everything according to the new naming conventions we discussed. 
- Added instructions for the new naming conventions to CONTRIBUTING.md
- I updated clustering rna-seq Rmd. I fixed its link and directory references and updated its line in the Snakefile. 
- Lastly, I re-ran snakemake which only really affected the clustering rna-seq html file. 

- Technically all analyses except clustering rna-seq Rmd are very much out of date, so I wouldn't bother with any of the other Rmds. They will be more completely revamped and otherwise made to comply with the CONTRIBUTING.md in good time: #173 #137 #112. 
- GitHub hates me for all this restructuring so ignore it when it thinks there are new notebooks. There are no new notebooks. 
- Unsure if my gitignore regex nonsense is the most efficient? `*/plots/*` ?


Snakefile
```diff
@@ -1,7 +1,7 @@
 rule target:
     input: 
         "01-getting-started/getting-started.html",
-	  "03-rnaseq/clustering/clustering_heatmap_example.html" 
+	    "03-rnaseq/clustering_rnaseq_01_heatmap.html"
```
jashapiro, line 4:
```suggestion
	  "03-rnaseq/clustering_rnaseq_01_heatmap.html" 
```

PR #187, opened 2020-08-18
Make some updates to CONTRIBUTING.md

### LINKED ISSUE #185: Docker hub: latest doesn't mean `latest`
@jashapiro found that the `latest` tag doesn't actually have default behavior that means its the `latest` version of the docker container when you pull it from the Docker hub. 

This is not what I thought the behavior for Docker hub was, so we need to switch strategies for Docker hub on this repo. 
I need to change this on CONTRIBUTING.md. 

We will make things simple. No tags, just keep pushing updated versions.

### LINKED ISSUE #186: Discussion: Consolidate "new notebook" instructions
When adding a new example notebook, there are now a few steps that need to be followed in addition to (before?) starting the main work of filling its contents.

- A file name needs to be chosen
- Add the expected output html file to snakemake
- Add a link to the html file to `_navbar.html` (with a short title)
- Data may need to be added to S3
- other stuff I am missing?

Should these be added to the PR template? Or maybe just a consolidated section of CONTRIBUTING.md (or both?).

cansavvy:
## Purpose:
Closes #185 
Closes #186 

## Strategy
- I made a reorganized and edited CONTRIBUTING.md to try to address the group things by concept but then have a `add a new analysis` that summarizes and links out to more information for each step. 
- I removed the previous Docker strategy and now we just push the image no tag. 

-  Does the `Add a new analysis` section fit what was asked for in #186 ?


CONTRIBUTING.md
```diff
@@ -202,6 +230,13 @@ Chunks preferably shouldn't be named.
 If we do end up using [bookdown](https://bookdown.org/yihui/bookdown/) at some point, repetitive chunk names like `import data` will cause havoc.
 Plus its just another thing to have to keep track of.
 
+#### Code Style
+
+These analyses follow the [Google R Style Guide](http://web.stanford.edu/class/cs109l/unrestricted/resources/google-style.html) which is based on the tidyverse style guide.
```
jaclyn-taroni, line 235: I believe `styler` is enforcing the tidyverse guide, so we should probably use the tidyverse guide and link that here? Or, at this point, these guides may be sufficiently similar that it doesn't matter.

cansavvy: > I believe `styler` is enforcing the tidyverse guide, so we should probably use the tidyverse guide and link that here? Or, at this point, these guides may be sufficiently similar that it doesn't matter.

This google link starts off with linking to tidyverse anyway so it's almost like both are linked. I just like the Google layout better. But they do have so much overlap by now, it doesn't matter much.

jaclyn-taroni: That link is the outdated guide. I think you want this one: https://google.github.io/styleguide/Rguide.html 

Though that guide includes:

> Google prefers identifying functions with `BigCamelCase` to clearly distinguish them from other objects.

Is that the plan? (I think custom functions are unlikely to come up often, to be fair.)

jashapiro: They also disagree on `return()` statements, though here I am a bit more on Google's side. I like explicit returns. https://style.tidyverse.org/functions.html#return

jashapiro: Odd that the google link isn't a full style guide anymore.

jaclyn-taroni: > They also disagree on `return()` statements, though here I am a bit more on Google's side. I like explicit returns. 

Agree with this!

cansavvy: I don't know if custom functions will come up (maybe in Advanced Topics) but I'm fine with following Google still. (But I'll update the link out here).

cansavvy:
> I think the content looks fine here, but I think the overall order might benefit from updating. Here is an order that makes sense to me, but by no means the only one. My general idea is to promote the things that are most universal to the top, and leave specific procedures to later...
> 
> * General Guidelines for analyses notebooks (including citationl and spell check)
> * Docker for refine-bio
> * Download the datasets
> * Add a new analysis
>   
>   * Setting up a new analysis file
> * Rendering notebooks

I'm all for rearranging the order for whatever is most helpful. I did some overall reordering in this PR, but kinda kept flip flopping on what made sense. I do like having Docker first because setting up the docker should happen before you edit notebooks. And the citations and spell checking has to do with editing notebooks. 

I do like the idea of putting the rendering notebooks section after the `Add a new analysis`.

cansavvy:
I still kept Docker as the first thing. But I otherwise did some header changes and reorganizattion.

CONTRIBUTING.md
```diff
@@ -175,17 +138,19 @@ Test that the files you've uploaded succesfully download by running the `downloa
 script.
 If you run the script and it says `Access Denied` you may have missed step 8, but you can go back to the file and click the `Make it Public` button and try testing your download again.
 
-### General guidelines for analyses notebooks
+### Guidelines for analyses notebooks
```
jashapiro, line 141: You may hate me for this, but Can we bump this and all of its children up a level? (also not spelling change)
```suggestion
## Guidelines for analysis notebooks
```

PR #197, opened 2020-08-24
Separate template changes from DE PCA to its own PR

cansavvy:
## Purpose:
There were some minor changes needed to be made to the template.Rmd file on #183 
I don't want those to get held up by the DE items. So I'm copying them over here so they can get merged more quickly and hopefully be made more visible. 

## Strategy
These are mostly tiny edits identified by myself or @envest over on #183


template/template_example.Rmd
```diff
@@ -188,8 +190,7 @@ metadata <- readr::read_tsv(file.path(data_dir, # Replace with path to your meta
 # Read in data TSV file
 df <- readr::read_tsv(file.path(data_dir, # Replace with path to your data file
                                 {{DATA_ACCESSION FILENAME}} # Replace with the name of your data file
-                                )) %>%
-  tibble::column_to_rownames('Gene')
+                                ))
 ```
 
 Let's ensure that the metadata and data are in the same sample order.
```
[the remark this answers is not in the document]
cansavvy: I think this needs some rearranging. Gotta put some more thought into which is more annoying/better. Do we use the negative indexing in the select statement or do we go straight to row names? Ultimately I think we can decide for each module, but I should think more about what we want in the template.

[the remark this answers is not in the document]
cansavvy: But you are correct that what is here is not what we want.

template/template_example.Rmd
```diff
@@ -108,26 +105,31 @@ For more details on the contents of this folder see [these docs on refine.bio](h
...
     - The gene expression  
     - The metadata TSV  
-
+- A folder for `plots` (currently empty)  
+- A folder for `results` (currently empty)  
+    
 Your example analysis folder should now look something like this (except with respective experiment accession id and analysis notebook name you are using): 
 
-<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/b78046845763272d0aa65a077b6a3e46977d5342/template/screenshots/analysis-folder-structure.png" width=400>
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/dcce9a49246a15bfc7bb94a50a082c300c78074f/template/screenshots/analysis-folder-structure.png" width=400>
```
cbethell, line 124:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/analysis-folder-structure.png" width=400>
```

This would need to be updated to use the  updated screenshot you added to this PR.

cansavvy: I don't know if changing the commit actually matters since the screenshots have not changed, but I can update them if you think its better. :+1:

cbethell: Ah the older version of the screenshot I have locally does not include the `plots` and `results` folders so I thought that was the part being updated in this PR.

cansavvy: It's a commit from #197 I changed all the commit ids here since that is probably less messy.

PR #203, opened 2020-09-01
Proof of concept: Add a way to skip to the analysis code section

jaclyn-taroni:
## Purpose

This is an idea that came up when I was reviewing #183.

Having a way to skip directly to the analysis code from the top of the notebook (e.g., not using the floating TOC) seems like something that could be handy when viewing the notebooks online. 

Here's a preview of what this looks like in action: https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/199ce6467a0eadeee0e6d9d38f0c41d9a512bf87/03-rnaseq/differential-expression_rnaseq_01.html

## Strategy

Add a link to the header where the analysis starts.  I originally had grand plans to use `<details>` to make the how to run the analysis section collapsible but that appears to only work around code chunks without additional packages, etc.

I am a little surprised to see version changes in the session info. I did the following:

```
docker pull ccdl/refinebio-examples
docker run --mount type=bind,target=/home/rstudio,source=$PWD -e PASSWORD=<my-excellent-password> -p 8787:8787 ccdl/refinebio-examples
```


jashapiro:
I like this! Needs more emoji though. ⚡️🚀

> I am a little surprised to see version changes in the session info. I did the following:

My guess is that whoever ran this before you might have rebuilt the docker image, so it was ahead of the version on dockerhub. The cran mirror for the rocker 4.0.2 container is not yet frozen, as 4.0.2 is still the current version of R. It should freeze when 4.0.3 is released.

cansavvy:
There were recent changes (sometime yesterday afternoon) I pushed to the docker image. So depending on the timing of you pulling the image and me pushing an updated one, I *think* that makes sense.

jaclyn-taroni:
I pulled it right before I made these changes this morning.

cansavvy:
Apparently my rebuild of the image did this, and I too get these same little decimal changes in the versions too.

jaclyn-taroni:
Just something to keep an eye on I suppose - we probably want to tie a GH release to a Docker tag in the future.

jaclyn-taroni:
Feel free to close this now that you have #204 @cansavvy !

Issue #126, opened 2020-07-09
Write introduction to advanced topics section header Rmd


cansavvy:
Related to #123 

This `introduction_to_advanced_topics.Rmd` is blank. Unsure what this introduction should include. Table of Contents makes sense, not sure what else, it will probably depend on what ends up being in this section.

PR #207, opened 2020-09-02
Add skip buttons!

### LINKED ISSUE #204: Add a "Skip" button to all the modules and the template
See the proof of concept PR for what this looks like: 
https://github.com/AlexsLemonade/refinebio-examples/pull/203

This needs to be added to the template with a `{{type deal}}`, and all the currently update modules, and everything needs to be re-ran.

cansavvy:
## Purpose:
Closes #204 

## Strategy
I added skip buttons everywhere including the template, and I reran everything using snakemake on the docker container. 

Did I miss any of the currently updated modules? 

Note we already discussed some of the docker related software version changes you see here: https://github.com/AlexsLemonade/refinebio-examples/pull/203#issuecomment-684921342


cansavvy:
Double checked, we have 5 current modules and the template, x2 for the Rmd and html each, and apparently the advanced-topics.html had some navbar updates, so that brought us to 13 file changes. This seems right. Except I wasn't expecting advanced-topics.html to change, but it hasn't been written yet anyway (#126) so I'm not too concerned about it for now.

jaclyn-taroni:
Maybe there's an easy enough fix to this... when I click **Jump to the analysis code** I am taken to the install libraries header which I find a little bit _odd_ and/or disorienting. I would expect the behavior to be - I click on **Jump to analysis code** and it takes me to number 4 (e.g., "Differential Expression", "Clustering Heatmap - RNA-seq").

Is there a way to do that without increasing the maintenance burden by linking with `` which is what was in the proof of concept in #203 originally?

cansavvy:
> Maybe there's an easy enough fix to this... when I click **Jump to the analysis code** I am taken to the install libraries header which I find a little bit _odd_ and/or disorienting. I would expect the behavior to be - I click on **Jump to analysis code** and it takes me to number 4 (e.g., "Differential Expression", "Clustering Heatmap - RNA-seq").
> 
> Is there a way to do that without increasing the maintenance burden by linking with `` which is what was in the proof of concept in #203 originally?

The first step of the analysis specific steps is package installation so that's why it looks that way (if I understand what you are describing).  Aka, the first step in `4 - Differential Expression` is `Install packages`.

cansavvy:
Ah, but I see for some reason, even though I have the link part above the main `Differential Expression` title, it still skips to `Install packages`. That isn't what I expected for its behavior based on where I put it.

jashapiro:
I think the reason it is jumping "too far" is that it is putting the anchor at the top of the page, but not accounting for the top navbar. There is probably some fancy javascript that could fix this, which I will investigate shortly: the TOC links work as they should, so it should be solveable.

02-microarray/dimension-reduction_microarray_01_pca.Rmd
```diff
@@ -143,6 +145,8 @@ From here you can customize this analysis example to fit your own scientific que
 
 ***
 
+<!-- Do not delete this line --> <a name="analysis"></a>
```
jashapiro, line 148: This should fix the anchor link issue. Only putting it here, but will obviously need to be copied to other files.

```suggestion
<!-- Do not delete this line --> <a name="analysis" style="padding-top:56px;margin-top:-56px;">&nbsp;</a>
```

cansavvy: Done!

jaclyn-taroni:
@cansavvy my guess is that you copied from the suggestion to the other notebooks in https://github.com/AlexsLemonade/refinebio-examples/pull/207/commits/d55ea366d1ddba82ab23e330c3443f58c1aa906c because now there is a random `+` in the HTML file.

Screen Shot 2020-09-02 at 3 23 51 PM

jaclyn-taroni (approved):
LGTM - I checked this branch out locally and looked at the clustering HTML  (this is how I found the `+` and "jump too far" issues previously)

PR #217, opened 2020-09-15
Be consistent about `library(ggplot2)` 

### LINKED ISSUE #211: Be consistent about `library(ggplot2)`
I think so far I might have been inconsistent and sometimes use `library(ggplot2)` and sometimes `ggplot2::` all the way through. 

For most packages, I am okay with us deciding on a case-by-case basis and sort of naturally showing `::` and `library()` 'out in the wild'. I also do NOT like loading the entire `tidyverse` because of how many things that includes. 

However, I think `ggplot2` is a special case because one ggplot requires you repeat `ggplot2::` a zillion times and plots are something I suspect users will often want to play around with the most and they may find the number of times they have to write (or forgot to write) `ggplot2::` burdensome. 

Because of this, I think we should change any notebooks we have already done that used the `ggplot2::` strategy, and change them to `library(ggplot2)`.

cansavvy:
## Purpose:
Closes #211 

Whether from luck or I didn't realize cbethell and I were already operating on this notion, all the currently updated notebooks do use `library(ggplot2)`. In the project, I searched for `ggplot2::` and only 4 notebooks that are not yet updated. I removed the `ggplot2::`'s anyway, but I also added a bullet point to CONTRIBUTING.md about this being a thing from now on. 

I re-rendered things with snakemake and everything was just fine. 

I also snuck in a link fix in _navbar.html for the clustering rna-seq file which wasn't working.

### DIFF
--- a/02-microarray/pathway_analysis_microarray_06_ssgsea.Rmd
+++ b/02-microarray/pathway_analysis_microarray_06_ssgsea.Rmd
@@ -321,24 +321,24 @@ comp_coag_df <- metadata_df %>%
 
 ```{r fig.height=11, fig.width=8.5}
 comp_coag_df %>% 
-  ggplot2::ggplot(ggplot2::aes(x = condition, 
+  ggplot(aes(x = condition, 
                                y = ssgsea_score)) +
-  ggplot2::geom_boxplot() +
-  ggplot2::facet_grid(strain ~ refinebio_specimen_part) +
-  ggplot2::labs(y = "KEGG Complement and Coagulation Cascades") +
-  ggplot2::theme_bw() +
+  geom_boxplot() +
+  facet_grid(strain ~ refinebio_specimen_part) +
+  labs(y = "KEGG Complement and Coagulation Cascades") +
+  theme_bw() +
   # x-axis text at a 45 degree angle to increase readability
-  ggplot2::theme(axis.text.x = ggplot2::element_text(angle = 45, hjust = 1),
-                 text = ggplot2::element_text(size = 15))
+  theme(axis.text.x = element_text(angle = 45, hjust = 1),
+                 text = element_text(size = 15))
 ```
 
 ```{r}
 # saving to file
-ggplot2::ggsave(file.path(plots_dir,
+ggsave(file.path(plots_dir,
                           "GSE75574_ssgsea_comp_coag_facet.pdf" # Replace with relevant output plot name
                           ),
-                plot = ggplot2::last_plot() + 
-                  ggplot2::theme( text = ggplot2::element_text(size = 10)))
+                plot = last_plot() + 
+                  theme( text = element_text(size = 10)))
 ```
 
 Looks like there may be some differences between tissues in this pathway, with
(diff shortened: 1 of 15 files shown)


02-microarray/pathway_analysis_microarray_06_ssgsea.Rmd
```diff
@@ -321,24 +321,24 @@ comp_coag_df <- metadata_df %>%
```
cbethell, line 321: Looks like `library(ggplot2)` needs to be added at the top of this notebook now as well.

04-advanced-topics/quantile_normalize_own_data_adv_topics_01.Rmd
```diff
@@ -175,10 +175,10 @@ exprs_long_df <- expression_df %>%
```
cbethell, line 175: Looks like `library(ggplot2)` also needs to be at the top of this notebook.

04-advanced-topics/validate_differential_expression_adv_topics_00_author_de.Rmd
```diff
@@ -180,10 +180,10 @@ test.probe <- author.df %>%
 # Make a dataframe with the group information
```
cbethell, line 180: Same comment re explicitly attaching the `ggplot2` library at the top of this notebook.

04-advanced-topics/validate_differential_expression_adv_topics_01.Rmd
```diff
@@ -391,10 +391,10 @@ test.gene <- refine.bio.df %>%
 box.plot <- data.frame(test.gene, subgroup = metadata$subgroup) %>%
```
cbethell, line 391: Same comment here re explicitly attaching the ggplot2 library at the top of this notebook.

cbethell (commented):
Looks good @cansavvy! I just have one pretty redundant comment re adding `library(ggplot2)` at the top of each notebook you made changes to in this PR.

02-microarray/pathway_analysis_microarray_06_ssgsea.Rmd
cansavvy: These notebooks haven't been updated at all, so I'll leave them as is and when we address #173 we can address that too.

04-advanced-topics/quantile_normalize_own_data_adv_topics_01.Rmd
cansavvy: Same comment as above.

04-advanced-topics/validate_differential_expression_adv_topics_01.Rmd
cansavvy: Same.

04-advanced-topics/validate_differential_expression_adv_topics_00_author_de.Rmd
cansavvy: Same.

cansavvy (commented):
Should have been more clear: It was easiest to search and replace everything in the project but I wasn't planning on making additional fixes to the notebooks that are not yet updated at all. Does that make sense? I figure we can do that all at once for each module when we update them.

cbethell (approved):
> Should have been more clear: It was easiest to search and replace everything in the project but I wasn't planning on making additional fixes to the notebooks that are not yet updated at all. Does that make sense? I figure we can do that all at once for each module when we update them.

Ah okay, yes that makes sense. In that case, this LGTM 🚀

PR #212, opened 2020-09-10
Add microarray gene ID conversion example

cbethell:
## Purpose:
What issue(s) does your PR address?

This PR closes #113  (the issue on creating a microarray gene ID conversion example notebook).

## Strategy
What was your strategy for this new or edited analysis?

I followed the comments on issue #113 and found a new (larger than n = 4 samples) mouse microarray dataset.

The dataset used in this paper is a [mouse glioma cancer stem cell dataset](Cancer Stem Cells Are Enriched In The Side-Population Cells In A Mouse Model Of Glioma) with n = 15 samples (which has been uploaded to the S3 bucket for testing/review).

I then copied the analysis steps from the RNA-seq gene ID conversion notebook, but changed the annotation package being loaded in to that relevant to the mouse genome.

## Concerns/Questions for reviewers:
What things should reviewers look out for?

- Is the current workflow missing any vital steps?
- Do the steps under the `Explore mapped data frame` section seem helpful? Are the added steps clear? Is there anything that should be added/removed from this section?

### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable)

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [x] Spell checked any Rmd file or md file.
* [x] Comments and documentation are up to date.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar).
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds.
* [x] In the Docker container, [snakemake was run for rendering](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#how-to-re-render-the-notebooks).

PR #218, opened 2020-09-15
Update RNA-seq gene ID conversion example

### LINKED ISSUE #213: Update RNA-seq gene id conversion module
This needs to be brought up to speed with the CONTRIBUTING.md guidelines

> - Have the getting started template material added to it(# 146)
> - Be updated to adhere to the style guide (#148)
> - Have its dataset added to the S3 bucket. (See https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-s3-bucket)
> - Be added to the Snakefile and rendered with snakemake on the Docker container.
> - Be evaluated for any content changes that we may have updated thoughts on (this one we will have to determine analysis by analysis)

- I think we should add a few quick steps in the vein of "look at your data" that explore how the mapping from gene identifier conversions went. This can largely be parallel to #113
- As discussed on #174, we want this module to be parallel but model different aspects than its microarray counterpart. This means at least a different species and different gene identifier, but we may want to think about if there's something else we should model. 
- Since this is a cross-technology module (#174) we should probably have a link to the microarray notebook and tell the users that even thought its a different technology, the steps are cross-tech compatible and they can reference the other notebook for a second example.

cbethell:
## Purpose:

This PR closes #213  (the issue on updating the RNA-seq gene ID conversion example notebook).

## Strategy

I followed the comments on issue #213 and copied what's in open PR #212 (the addition of the microarray gene ID conversion example) to update the RNA-seq gene ID conversion example notebook in this PR.

I copied what was in PR #212 up until the rerun of `mapIds()` to produce the filtered data frame with only 1:1 mappings. We will likely want to provide an example of a different way users can handle the data at this point.
So far in this PR, I took the route of finding the count of multi-mapped values for each Ensembl ID and saving all of the values found, along with the RNA-seq data, in the final file.

I also made a note in the notebook that I am not sure if `UNIPROT` is the right identifier to use here, but using `REFSEQ` made the example pretty memory intensive. We may want to try a few different identifiers here.

- Which identifier should we use in this example?
- What final dataset would we like to end up with (i.e. one filtered for only 1:1 mappings or one with all of the mappings and the count of multi-mappings for each Ensembl ID)?
- Should we be adding the individual `org.Dr.eg.db` organism package to the Dockerfile?

### Content checks

### Formatting Checks
* [x] Spell checked any Rmd file or md file.

### Add datasets to S3


03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,316 @@
...
+output:   
+  html_notebook:
+    toc: true
+    toc_float: true
+    number_sections: true
+---
+
+# Purpose of this analysis
+
+The purpose of this notebook is to provide an example of mapping gene IDs for RNA-seq data obtained from refine.bio using `AnnotationDbi` packages.
```
cansavvy, line 14: Probably should have a citation here.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,316 @@
...
+Your example analysis folder should now look something like this (except with respective experiment accession ID and analysis notebook name you are using): 
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/analysis-folder-structure.png" width=400>
+
+In order for our example here to run without a hitch, we need these files to be in these locations so we've constructed a test to check before we get started with the analysis. 
+Run this chunk to double check that your files are in the right place. 
+
+```{r}
+# Define the file path to the data directory
+data_dir <- file.path("data", "SRP056136")
```
cansavvy, line 119: I think there were spaces in between these lines?

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,316 @@
...
+# Check if this is in the same order
+all.equal(colnames(df), metadata$refinebio_accession_code)
+
+# Bring back the "Gene" column in preparation for mapping
+df <- df %>%
+  tibble::rownames_to_column("Gene")
+```
+
+
+## Map Ensembl IDs to Uniprot IDs
```
cansavvy, line 211: IDK if mapping to protein ids is a great idea. In general RNA and protein are pretty poorly correlated, so we may not want to encourage this science wise, though it is a thing you can do. 

What about RefSeq ids or another RNA-based id? 

Do the RefSeq IDs only have `NM_` accessions? and not `XM_` or etc?

cansavvy (commented):
Some quick comments for now!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jaclyn-taroni: Let's do Entrez IDs - looks like #212 includes a symbols example. I expect these are going to be the most common use cases.

cbethell: I can try RefSeq again, but it was so memory intensive that it didn't complete locally on my machine. I can give it another try but I am not sure that's the way to go here.

cansavvy: Perfect!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,318 @@
...
+Fill out the pop up window with your email and our Terms and Conditions:  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-email.png" width=500>  
+
+It may take a few minutes for the dataset to process.
+You will get an email when it is ready. 
+
+## About the dataset we are using for this example
+
+For this example analysis, we will use this [RNA-seq zebrafish dataset](https://www.refine.bio/experiments/SRP056136/rna-seq-of-control-and-foxo3b-mo-zebrafish-cells).
```
cansavvy, line 79: I'm wondering if we want to use a bigger dataset. n = 4 is pretty small, although it doesn't really matter since we won't be using the data itself for anything. 

May want to add a bit more about the dataset either way.

cbethell: I had the same thoughts and ended up leaving the dataset as is. Good catch, more description is definitely needed here either way. 

Not sure if this means we should dig into finding a new dataset however.

cansavvy: I think switching to a dataset users can use for more than just this might be good. This is the time to make these changes I suppose.

cbethell: Gotcha, I'll look for one and post what I found on the issue relevant to this PR.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,319 @@
...
+We have a total of 542 out of 31882 Ensembl IDs with multiple mappings to Entrez IDs.
+With that rate, we can filter out the multi-mapped IDs if we want to keep 1:1 mappings for our downstream analyses by supplying the `"filter"` option to the `multiVals` argument og `mapIds()`, like we did in the [microarray gene identifier annotation example notebook](https://github.com/AlexsLemonade/refinebio-examples/tree/master/02-microarray/gene-id-annotation_microarray_01_ensembl.html).
+
+However, we may also want to keep the multi-mapped information for our downstream analyses so let's go ahead and write our mapped results and data to file!
+
+## Write mapped results to file
+
+```{r}
+# Write mapped and annotated data frame to output file
+readr::write_tsv(multi_mapped_df, file.path(
```
cansavvy, line 299: I don't really like the idea of keeping so many rows of redundant data, for fear that a user might think its appropriate to use this dataset (with duplicate rows of data) elsewhere. Keeping duplicate rows of values feels not ideal since we didn't actually measure each of these Entrez genes separately. However, I'm also not sure of any rationale to decide on Entrez ID over another. 

I have two ideas, neither of which I think are that great: 

1) We tell the user, we don't have a particular reason to choose one over the other, so let's just go with the first Entrez ID that comes up, then we re-run with `multiVals = "first"`, save to TSV.  

2) We say, we're not really sure which Entrez ID is relevant, but we'd like to store this information for future use. Then we can collapse all the Entrez IDs into one column where we store them all. We could do this by.   
```
df %>% 
  dplyr::mutate(entrez_ids = sapply(mapped_list, paste, collapse = "-")) %>% 
  dplyr::select(entrez_ids, dplyr::everything())
```

Option 1 would be easier for the user to use in the future, but if they can't find a Entrez ID they are looking for, they night have issues figuring that out but Option 2 allows them to keep all the information, should they need to access it. (However, I am concerned a concatenated column might not be easy for our users to use ether).

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,319 @@
...
+```{r}
+multi_mapped_df <- mapped_df %>%
+  # Let's group by the Ensembl ID's in the `Ensembl` column
+  dplyr::group_by(Ensembl) %>%
+  # Create a new variable containing the number of Entrez IDs mapped to each ID
+  dplyr::mutate(entrez_id_count = dplyr::n()) %>%
+  # Arrange by the genes with the highest number of Entrez IDs mapped
+  dplyr::arrange(desc(entrez_id_count)) %>%
+  # Join the remaining data from `df` using the Ensembl IDs
+  dplyr::inner_join(df, by = c("Ensembl" = "Gene"))
```
cansavvy, line 283: If we go with either of the options I mentioned, we could drop the inner_join here OR we could keep it so its abundantly clear that multi-mappings all end up having the same values.

cansavvy (commented):
Just a couple thoughts on the data to save to the TSV. I'm not exactly sure what makes the most sense but I have a few ideas.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,319 @@
...
+```{r}
+# Make the data in the order of the metadata
+df <- df %>%
+  dplyr::select(metadata$refinebio_accession_code)
+
+# Check if this is in the same order
+all.equal(colnames(df), metadata$refinebio_accession_code)
+
+# Bring back the "Gene" column in preparation for mapping
+df <- df %>%
```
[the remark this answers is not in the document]
cbethell: I left it as is because thinks looked a little more complex with the negative indexing (I agree, either way isn't perfect but at least this way we can stick with tidyverse).

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
cbethell: I agree that Option 1 would be easier for downstream use, while Option 2 allows them to keep all the information which many people would probably desire (even if just to save in a TSV or CSV format).

I incorporated both ideas in the last commit but end up saving Option 1 to the TSV file. I tried to keep it as short as possible but let me know if you still think we should just choose one! (Then we will probably revisit your [comment on keeping/dropping the `inner_join` of the expression values](https://github.com/AlexsLemonade/refinebio-examples/pull/218#discussion_r490950035)).

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
cbethell: I agree with both points here, however, I chose to drop it in the last commit because I do not think it is completely necessary here as they are able to see that multi-mappings have the same values in the `all_mapped_df` object (if we choose to drop that object, we may want to revisit this comment/thought).

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
cansavvy: We should probably choose one so users aren't confused on why we are making two data.frames and only saving one. I'm not sure which one though.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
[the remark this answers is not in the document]
cansavvy: I'm fine with that, and if the next reviewer disagrees, then we can revisit it.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jashapiro: I would probably do a hybrid: a column with the first (or best?) Entrez ID and one with all, separated by semicolons (Hyphens will be confusing, I think). I do think there should be some discussion of how you *might* pick one, with first as one option, but some other potential reasons discussed. Maybe you have a list of Entrez IDs you are interested in? Maybe you want the longest version of the gene? Maybe you want the oldest Entrez ID (which is probably first). I don't think we need to make a recommendation, but I do think there should be some acknowledgement of the issue.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,369 @@
...
+
+It looks like we have Entrez IDs that were successfully mapped to the Ensembl IDs we provided.
+However, the data is now in a `list` object, making it a little more difficult to explore.
+We are going to turn our list object into a data frame object in the next chunk.
+
+```{r}
+# Let's make our object a bit more manageable for exploration by turning it into a data frame using the `reshape2::melt()` function
+mapped_df <- mapped_list %>%
+  reshape2::melt(value.name = "Entrez") %>%
+  dplyr::rename("Ensembl" = "L1") # Here we are renaming the column holding the Ensembl IDs as good practice
```
jashapiro, line 259: We can do this in a more "modern" way without having to install `reshape2`
```suggestion
mapped_df <- mapped_list %>%
  tibble::enframe(name = "Ensembl", value = "Entrez") %>%
  tidyr::unnest() # enframe makes a `list` column, so here we are converting that to simpler format
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,369 @@
...
+Now let's check to see how many genes we have that were mapped to multiple IDs.
+
+```{r}
+multi_mapped_df <- mapped_df %>%
+  # Let's group by the Ensembl ID's in the `Ensembl` column
+  dplyr::group_by(Ensembl) %>%
+  # Create a new variable containing the number of Entrez IDs mapped to each ID
+  dplyr::mutate(entrez_id_count = dplyr::n()) %>%
+  # Arrange by the genes with the highest number of Entrez IDs mapped
+  dplyr::arrange(desc(entrez_id_count))
```
jashapiro, line 289: The resulting data table here is a bit confusing, as it still has both the Ensembl ID and the Entrez ID, so there are multiple rows for each Ensembl ID. The smallest change is below, using `summarize` instead of `mutate`:

```suggestion
multi_mapped_df <- mapped_df %>%
  # Let's group by the Ensembl ID's in the `Ensembl` column
  dplyr::group_by(Ensembl) %>%
  # Create a new variable containing the number of Entrez IDs mapped to each ID
  dplyr::summarize(entrez_id_count = dplyr::n()) %>%
  # Arrange by the genes with the highest number of Entrez IDs mapped
  dplyr::arrange(desc(entrez_id_count))
```

But if we want to be fancy, we can make it simpler with `count`
```
multi_mapped_df <- mapped_df %>%
  # Let's count the number of times each Ensembl ID appears in `Ensembl` column
  dplyr::count(Ensembl, name = "entrez_id_count") %>%
  dplyr::arrange(desc(entrez_id_count))
```

cbethell:
> I would probably do a hybrid: a column with the first (or best?) Entrez ID and one with all, separated by semicolons (Hyphens will be confusing, I think). I do think there should be some discussion of how you might pick one, with first as one option, but some other potential reasons discussed. Maybe you have a list of Entrez IDs you are interested in? Maybe you want the longest version of the gene? Maybe you want the oldest Entrez ID (which is probably first). I don't think we need to make a recommendation, but I do think there should be some acknowledgement of the issue.

I believe I successfully executed what you suggested here @jashapiro. Would you mind taking a look to see that I did? Feel free to make any additional suggestions throughout the notebook!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+
+### Store all mapped values
+
+We are not sure which Entrez ID is relevant as yet, but we could store all of the mapped information for future use.
+In the next chunk, we show how we can collapse all the Entrez IDs into one column where we store them all for each unique Ensembl ID.
+
+```{r}
+final_mapped_df <- df %>%
+  # Collapse the mapped values in `mapped_list` into one column named `entrez_ids`
+  dplyr::mutate(entrez_ids = sapply(mapped_list, paste, collapse = ";")) %>%
```
cansavvy, line 306:
```suggestion
  dplyr::mutate(all_entrez_ids = sapply(mapped_list, paste, collapse = ";")) %>%
```
Since below you named the next column, `first_entrez_ids`, you can name this one in parallel. (If you do this you'll need to change the next couple lines too.

cansavvy: Including the comments!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+
+### Map Ensembl IDs to gene symbols -- keeping only the first mapped value
+
+Let's re-run `mapIds()`, this time using `multiVals = "first"` and add this information to our `final_mapped_df` object.
+
+```{r}
+final_mapped_df <- data.frame(
+  "first_mapped_entrez_id" = mapIds(
+    org.Dr.eg.db, # Replace with annotation package for the organism relevant to your data
+    keys = df$Gene,
```
cansavvy, line 335:
```suggestion
    keys = final_mapped_df$Gene,
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+    column = "ENTREZID", # Replace with the type of gene identifiers you would like to map to
+    keytype = "ENSEMBL", # Replace with the type of gene identifiers in your data
+    multiVals = "first" # This will keep only the first mapped value for each Ensembl ID
+  )
+) %>%
+  # Make an `Ensembl` column to store the rownames
+  tibble::rownames_to_column("Ensembl") %>%
+  # Join the remaining data from `df` using the Ensembl IDs
+  dplyr::inner_join(final_mapped_df, by = c("Ensembl" = "Gene"))
```
cansavvy, line 344: I haven't run this locally, but I don't think you need to `inner_join()` this? I think you will already have the data from before?

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+
+```{r}
+final_mapped_df <- data.frame(
+  "first_mapped_entrez_id" = mapIds(
+    org.Dr.eg.db, # Replace with annotation package for the organism relevant to your data
+    keys = df$Gene,
+    column = "ENTREZID", # Replace with the type of gene identifiers you would like to map to
+    keytype = "ENSEMBL", # Replace with the type of gene identifiers in your data
+    multiVals = "first" # This will keep only the first mapped value for each Ensembl ID
+  )
```
cansavvy, line 339:
```suggestion
  ), final_mapped_df
```

Then no `inner_join` is needed. (I think this should be fine). Unless we find it too confusing for users to follow? (II personally find it easier to follow than a join.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+  tibble::rownames_to_column("Ensembl") %>%
+  # Join the remaining data from `df` using the Ensembl IDs
+  dplyr::inner_join(final_mapped_df, by = c("Ensembl" = "Gene"))
+```
+
+Our `final_mapped_df` object now has a column named `first_mapped_entrez_id` that contains only the first mapped Entrez ID, in addition to the `entrez_ids` column that contains all mapped Entrez IDs per Ensembl ID.
+
+```{r}
+final_mapped_df %>%
+  dplyr::filter(stringr::str_detect(entrez_ids, ";")) %>%
```
cansavvy, line 351: Is this filter just so you can see the preview of multiples? We may wanna put a comment to explain that.

cansavvy (commented):
This looks good! I like the combo idea! I have a couple ideas and wording suggestions.

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,380 @@
...
+  # Let's count the number of times each Ensembl ID appears in `Ensembl` column
+  dplyr::count(Ensembl, name = "entrez_id_count") %>%
+  # Arrange by the genes with the highest number of Entrez IDs mapped
+  dplyr::arrange(desc(entrez_id_count))
+
+# Let's look at the first 6 rows of our `multi_mapped` object
+head(multi_mapped_df)
+```
+
+Looks like we have some cases where 15 Entrez IDs mapped to a single Ensembl ID.
```
jashapiro, line 293: Keeping the exclamation point, but changing direction to match the way we are going...
```suggestion
Looks like we have one case where one Ensembl id mapped to 15 Entrez IDs!
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jashapiro, line 344: Just for consistency, since the other columns will have `id` in the name.
```suggestion
  tibble::rownames_to_column("Ensembl_id") %>%
  # Join the remaining data from `df` using the Ensembl IDs
  dplyr::inner_join(final_mapped_df, by = c("Ensembl_id" = "Gene"))
```

jashapiro: It is always safer to use a join in case order has changed, so I would leave it 😄

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jashapiro: I think this would result in a duplicated column if you leave the `rownames_to_column` step... but I like joins. They are a good thing to learn and a good habit because I have been burned too many times by functions that don't preserve order (`mapIds` seems to, but I still don't trust it).

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jashapiro, line 306: Related to my fear of order changes, I wouldn't trust this to return a list in the same order as `df`, so I would probably do this separately and join them.

Since you already have this in a data frame, from the step before, you could do this with `dplyr::summarize` pretty neatly: 
```
collapsed_mapped_df <- mapped_df %>%
  dplyr::group_by(Ensembl) %>%
  dplyr::summarize(entrez_ids = paste(Entrez, collapse = ";"))
```

jashapiro (commented):
I just made some comments on the multimapping sections... I like what you did, so I just made some alternative tidy suggestions which you can use or not. 

I also support your use of joins!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
jashapiro: I maybe don't like this after seeing how the data comes out of mapIds. Ignore this and stick with just plain `Ensembl`

cansavvy: The order doesn't change based on my suggestion because you'd directly provide final_mapped_df$Ensembl (or whatever the correct column name is) to be mapped.

jashapiro: `final_mapped_df$Ensembl` is not guaranteed  to be in the same order as `df$Gene`. In fact, in this case `final_mapped_df` was sorted by Entrez_id on line 310, so I think it *won't* be in the same order. The fact that this kind of thing is hard to keep track of is exactly why I favor joins.

jashapiro: Oh, I see what you are saying... I missed that change. But I stand by my general point, because I still don't trust `mapIds` not to rearrange things.

Dockerfile
```diff
@@ -59,8 +59,7 @@ RUN install2.r --error --deps TRUE \
     pheatmap \
     Rtsne \
     umap  \
-    VennDiagram \
-    org.Mm.eg.db
```
cansavvy, line 63: We will want to install this still, I think this is an indication that the branch is out of date? I think this PR predated Jackie's `up-to-date-with-master` changes: https://github.com/AlexsLemonade/refinebio-examples/pull/221

So we should be cognizant of that before we merge with master we need to resolve this.

cbethell: Ah this PR is actually up to date with master @cansavvy but as noted in this PR's original comment, I had the question of: `Should we be adding the individual org.Dr.eg.db organism package to the Dockerfile?`

and also began getting the following error when building the Docker image with `org.Mm.eg.db` being installed:
```
The downloaded source packages are in
        ‘/tmp/downloaded_packages’
Error: package ‘org.Mm.eg.db’ is not available (for R version 4.0.2)
In addition: Warning message:
package ‘org.Mm.eg.db’ is not available (for R version 4.0.2) 
```

cansavvy: I think it needs to be installed with Bioconductor.

cansavvy: I think given the purposes of just making sure those of us developing the material are using all the same packages, it makes sense to have `org.Dr.eg.db` added to the Dockerfile. 

Did you have some thoughts about why we shouldn't add it?

cbethell: Ah okay I got it to build successfully. I have no arguments on why it shouldn't be added here so I added both the mouse and zebrafish annotation packages 👍 (will push Docker image after PR is merged as instructed in `CONTRIBUTING.md`).

cansavvy (approved):
LGTM! Just one teenie tiny comment!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+A full list of the annotation R packages from Bioconductor is at this [link](https://bioconductor.org/packages/release/BiocViews.html#___AnnotationData) [@annotation-packages].
+
+***
+
+<!-- Do not delete this line --> <a name="analysis" style="padding-top:56px;margin-top:-56px;">&nbsp;</a>
+
+# Obtaining Annotation for Ensembl IDs - RNA-seq
+
+Ensembl IDs can be used to obtain various different annotations at the gene/transcript level.
+Let's get ready to use the Ensembl IDs from our mouse dataset to obtain the associated Entrez IDs.
```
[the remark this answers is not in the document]
cbethell: Good catch!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+
+It looks like we have Entrez IDs that were successfully mapped to the Ensembl IDs we provided.
+However, the data is now in a `list` object, making it a little more difficult to explore.
+We are going to turn our list object into a data frame object in the next chunk.
+
+```{r}
+# Let's make our object a bit more manageable for exploration by turning it into a data frame
+mapped_df <- mapped_list %>%
+  tibble::enframe(name = "Ensembl", value = "Entrez") %>%
+  tidyr::unnest(cols = Entrez) # enframe makes a `list` column, so here we are converting that to simpler format
```
jashapiro, line 259: I thought this comment could use a bit more detail.
```suggestion
  # enframe makes a `list` column, so we will convert that to simpler format with `unnest()
  # This will result in one row of our data frame per list item
  tidyr::unnest(cols = Entrez) 
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  # Arrange by the genes with the highest number of Entrez IDs mapped
+  dplyr::arrange(desc(entrez_id_count))
+
+# Let's look at the first 6 rows of our `multi_mapped` object
+head(multi_mapped_df)
+```
+
+Looks like we have one case where one Ensembl ID mapped to 15 Entrez IDs!
+We have a total of 235 out of 31882 Ensembl IDs with multiple mappings to Entrez IDs.
+With that rate, we can filter out the multi-mapped IDs if we want to keep 1:1 mappings for our downstream analyses by supplying the `"filter"` option to the `multiVals` argument of `mapIds()`.
```
jashapiro, line 294: Is this something we would really want to do? I feel like suggesting filtering out the multimaps is dangerous...I see that we do it elsewhere, but I am not sure I am keen on it. 

More relevant here, I think this choice probably not particularly dependent on the rate so much as what genes would be lost, so I might suggest this with more ambiguity as to why you might do it: 
```suggestion
One option is to filter out the multi-mapped IDs and keep only the 1:1 mappings for our downstream analyses by supplying the `"filter"` option to the `multiVals` argument of `mapIds()`.
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+```{r}
+collapsed_mapped_df %>%
+  # Filter `collapsed_mapped_df` to include only the rows where `all_entrez_ids` values include the ";" character -- we know that these are the rows with multiple mapped values
+  dplyr::filter(stringr::str_detect(all_entrez_ids, ";")) %>%
+  # We only need a preview here
+  head()
+```
+
+You may have a list of Entrez IDs you are interested in, in which case, the above format may work best for you.
+In a different study, you may want the oldest Entrez ID (which is probably first), in which case, you can create a column that stores just the first mapped Entrez ID that comes up for each Ensembl ID.
```
jashapiro, line 321: Do we have confirmation that the first Entrez ID is the lowest number? i.e. is the list we get here sorted?

jashapiro (approved):
Looks good!

I have a few very minor comments, but I don't think I will need to rereview this!

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+Click the "Download Now" button on the right side of this screen.  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-now.png" width=200>  
+
+Fill out the pop up window with your email and our Terms and Conditions:  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-email.png" width=500>  
+
+If you want to use this dataset with other examples you may want to download it non-quantile normalized.
+See the [RNA-seq header section](https://github.com/AlexsLemonade/refinebio-examples/tree/master/03-rnaseq/00-intro-to-rnaseq.html#14About_quantile_normalization) about quantile normalization.
```
jashapiro, line 79: Fix link?
```suggestion
See the [RNA-seq header section](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/00-intro-to-rnaseq.html#about-quantile-normalization) about quantile normalization.
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+We suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
+From here you can customize this analysis example to fit your own scientific questions and preferences. 
+
+refine.bio data comes with gene level data with Ensembl IDs. 
+Although this example script uses Ensembl IDs from Zebrafish, (<i>Danio rerio</i>), to obtain Entrez IDs this script can be easily converted for use with different species or annotation types <i>eg</i> protein IDs, gene ontology, accession numbers.
+
+<b>For different species</b>, wherever the abbreviation `org.Dr.eg.db` or `Dr` is written, it must be replaced with the respective species abbreviation <i>eg</i> for<i> Homo sapiens</i> `org.Hs.eg.db` or `Hs` would be used.
+In the case of our [microarray gene identifier annotation example notebook](https://github.com/AlexsLemonade/refinebio-examples/tree/master/02-microarray/gene-id-annotation_microarray_01_ensembl.html), a Mouse (<i>Mus musculus</i>) dataset is used, meaning `org.Mm.eg.db` or `Mm` would also need to be used there.
```
jashapiro, line 150: Another link fix.
```suggestion
In the case of our [microarray gene identifier annotation example notebook](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html), a Mouse (<i>Mus musculus</i>) dataset is used, meaning `org.Mm.eg.db` or `Mm` would also need to be used there.
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  dplyr::arrange(desc(entrez_id_count))
+
+# Let's look at the first 6 rows of our `multi_mapped` object
+head(multi_mapped_df)
+```
+
+Looks like we have one case where one Ensembl ID mapped to 15 Entrez IDs!
+We have a total of 235 out of 31882 Ensembl IDs with multiple mappings to Entrez IDs.
+With that rate, we can filter out the multi-mapped IDs if we want to keep 1:1 mappings for our downstream analyses by supplying the `"filter"` option to the `multiVals` argument of `mapIds()`.
+If you want to see an example of a `multiVals = "filter"` strategy see the [microarray gene identifier annotation example notebook](https://github.com/AlexsLemonade/refinebio-examples/tree/master/02-microarray/gene-id-annotation_microarray_01_ensembl.html).
```
jashapiro, line 295: One more link fix...
```suggestion
If you want to see an example of a `multiVals = "filter"` strategy see the [microarray gene identifier annotation example notebook](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html).
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
```diff
@@ -0,0 +1,382 @@
...
+readr::write_tsv(final_mapped_df, file.path(
+  results_dir,
+  "SRP040561_Entrez_IDs.tsv" # Replace with a relevant output file name
+))
+```
+
+# Resources for further learning
+
+- Marc Carlson has prepared a nice [Introduction to Bioconductor Annotation Packages](https://bioconductor.org/packages/release/bioc/vignettes/AnnotationDbi/inst/doc/IntroToAnnotationPackages.pdf) [@Carlson2020-vignette]
+- See our [microarray gene ID conversion notebook](https://github.com/AlexsLemonade/refinebio-examples/tree/master/02-microarray/gene-id-annotation_microarray_01_ensembl.html) as another applicable example, since the steps for this workflow do not change with technology [@gene-id-annotation-microarray].
```
jashapiro, line 372:
```suggestion
- See our [microarray gene ID conversion notebook](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html) as another applicable example, since the steps for this workflow do not change with technology [@gene-id-annotation-microarray].
```

03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd
cbethell: Good question! It does not explicitly that the returned list is sorted in the `AnnotatationDbi` or `mapIDs()` docs but the lists of mapped values returned does appear to be sorted from my investigation.

PR #221, opened 2020-09-18
Create "Hello, world!" GH action to enforce that PR branches are up to date with master

jaclyn-taroni:
## Purpose

On the OpenPBTA repositories, we require branches are up to date before merging. The way to set that up is to require status checks (CircleCI, AppVeyor, what have you) pass before merging and then you can require branches to be up to date before merging, quoting from the settings here:

> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below). 

I currently checked the Page Build status, but it never finishes checking so we can never merge a PR without admin powers.

## Strategy

Create GH action that is a one-liner (`echo Hello, world!`) from their simple workflow template. This should almost never fail and should run quickly, which means we do not expect that it will hold up development and we expect it will allow us to require branches be up to date before merging.

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -1,0 +1,28 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


cansavvy (approved):
Worth a shot!

PR #222, opened 2020-09-18
Add a section about the PR status check to contributing guidelines

jaclyn-taroni:
## Purpose

In #221, we added a `echo Hello, world!` to GitHub Actions so we have a status check that allows us to require that pull requests are up to date with `master` before merging but that runs very quickly and should never fail. In 6 months, we might forget that that isn't _really doing anything_ so I wrote something about it in the contributing guidelines.


cansavvy (approved):
Looks great! I think this will be super helpful!

Issue #236, opened 2020-09-26
Automation: Build Docker image and push to Dockerhub upon merge


jaclyn-taroni:
Related to #228 

[Our contributing guidelines](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#pushing-docker-image-updates) state:

> All necessary packages needed for running all analyses should be added to the Dockerfile and it should be re-built to make sure it builds successfully.
>
> ...
>
> After a PR with Dockerfile changes is merged, its associated image should be pushed to the Docker hub repository.

It would be great to automate:

1. Ensuring the Docker image can build upon filing a pull request.
2. Build the Docker image and push to Dockerhub (`ccdl/refinebio-examples:latest`) upon merge to `master`. 

Here are a few examples/links I found that could come in handy:

- [Starter workflow YAML](https://github.com/actions/starter-workflows/blob/dacfd0a22a5a696b74a41f0b49c98ff41ef88427/ci/docker-image.yml)
- [Example for pushing to Dockerhub](https://github.com/containrrr/watchtower/pull/438/files)
- [Build and push Docker images from Marketplace](https://github.com/marketplace/actions/build-and-push-docker-images)

First we should establish how long we expect this to take by building locally with `--no-cache`. We may need help setting up the secrets from the engineering team.

jashapiro:
I think we can tackle this in two steps as outlined. We can do #1 without any secrets, so I think that is worth setting up on its own first.

I can take that step to start.

jashapiro:
It seems like there is now the ability to leverage caches, but I have not explored this yet: https://github.com/docker/build-push-action#leverage-github-cache

Issue #235, opened 2020-09-25
Remove install2.r from Dockerfile after 4.0.2 rocker freeze


cansavvy:
From @jashapiro 's comment on #225 

https://github.com/AlexsLemonade/refinebio-examples/pull/225#discussion_r495054016

> At that point the CRAN mirror should also freeze. One other note is that we may want to remove install2.r from installation lines, because it seems to somehow not delete cached downloads. Not a big deal, but eliminating it did cut some bulk from openpbta (AlexsLemonade/OpenPBTA-analysis#776).

Sounds like this should help make the Docker image a little smaller.

PR #237, opened 2020-09-26
Create docker build workflow

jashapiro:
## Purpose:
This draft PR is to test automatic docker builds through github actions.
Partially addresses #236 

## Strategy
Modify the blank workflow to build the docker image instead.

To prevent the full repository from going into the build context, I moved the Docker file to a subdirectory.

Does it work?


jashapiro:
It works! Takes 14 minutes with no caching. About to test using github caching actions, using example from https://github.com/docker/build-push-action#leverage-github-cache

jashapiro:
Using the github cache adds some overhead, but it does speed things up in general: rebuilding (with reinstallation of python packages since I changed the ordering there) took ~5 minutes, including pulling from the cache and posting to it.

Note that the cache has a size limit of 5GB (compressed, I think), if we think we are likely to hit this with docker images, we might think about another solution. I don't know how to check the current cache usage...

jaclyn-taroni:
Seems pretty easy for us to roll back the changes to building without caching (https://github.com/AlexsLemonade/refinebio-examples/pull/237/commits/b40b0ce44cb979b2f2b80c6efd6a55ec5d80b4f2) should we run into an issue, so I say let's try it! I assume addressing #235 will help us keep the cache size down, too..

jashapiro:
Having now update `CONTRIBUTING.md` with the new Dockerfile path, I am now marking this ready for review. 

Running the Build action with no changes to the Dockerfile takes ~4.5 minutes. Most of that seems to be uncompressing and recompressing the cache files.

Oh, and the cache size is reported in the action logs as ~1536 MB. (Not sure if that is just this item, or total, but it doesn't matter in this case/)

jaclyn-taroni (approved):
One comment that's not easily made in-line on this PR! Should we get rid of (lines 8-9):

```yaml
  push:
    branches: [ master ]
```

For the moment. That is a remnant of the hello world template I used so we could get enforcing up to date branches up and running and I don't see a reason we would need to build things after we merge.

Otherwise, LGTM 👍

jashapiro:
> One comment that's not easily made in-line on this PR! Should we get rid of (lines 8-9):
> 
> 
> 
> ```yaml
> 
>   push:
> 
>     branches: [ master ]
> 
> ```
> 
> 
> 
> For the moment. That is a remnant of the hello world template I used so we could get enforcing up to date branches up and running and I don't see a reason we would need to build things after we merge.
> 
> 
> 
> Otherwise, LGTM 👍 

I left it just in case... I was thinking we would make a separate action for pushes to master that does a full build (no cache?) and pushes to dockerhub.

jaclyn-taroni:
> I was thinking we would make a separate action for pushes to master that does a full build (no cache?) and pushes to dockerhub.

This sounds good - separate action was what I was thinking, too, and it makes sense that we are not particularly concerned with how long that one will take.

jaclyn-taroni:
Just remembered I added a section to the contributing guidelines that covers status checks: https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#pull-request-status-checks

To quote myself from #222

> In 6 months, we might forget that that isn't _really doing anything_ so I wrote something about it in the contributing guidelines.

We should probably update this section to reflect what this is now doing with these changes.

jaclyn-taroni (approved):
🚢

Issue #228, opened 2020-09-23
The PR Checklist is long; what can we automate reliably?


cansavvy:
There's a lot of items in the checklist for our human brains to do and remember. This takes a toll on authors of new examples and reviewers can miss some of these items over the iterative process of the PR reviewing. 

Now that we have some basic GitHub actions set up, we should see what checklist items the computers can take over for us. 

This will probably involve some edits to: https://github.com/AlexsLemonade/refinebio-examples/blob/master/.github/workflows/blank.yml

Beyond that I am unsure what this will involve at this point. Here is the PR checklist in its current state as well as some comments from me. 

Items that I don't think can really be automated because they are more content based: 
> *  All `{{BLANKS}}` have been replaced with the correct content.
> * [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
> * Seed is set (if applicable)
> * Comments and documentation are up to date. 
> * Removed any manual numbering of sections.
> * Removed any instances of chunk naming.
> * Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar).

This could be more automated, but we'd need to figure out feedback:
> * Spell checked any Rmd file or md file. 
> * Check links

I don't think we want this automated because security on AWS would be sticky?
> * Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

Probably the most prime suspects for automation: 
> * Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds.
> * In the Docker container, [snakemake was run for rendering](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#how-to-re-render-the-notebooks).

jashapiro:
As we get into more an more modules, every addition requires a small change to every previous file (to keep the html up to date). This is a bit annoying for review, and it would be nice if we could separate the changed `html` files from the `Rmd` files. I think this should be possible with some automation; commits to master could be set to rerender the full site, and we could change the contribution guidelines to suggest not committing html files (or even add them to the ignore list)?

This comment is partly born of clicking past a bunch of html files with changed headers...

jashapiro:
[...]

PR #238, opened 2020-09-27
Add action for docker build and Push

### LINKED ISSUE #236: Automation: Build Docker image and push to Dockerhub upon merge

jashapiro:
## Purpose:

If this works, this PR adds an action to automatically build and push the Docker image when a PR is merged to master. 
~~Due to questions about secrets (see below), I have filed this as a draft.~~ Now live!

Closes #236

## Strategy

I added a new workflow that will build the dockerfile and push to dockerhub.

This will not work until we add docker credentials in the github secrets, storing them in ~~`DOCKERHUB_USERNAME`~~ `DOCKER_ID` and ~~`DOCKERHUB_TOKEN`~~ `DOCKER_PASSWORD`. We should talk to the engineering team about whether they have credentials [stored for the organization](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization) that we can use, or whether we should have a separate set for this repository. 

I made this build without using the cache, so that dockerhub builds will be clean and fresh. 
However, this could potentially result in some mismatch between the version that is tested in `docker-build.yml` and the version that ultimately pushed to Dockerhub. I don't expect this to be a problem, as the one in `docker-build.yml` is just a check and is not stored anywhere that anyone else would access it, but I thought it should be mentioned. 

It should be possible to extend this workflow to build the site with a `snakemake` step. The part I don't know is how to get the html files out after the `snakemake` build and committed to the repository. I think it has something to do with artifacts. But this is a problem for later.

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -5,8 +5,8 @@ name: Build and Push Docker
 on:
   push:
     branches: [ master ]
+    path: docker/Dockerfile
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build-push"
   build-push:
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -5,7 +5,7 @@ name: Build and Push Docker
 on:
   push:
     branches: [ master ]
-    path: docker/Dockerfile
+    paths: [ docker/Dockerfile ]
 
 jobs:
   # This workflow contains a single job called "build-push"
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -20,8 +20,8 @@ jobs:
       - name: Login to DockerHub
         uses: docker/login-action@v1 
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          username: ${{ secrets.DOCKER_ID }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
       # Build docker image (We are not using caching here to force a clean build)
       - name: Build and Push Docker image
         uses: docker/build-push-action@v2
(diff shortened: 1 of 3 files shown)


jaclyn-taroni:
I, too, think it has something to do with artifacts. Here's an `r-lib` example that makes use of `container:` to render and upload the HTML: https://github.com/r-lib/actions/tree/master/examples#docker-based-workflow

Our use case may be a bit different if we need to _build_ our image and then run inside of the container from the build as part of the same job (probably will require `COPY` in the Dockerfile 🤔 ). It depends, of course. I am planning to file an issue today about rendering specifically based on the convo in #228.

jashapiro:
I added checking for changes to the Dockerfile to the trigger for this action, on the logic that if there are no changes to the Dockerfile we don't need to rebuild and push. However, I think this would interfere if we intend to use the same action for building the site with the latest changes to Rmd files. I think we will want to do that, as there will certainly be times when there are changes to the Dockerfile in addition to Rmd changes, and there doesn't seem to be a clean way to make sure this action has run before another action (I might not have found it).

In any case, I left it here partly just as a temporary test of the functionality.

kurtwheeler (approved):
LGTM!

PR #246, opened 2020-09-28
Update Microarray Ortholog Mapping example

### LINKED ISSUE #202: Microarray Analysis Updates: Ortholog mapping
Besides the main update items listed on #173  : 

> - Have the getting started template material added to it (# 146)
> - Update to adhere to the style guide (#148)
> - Add the dataset added to the S3 bucket. (See https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-s3-bucket)
> - Add to the Snakefile and render with snakemake on the Docker container. 

- We'll need to get rid of the bit about `X1` column since that has since been fixed in refine.bio downloads. 
- Are the installation complications for `hcop` that are described in this notebook still an issue or can we go with a more straight forward installation approach? 
- I think it seems fine to keep this as zebrafish -> human but on that note, we will keep in mind what the parallel RNA-seq issue #130 should model that is the same general steps but incorporates different variation.

I think we should add some exercise steps that shows the user how to investigate their mapping across species. How many things mapped vs didn't map? And ultimately we should show them the idea that you shouldn't expect most/everything things to map across species and that is okay.

cbethell:
## Purpose:

This PR closes #202.

## Strategy

I followed the directions in the [original comment on the issue](https://github.com/AlexsLemonade/refinebio-examples/issues/202#issue-690184600) to update the notebook using `template_example.Rmd` and `CONTRIBUTING.md`.

I also added steps to investigate the mapping, i.e. "How many things mapped vs didn't map"?

I chose to save the uncollapsed and collapsed versions of the data frame (collapsing gene symbols for each unique Ensembl ID).

Also note that I replace the samples n = 4 zebrafish dataset that was being used in this notebook with the samples n = 10 zebrafish dataset that was also used in the [differential expression microarray example](https://github.com/AlexsLemonade/refinebio-examples/blob/master/02-microarray/differential-expression_microarray_01_2-groups.Rmd).

- As noted above, I chose to save the uncollapsed and collapsed versions of the data frame (collapsing gene symbols for each unique Ensembl ID). 

  However, as noted in the notebook, multiple zebrafish Ensembl gene IDs map to the same human symbol 
 which results in duplicated human gene symbols and some downstream tools may need this to be resolved.

  I added a note that 
  "One way to resolve this may be to map to a different gene identifier.
  See our [gene ID annotation microarray notebook](https://alexslemonade.github.io/refinebio-examples/02-microarrary/gene-id-annotation_microarray_01_ensembl.html) for an example of mapping between gene 
  identifiers"
  and also showed users how to collapse the Ensembl IDs for each unique gene symbol.
 
  I am not sure that this was the best way to account for the multi-mappings in this case or if these are good 
  ways to potentially resolve for downstream tools.

### Content checks

### Formatting Checks
* [x] Spell checked any Rmd file or md file.

### Add datasets to S3


02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-email.png" width=500>  
+
+It may take a few minutes for the dataset to process.
+You will get an email when it is ready. 
+
+## About the dataset we are using for this example
+
+For this example analysis, we will use this [CREB overexpression zebrafish dataset](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process), which contains 10 samples.
+
+@Tregnago2016 measured microarray gene expression of zebrafish samples overexpressing human CREB, as well as control samples.
```
cansavvy, line 86:
```suggestion
@Tregnago2016 measured microarray gene expression of ten zebrafish samples, five overexpressing human CREB, as well as five control samples.
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+Fill out the pop up window with your email and our Terms and Conditions:  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-email.png" width=500>  
+
+It may take a few minutes for the dataset to process.
+You will get an email when it is ready. 
+
+## About the dataset we are using for this example
+
+For this example analysis, we will use this [CREB overexpression zebrafish dataset](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process), which contains 10 samples.
```
cansavvy, line 84:
```suggestion
For this example analysis, we will use this [CREB overexpression zebrafish dataset](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process).
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+
```
cansavvy, line 85:
```suggestion
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+```
+
+We're interested in the `human_symbol` and `zebrafish_ensembl` columns specifically.
+
+```{r}
+human_symbol_df <- zebrafish %>%
+  # we'll want to subset zebrafish to only the columns we're interested in
+  dplyr::select(zebrafish_ensembl, human_symbol) %>%
+  # only retain the genes from the expression matrix that are in zebrafish data
+  dplyr::inner_join(df, by = c("zebrafish_ensembl" = "Gene"))
```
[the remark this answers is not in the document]
cansavvy: But, I would reverse the order of the data frames and do a left join instead. I think `left_join()`s are more intuitive to follow (could be just me).

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+```
+
+Looks like we have mapped symbols!
+It does not explicitly tell us if we have IDs that did not map to symbols, which would result in NAs so let's explore to find out if we do have any.
+
+```{r}
+any(is.na(human_symbol_df$human_symbol)) # Logical statement to check if there are any NAs in the `human_symbol` column of our mapped data frame
+```
+
+There are no NAs!
```
cansavvy, line 265: I find it unlikely (maybe impossible?) that a no NAs situation would happen when mapping across species. So that's what was suspicious to me and made me look again at your `inner_join()` step.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+    "GSE71270_ensembl_to_human_gene_symbol.tsv" # Replace with a relevant output file name
+  )
+)
+```
+
+*Some things to note:*
+
+* Multiple zebrafish Ensembl gene IDs map to the same human symbol which results in duplicated human gene symbols.
+Some downstream tools may need this to be resolved.
+One way to resolve this may be to map to a different gene identifier.
```
cansavvy, line 289: I know you pointed this out in your PR summary, but can you say more about what you mean by this?

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+readr::write_tsv(
+  human_symbol_df,
+  file.path(
+    results_dir,
+    "GSE71270_ensembl_to_human_gene_symbol.tsv" # Replace with a relevant output file name
+  )
+)
+```
+
+*Some things to note:*
```
cansavvy, line 285: I think this should become its own section of sorts. Something like `Sorting out multi mappings`

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+any(is.na(human_symbol_df$human_symbol)) # Logical statement to check if there are any NAs in the `human_symbol` column of our mapped data frame
+```
+
+There are no NAs!
+This means that all of our zebrafish Ensembl IDs mapped to at least one human gene symbol!
+This is exciting because it is not always the case everything maps across species and that is okay.
+
+However, if you have almost all NAs it is possible that the function was executed incorrectly or you may want to consider using a different gene identifier, if possible.
+
+## Write results to file
```
cansavvy, line 271: Now that I see what's coming up next, I wouldn't go to writing this file yet before we explore what's here in the next sections. So I'd move this section down toward the end of this notebook.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+
+* Multiple zebrafish Ensembl gene IDs map to the same human symbol which results in duplicated human gene symbols.
+Some downstream tools may need this to be resolved.
+One way to resolve this may be to map to a different gene identifier.
+See our [gene ID annotation microarray notebook](https://alexslemonade.github.io/refinebio-examples/02-microarrary/gene-id-annotation_microarray_01_ensembl.html) for an example of mapping between gene identifiers.
+* If a zebrafish Ensembl gene ID maps to multiple human symbols, the gene expression values get duplicated.
+Let's look at the `ENSDARG00000069142` example below.
+
+```{r}
+human_symbol_df %>%
```
cansavvy, line 295: I like this using an example to illustrate the idea you are explaining can you do the same for the first bullet point as well? 

I think it makes sense to split this into two sections, something like: `collapsing duplicate human IDs` and `collapsing/summarizing multiple zebrafish IDs`. 

But of course the `collapsing/summarizing multiple zebrafish IDs` strategy we will have to think about what makes the most sense. This will be dependent on what a user would be using this for downstream, which we can't know for certainty, but we can try to model what we think the most common use cases would be.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+  # Collapse the mapped values in `human_symbol_df` into one column named `all_human_symbols`
+  dplyr::summarize(all_human_symbols = paste(human_symbol, collapse = ";"))
+
+head(collapsed_symbol_df)
+```
+
+Since multiple zebrafish Ensembl gene IDs map to the same human symbol, we can also collapse all Ensembl IDs that each unique gene symbol mapped to.
+
+```{r}
+collapsed_ensembl_df <- human_symbol_df %>%
```
cansavvy, line 318: Currently, this doesn't end up having the data associated with it, and I'm hesitant to combine values across zebrafish transcripts using an average, because that is a lot of data to lose and will likely not be representative of the zebrafish biology (which is an inherent issue with ortholog mapping).

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+    results_dir,
+    "GSE71270_collapsed_ensembl_to_human_gene_symbol.tsv" # Replace with a relevant output file name
+  )
+)
+```
+
+# Resources for further learning
+
+- [HCOP: the HGNC comparison of orthology predictions search tool](https://pubmed.ncbi.nlm.nih.gov/16284797/) [@Wright2005]
+
```
cansavvy, line 345: When the RNA-seq example is done we will want to link it here.

cansavvy (commented):
I have some reorganization suggestions and we should discuss what strategy of collapsing multi mappings (or not collapsing) makes the most sense in the most contexts. 

Because I'm worried about losing data or manipulating it in an irresponsible way, I think it makes sense to not collapse zebrafish values to human IDs, because this would have to involve some sort of averaging which I don't think we represent the original measurements, particularly since we don't know what the downstream use cases would be for this. 

For collapsing human values though, I think it is fine to leave NAs in (see my comment about your inner_join) and take a look at the mapping rates. Then I think its fine for us to take a similar strategy as we did in the [rna-seq annotation example where we had multi mappings](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.html#44_Explore_gene_ID_conversion): show how to just keep one mapping as well as how to collapse and keep all mappings (which you are already doing here). 

We may want to have a call to discuss these strategies.

cansavvy:
Another thought, how much better (or maybe more likely worse) is mapping zebrafish ensembl to human ensembl? I see you already have the data there.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,355 @@
...
+```{r}
+# We can use the `summary()` function to get a better idea of the distribution of symbols in the `Symbol` column
+summary(as.factor(human_symbol_df$human_symbol)) # We need to use the `as.factor()` function here in order to get the count of each unique value returned
+```
+
+Looks like we have mapped symbols!
+It does not explicitly tell us if we have IDs that did not map to symbols, which would result in NAs so let's explore to find out if we do have any.
+
+```{r}
+any(is.na(human_symbol_df$human_symbol)) # Logical statement to check if there are any NAs in the `human_symbol` column of our mapped data frame
```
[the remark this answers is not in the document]
cbethell: Ah, of course!! I did find the no NAs situation strange as well but completely overlooked this issue in the order of things.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cbethell: In other words, another gene identifier may have a better 1:1 ratio to the zebrafish Ensembl IDs, i.e. human Ensembl IDs instead of symbols (like I see you mention below).

cansavvy: I would use the wording here instead of what you have in the notebook. And probably move away from using the word `resolve` in this context since it's not so much a problem to be solved as it is just the nature of how IDs map across species and between ID types (as in ortholog mapping is shaky at best).

cbethell:
@cansavvy, I believe I addressed things as you mention above and as we discussed on our call yesterday. Would you take a look to ensure that:

- I have successfully replaced `inner_join` with `left_join` as you suggested and where needed
- I re-arranged the notebook appropriately
- I got the collapsing of the zebrafish IDs section to the furthest point we discussed, without manipulating and losing _all of the data_

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+## Mapping zebrafish Ensembl gene IDs to human symbols
+
+refine.bio data uses Ensembl gene identifiers, which will be in the first column.
+
+```{r}
+# Let's take a look at the first 6 items in the `Gene` column of `df`
+head(df$Gene)
+```
+
+Enembl gene identifiers have different species-specific prefixes.
```
cansavvy, line 226:
```suggestion
Ensembl gene identifiers have different species-specific prefixes. 
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+```{r}
+zebrafish
+```
+
+We're interested in the `human_symbol` and `zebrafish_ensembl` columns specifically.
+
+```{r}
+human_zebrafish_key <- zebrafish %>%
+  # we'll want to subset zebrafish to only the columns we're interested in
+  dplyr::select(zebrafish_ensembl, human_symbol)
```
cansavvy, line 241: Can make this its own chunk? I see it as being its own step.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+# We can use the `count()` function to get a count of the distribution of values in the `human_symbol` column
+human_zebrafish_mapped_df %>%
+  dplyr::group_by(human_symbol) %>%
+  dplyr::count() %>%
+  # Sort by highest `n` which would be the human gene symbol with the most mapped zebrafish Ensembl IDs
+  dplyr::arrange(desc(n))
+```
+
+Looks like we have mapped symbols!
+We also have 415 NAs, which is okay, because we do not expect everything to map across species.
```
cansavvy, line 266:
```suggestion
We also have 415 NAs, which is okay, because we do not expect everything to map across species.
```
This finding, although good to have, is not clear from the output shown here, can add a small chunk that shows how to get this number?

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+
+## Sort out multi mappings
```
cansavvy, line 268:
```suggestion
## Sort out multi-mappings
```
Feels like this should have a hyphen? Whatever we have done elsewhere we should do here.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+Using another gene identifier may result in a better 1:1 ratio to the zebrafish Ensembl IDs, i.e. human Ensembl IDs instead of human gene symbols.
+See our [gene ID annotation microarray notebook](https://alexslemonade.github.io/refinebio-examples/02-microarrary/gene-id-annotation_microarray_01_ensembl.html) for an example of mapping between gene identifiers.
+
+* If a zebrafish Ensembl gene ID maps to multiple human symbols, the gene expression values get duplicated.
+Let's look at the `ENSDARG00000069142` example below.
+
+```{r}
+human_zebrafish_mapped_df %>%
+  dplyr::filter(Gene == "ENSDARG00000069142") %>%
+  dplyr::select(Gene, human_symbol, dplyr::everything())
```
cansavvy, line 281: Yes, I would do this step earlier so you only need to do it once. (see earlier comment)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+```
+
+* On the other, if a human gene symbol maps to multiple zebrafish Ensembl IDs, the gene expression values do not necessarily get duplicated.
```
cansavvy, line 284: They should not unless in the original `zebrrafish` dataset there was a different id that meant we needed duplicates. We will want to get rid of these instance where data is duplicated for no reason, I will comment on where I think that needs to happen.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: Also based on the output where we have duplicated values, this tells me that there are duplicated keys here (because remember `zebrafish` has more the `symbols` so they would repeat `zebrafish_ensembl` and  `human_symbol` for these instances. Because of this, we should make sure our key doesn't have replicates before we do a `join()`. Try this and let me know if that works. 
```suggestion
  dplyr::select(zebrafish_ensembl, human_symbol) %>% 
  dplyr::distinct() 
```
We will want to carefully explain what this `distinct()` and why it's necessary. May even want to take a minute to show the duplicated data before applying it.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: I _think_ if you use the change I suggest regarding `distinct()` you will be able to say this statement more absolutely. 
```suggestion
Remember that our original data is associate to zebrafish Ensembl IDs, so when a human gene symbol maps to multiple zebrafish Ensembl IDs, the data will not get duplicated, but you will have multiple rows associated with that human symbol. 
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+We also have 415 NAs, which is okay, because we do not expect everything to map across species.
+
+## Sort out multi mappings
+
+* Multiple zebrafish Ensembl gene IDs map to the same human symbol which results in duplicated human gene symbols.
+Some downstream tools may require the data be simplified to 1:1 mappings.
+Using another gene identifier may result in a better 1:1 ratio to the zebrafish Ensembl IDs, i.e. human Ensembl IDs instead of human gene symbols.
+See our [gene ID annotation microarray notebook](https://alexslemonade.github.io/refinebio-examples/02-microarrary/gene-id-annotation_microarray_01_ensembl.html) for an example of mapping between gene identifiers.
+
+* If a zebrafish Ensembl gene ID maps to multiple human symbols, the gene expression values get duplicated.
```
cansavvy, line 275:
```suggestion
If a zebrafish Ensembl gene ID maps to multiple human symbols, the gene expression values get duplicated.
```
Not sure if the bullet points really add to clarity in the rendering anymore.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+* On the other, if a human gene symbol maps to multiple zebrafish Ensembl IDs, the gene expression values do not necessarily get duplicated.
+Let's look at the `A1BG` example below.
+
+```{r}
+human_zebrafish_mapped_df %>%
+  dplyr::filter(human_symbol == "MATR3") %>%
+  dplyr::select(Gene, human_symbol, dplyr::everything())
+```
+
+## Collapse multiple mapped human gene symbols
```
cansavvy, line 293:
```suggestion
## Collapse multi-mapped human gene symbols
```
Whatever we call them, we should be consistent.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+
+We can collapse the multi-mapped values into a list for each Ensembl ID or we can keep it in the format above with one row for each mapped ID (resulting in duplicated rows for some Ensembl IDs).
```
cansavvy, line 295: Here we have a hyphen.

cansavvy, line 295: Since these multi-mappings can go either way `1zebra:manyhumans` or `manyzebras:1human`, we should try to make this wording more clear. 

Using a consistent wording/notation would be good. I think what I used above is kinda silly, but something like it might be good.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+readr::write_tsv(
+  collapsed_human_symbol_df,
+  file.path(
+    results_dir,
+    "GSE71270_zebrafish_ensembl_to_collapsed_human_gene_symbol.tsv" # Replace with a relevant output file name
+  )
+)
+```
+
+## Collapse multiple mapped zebrafish Ensembl IDs
```
cansavvy, line 325:
```suggestion
## Collapse multi-mapped zebrafish Ensembl IDs
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+
+Since multiple zebrafish Ensembl gene IDs map to the same human symbol, we can also collapse all Ensembl IDs that each unique gene symbol mapped to.
```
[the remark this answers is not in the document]
cansavvy: Maybe this kind of file is still useful? I don't know why it would be, though. Will need others to weigh in here.

cansavvy (commented):
This is coming along nicely, now I just have comments about wording since some of these concepts and steps can be confusing to follow. 

I think a consistent notation and wording might help since the multi-mappings we are talking about can be either `1zebra:manyhumans` or `manyzebras:1human` and we are dealing with both. 

I also am still not sure what to do with `manyzebras:1human` situation. I am tempted to do nothing with that and let users figure out what is most useful, but perhaps someone else might have a better suggestion for navigating what to do with data for `manyzebras:1human` situations. I don't want to encourage users to drop or improperly manipulate their data. But I am not sure what makes most sense.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,371 @@
...
+```
+
+Looks like we have mapped symbols!
+We also have 415 NAs, which is okay, because we do not expect everything to map across species.
+
+## Sort out multi mappings
+
+* Multiple zebrafish Ensembl gene IDs map to the same human symbol which results in duplicated human gene symbols.
+Some downstream tools may require the data be simplified to 1:1 mappings.
+Using another gene identifier may result in a better 1:1 ratio to the zebrafish Ensembl IDs, i.e. human Ensembl IDs instead of human gene symbols.
```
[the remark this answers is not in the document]
cbethell: I got rid of this part (and the sentence above it which I realize was duplicated). I can put it back if another reviewer finds it useful but I understand your point here and was not sure myself how useful it really was.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+We'll need to use the `devtools` package to install `hcop` from GitHub [@Wickham2020].
+
+```{r}
+if (!("devtools" %in% installed.packages())) {
+  install.packages("devtools")
+}
+```
+
+Now we can install `hcop` from Github.
+You can control what version of the package is installed by using the `ref` command of `devtools::install_github`.
```
cansavvy, line 164:
```suggestion
You can control what version of the package is installed by using the `ref` argument of `devtools::install_github()`.
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+Now, let's get an idea of how many zebrafish Ensembl IDs we have that were not mapped to human gene symbols.
+
+```{r}
+sum(is.na(human_zebrafish_mapped_df$human_symbol))
+```
+
+We have 415 NAs, which means we have 415 zebrafish Ensembl IDs that were not mapped to human gene symbols.
+This is okay because we do not expect everything to map across species.
+
+## Sort out multi-mappings
```
cansavvy, line 293:
```suggestion
## Take a look at some multi-mappings
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+
+If a zebrafish Ensembl gene ID maps to multiple human symbols, the gene expression values get duplicated.
+Let's look at the `ENSDARG00000069142` example below.
+
+```{r}
+human_zebrafish_mapped_df %>%
+  dplyr::filter(Gene == "ENSDARG00000069142")
+```
+
+Remember that our original data is associate to zebrafish Ensembl IDs, so when a human gene symbol maps to multiple zebrafish Ensembl IDs, the data will not get duplicated, but you will have multiple rows associated with that human symbol.
```
cansavvy, line 303:
```suggestion
Remember that our original data is associated to zebrafish Ensembl IDs, so when a human gene symbol maps to multiple zebrafish Ensembl IDs, the data will not get duplicated, but you will have multiple rows associated with that human symbol.
```
That was probably a typo I left in my suggestion?

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+human_zebrafish_key <- zebrafish %>%
+  # We'll want to subset zebrafish to only the columns we're interested in
+  dplyr::select(zebrafish_ensembl, human_symbol)
+
+# Since we ignored the additional columns in `zebrafish`, let's check to see if we have any duplicates in our `human_zebrafish_key`
+any(duplicated(human_zebrafish_key))
+```
+
+We do have duplicates!
+We don't want to handle duplicate data, so let's remove those duplicates before moving forward.
```
cansavvy, line 248: This is great. +1

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+Here's what the new data frame looks like:
+
+```{r}
+head(human_zebrafish_mapped_df, n = 25)
+```
+
+Let's get a summary of the gene symbols returned in the `human_symbol` column of our mapped data frame.
+
+```{r}
+# We can use the `count()` function to get a count of the distribution of values in the `human_symbol` column
```
cansavvy, line 274:
```suggestion
# We can use this `count()` function after `group_by()`to get a count of how many `zebrafish_ensembl` IDs there are per `human_symbol`s 
```
Trying to make this a tad more specific.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+  dplyr::select(Gene, human_symbol, dplyr::everything())
+```
+
+Here's what the new data frame looks like:
+
+```{r}
+head(human_zebrafish_mapped_df, n = 25)
+```
+
+Let's get a summary of the gene symbols returned in the `human_symbol` column of our mapped data frame.
```
cansavvy, line 271:
```suggestion
Let's get a summary of the human gene symbols returned in our mapped data frame, `human_zebrafish_mapped_df`. 
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+
+# Check if this is in the same order
+all.equal(colnames(df), metadata$geo_accession)
+
+# Bring back the "Gene" column in preparation for mapping
+df <- df %>%
+  tibble::rownames_to_column("Gene")
+```
+
+## Mapping zebrafish Ensembl gene IDs to human symbols
```
cansavvy, line 217:
```suggestion
## Mapping human gene symbols to zebrafish Ensembl gene IDs
```
I think ordering helps.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+
+```{r}
+# Let's take a look at the first 6 items in the `Gene` column of `df`
+head(df$Gene)
+```
+
+Ensembl gene identifiers have different species-specific prefixes. 
+In zebrafish, Ensembl gene identifiers begin with `ENSDARG` (in human, `ENSG`, etc.).
+
+Now let's do the conversion!
```
cansavvy, line 229:
```suggestion
Now let's do the mapping!
```

cansavvy (commented):
Okay, just a couple little wording comments and then we might want to get someone else to take a look.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+  html_notebook:
+    toc: true
+    toc_float: true
+    number_sections: true
+---
+
+# Purpose of this analysis
+
+This notebook demonstrates how you can use the [`hcop`](https://github.com/stephenturner/hcop) package to perform ortholog mapping for microarray data obtained from refine.bio.
+In this notebook, we will use zebrafish data from refine.bio and annotate it with human gene symbols from HGNC Comparison of Orthology Predictions (HCOP).
```
jashapiro, line 15: Worth linking to the original source here? https://www.genenames.org/tools/hcop/

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+In zebrafish, Ensembl gene identifiers begin with `ENSDARG` (in human, `ENSG`, etc.).
+
+Now let's do the mapping!
+We'll use the `zebrafish` data from `hcop`; this is what `zebrafish` looks like.
+
+```{r}
+zebrafish
+```
+
+We're interested in the `human_symbol` and `zebrafish_ensembl` columns specifically.
```
jashapiro, line 236: The power of `hcop` is that it is a compilation of multiple orthology predctions. Because of that, one of the most important parts of the data is the `support` column, so we should make use of that as well. This suggests a way to resolve some of the multimappings: take the one with the most support, or with support from the databases you trust. (Most support is probably the best choice for this notebook.)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,390 @@
...
+<!-- Do not delete this line --> <a name="analysis" style="padding-top:56px;margin-top:-56px;">&nbsp;</a>
+
+
+# Ortholog Mapping with `hcop` - Microarray
+
+## Install libraries
+
+See our Getting Started page with [instructions for package installation](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#what-you-need-to-install) for a list of the other software you will need, as well as more tips and resources.
+
+In this analysis, we will be using `hcop` to map between zebrafish and human identifiers [@Turner2018].
```
jashapiro, line 153: I think we need a lot more information here (or at the top) about what HCOP is.

jashapiro (commented):
I started to review this, but I want to propose some pretty radical (but less radical than they seem) changes.

The main thing is that we not want use the `hcop` package, as it hasn't been updated in a 3 years, so the data within it are probably well out of date. We also seem to be only using the data tables, not any of the functions from the package.

It should be pretty easy to regenerate the data, as the code is well described in https://github.com/stephenturner/hcop/blob/master/data-raw/hcop.R We could use that code as a template for downloading the data directly from HCOP proper, and we should *definitely* cite the package, but we might not even need to go that far.

The alternative is that we could send them to https://www.genenames.org/tools/hcop/, where the **Bulk Downloads** option will let them download the human-zebrafish table. 

We don't want to be downloading the data files repeatedly, so we will probably want to stash them somewhere in this repository and/or on S3, only downloading them if they don't exist.

The other high level point I want to make, (also called out in an individual comment) is that I think we need more description of what HCOP is and why we are using it. Orthology mapping is hard, and I think it is quite important that users not fall into a mode where they think it is a solved problem. We may present one way of doing it here, and one database to use, but that is far from the only way it could be done. Some people will even argue about whether it *should* be done (they go too far)!

cbethell:
@jashapiro, I believe I addressed all of your review comments. Please let me know if I misinterpreted any or if you believe that the context added should be more extensive.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,398 @@
...
+# Read in the data from HGNC
+zebrafish <-
+  paste0(
+    "ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/",
+    "human_",
+    "zebrafish", # Replace with the organism you have data for
+    "_hcop_fifteen_column.txt.gz"
+  ) %>%
+  data.table::fread() %>%
+  set_names(names(.) %>% gsub("_gene$", "", .)) # Removing some extra wording in some of the column names
```
cansavvy, line 181:
```suggestion
  set_names(names(.) %>% 
  gsub("_gene$", "", .)) # Removing some extra wording in some of the column names
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,398 @@
...
+HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes, including zebrafish, which we will use in this notebook [@hcop-help].
+
+In the next chunk, we have adapted code from the [`hcop` package R script](https://github.com/stephenturner/hcop/blob/master/data-raw/hcop.R) to retrieve the zebrafish and human mapped data from the HGNC database [@Turner2018].
+
+```{r}
+# Read in the data from HGNC
+zebrafish <-
+  paste0(
+    "ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/",
+    "human_",
```
cansavvy, line 176: It may be nice to have more advice around how one would figure out how to change this for other species and how to see what species and options are available.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,398 @@
...
+    "zebrafish", # Replace with the organism you have data for
+    "_hcop_fifteen_column.txt.gz"
+  ) %>%
+  data.table::fread() %>%
```
cansavvy, line 180: I would probably declare the URL in its own chunk and then read in with `data.table::fread()` in the next. This will be the first time (I think) that we use something besides `readr::read_tsv()` so we may want to give a quick comment on why we the change (I am assuming this file is large and that is why).

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,398 @@
...
+  set_names(names(.) %>% gsub("_gene$", "", .)) # Removing some extra wording in some of the column names
+```
+
```
cansavvy, line 183: I would print out a preview of the first few lines of `zebrafish` so users can see what it looks like since this is a new thing. (aka always be telling them to look at the data to get familiar)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,398 @@
...
+Ensembl gene identifiers have different species-specific prefixes. 
+In zebrafish, Ensembl gene identifiers begin with `ENSDARG` (in human, `ENSG`, etc.).
+
+Now let's do the mapping!
+We'll use the `zebrafish` data from the HGNC database; this is what `zebrafish` looks like.
+
+```{r}
+zebrafish
```
cansavvy, line 236: Oh I see you still have this here, probably just move this earlier then.

cansavvy (commented):
I just have a couple little comments for now. In addition to that you can add this notebook to the navbar in this PR (I know the recommendations on that were fluctuating a lot last week but now its set up so you can add to the navbar and everything won't be run just yet)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,404 @@
...
+```{r}
+collapsed_human_symbol_df <- human_zebrafish_mapped_df %>%
+  # Group by zebrafish Ensembl IDs
+  dplyr::group_by(Gene) %>%
+  # Collapse the mapped values in `human_zebrafish_mapped_df` into one column named `all_human_symbols`
+  dplyr::summarize(all_human_symbols = paste(human_symbol, collapse = ";")) %>%
+  # Now we need get the gene expression data back
+  dplyr::left_join(df, by = "Gene")
+
+head(collapsed_human_symbol_df)
```
jashapiro, line 338: We lose the `support` column here. That may not be a problem, but it should probably be noted.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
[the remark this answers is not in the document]
jashapiro: I think this really depends on the analysis. I can see times when you might want to see changes in any gene related to a human gene, so you would want to keep all rows, despite multiple mapping of human to <organism>. But you might want to drop things that are more likely to be spurious. So maybe not the most supporting databases, but maybe include ones that are in more than one? 

I do think we want to be careful here not to encourage people to throw away perfectly good data. What would it mean if there was a large change in expression in a gene that was one of 5 that mapped to a single human gene? Would we want to throw it out? Almost certainly not! 

On the other hand, we would probably would want to know if an orthology relationship was not well supported by the databases.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,404 @@
...
+A known issue in ortholog mapping is how to handle this data because combining values across zebrafish transcripts using an average may result in the loss of a lot of data and will likely not be representative of the zebrafish biology.
+Therefore, we will not add the expression data here since the expression values are not the same across Ensembl IDs, despite mapping to the same human gene symbol.
+
+Instead, we will use the `support` column to decide which mappings to retain.
+As we noted earlier, the prediction with the most number of databases which concur is considered the most reliable.
+Therefore, where we have multi-mapped zebrafish Ensembl gene IDs, we will take the mapping with the most databases that concur according to the `support` column.
+
+```{r}
+filtered_zebrafish_ensembl_df <- human_zebrafish_mapped_df %>%
+  dplyr::mutate(n_databases = stringr::str_count(support, ",") + 1) %>% # Count the number of databases in the support column for each prediction
```
jashapiro, line 368: Nice!

jashapiro (commented):
I like the changes that you have made. Always nice to reduce reliance on other packages when it makes sense! Some of the numbers did change though, so that is worth looking at.

One thing that I want to ask is why this is part of the Microarray section... we don't really use the expression data at all, so I am wondering if this should be done without that? I am partly thinking in response to the many fish:one human mappings. @cansavvy correctly points out that we don't want to throw out data, but I might go a step further and ask why we want to do this before differential expression analysis (or whatever other analysis) at all? (I suppose I can think of doing it for something like GSEA, where I might be able to get better annotations from the human data, but I would be pretty wary there too). 

Maybe people would want to see how a human expression dataset compared to zebrafish expression? Is that something we want to encourage? Here be monsters!

I think my point here is that we could probably simplify this by looking not at the whole expression data frame, but rather just at a list (or lists) of genes. Those might be a whole dataset, or they might be the output of a differential expression analysis, or whatever. In either case, this might allow consolidation of the two Ortholog Mapping pages into one.

jaclyn-taroni:
Regarding the point about consolidation - I thought including _both_ might be an opportunity for _variety_ in a way that might be helpful to users who specifically go through multiple examples. (Do those users exist? Great question.) 

I'll note that I haven't looked at what exactly this example is doing, but the general idea was "do ortholog mapping for two gene lists" (one for microarray, one for RNA-seq) without much focus on any kind of analysis.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,404 @@
...
+
+Looks like we have mapped symbols!
+
+Now, let's get an idea of how many zebrafish Ensembl IDs we have that were not mapped to human gene symbols.
+
+```{r}
+sum(is.na(human_zebrafish_mapped_df$human_symbol))
+```
+
+We have 464 NAs, which means we have 464 zebrafish Ensembl IDs that were not mapped to human gene symbols.
```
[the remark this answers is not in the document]
cbethell: I could not reproduce the 3538 number of NAs you got here. I did fix the typo here as I got 463 not 464, but would you be able to tell me the steps you took when you got 3538?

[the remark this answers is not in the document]
jashapiro: I was looking at the html file you had included. But now 463 does seem correct (I doubt it was a typo, probably it changed with the new data source!)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,404 @@
...
+# Read in the data from HGNC
+zebrafish <- data.table::fread(zebrafish_hgnc_url) %>% # Here we use `data.table::fread()` instead of `readr::read_tsv()` because we expect this file to be large so the `fread()` function is faster/more efficient here
+  set_names(names(.) %>%
+    gsub("_gene$", "", .)) # Removing some extra wording in some of the column names
+```
+
+Let's take a look at what `zebrafish` looks like.
+
+```{r}
+zebrafish
```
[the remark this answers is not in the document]
cbethell: I took a dive into the docs and found that `df_print` option seems like the best solution for this issue. I implemented `df_print: paged` in this notebook and added it to `scripts/render-notebooks.R`. Let me know if this seems like a good solution or if we should explore another.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,404 @@
...
+HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes, including zebrafish, which we will use in this notebook [@hcop-help].
+
+In the next chunk, we have adapted code from the [`hcop` package R script](https://github.com/stephenturner/hcop/blob/master/data-raw/hcop.R) to retrieve the zebrafish and human mapped data from the HGNC database [@Turner2018].
+
+```{r}
+# Define URL to desired organism ortholog mapping file from the HGNC database
+zebrafish_hgnc_url <- paste0(
+  "ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/",
+  "human_",
+  "zebrafish",
```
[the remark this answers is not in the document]
cbethell: Let me know what you think on the way I restructured this section @jashapiro, I have not yet added the file on s3 because I was not entirely sure if we would want to keep the file in `02-microarray/data` or if we would want to create a specific folder within data for this file, like `02-microarray/data/zebrafish-ortholog-mapping`.

cbethell:
> I think my point here is that we could probably simplify this by looking not at the whole expression data frame, but rather just at a list (or lists) of genes. Those might be a whole dataset, or they might be the output of a differential expression analysis, or whatever. In either case, this might allow consolidation of the two Ortholog Mapping pages into one.

@jashapiro, please let me know if I have appropriately addressed your review comments. I excluded the use of the expression values and focused on the list of genes like you mentioned above, and propagated changes as needed throughout (in addition to incorporating your other review suggestions).

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,413 @@
+---
+title: "Ortholog Mapping - Microarray"
+author: "CCDL for ALSF"
+date: "`r format(Sys.time(), '%B %Y')`"
+output:
+  html_notebook:
+    toc: true
+    toc_float: true
+    number_sections: true
+    df_print: paged
```
cansavvy, line 10: This will mostly be unnecessary because we render by the script. But if we want it to also show up for the users ini their locally rendering, then I suppose we should keep it, but then we need to add it to all the other modules.

scripts/render-notebooks.R
```diff
@@ -111,7 +111,8 @@ readr::write_lines(new_lines, tmp_file)
 rmarkdown::render(tmp_file,
   output_format = rmarkdown::html_document(
     toc = TRUE, toc_depth = 2,
-    toc_float = TRUE, number_sections = TRUE
+    toc_float = TRUE, number_sections = TRUE,
+    df_print = "paged"
```
jashapiro, line 115: Nice find!

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: If we do want to add this to all headers (which I'm 50-50 on) then we should file a separate "search and replace in the whole project" type issue where we implement this everywhere.e

cbethell: As noted by Josh in a Slack discussion, the notebook headers should not need to be changed as `paged_df` is the default for `html_notebook` so it will look the same either way for people running locally. 

I'll remove it from the notebook header in this PR but keep the change to the `render-notebooks.R` script so that the html output files render with this option.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,413 @@
...
+    "GSE71270_zebrafish_ensembl_to_collapsed_human_gene_symbol.tsv"
+  )
+)
+```
+
+## Collapse human genes mapping to multiple zebrafish genes
+
+Since multiple zebrafish Ensembl gene IDs map to the same human symbol, we can also collapse all Ensembl IDs that each unique human gene symbol mapped to.
+
+A known issue in ortholog mapping is how to handle this data because combining values across zebrafish transcripts using an average may result in the loss of a lot of data and will likely not be representative of the zebrafish biology.
```
jashapiro, line 358: I feel like this is still a bit unclear.. What are we trying to say here? What is the goal?

Are we are trying to find the "most similar" zebrafish gene to a given human gene? If so, looking at the level of support across databases might be one way to handle this, but it is not the only one. I think we should state very clearly here that this is not the only, or even recommended,  way to handle these multimappings.

I have written a bit below to try to get across my thoughts on the matter. 

```suggestion
Since multiple zebrafish Ensembl gene IDs map to the same human symbol, we may want to identify which one of these mappings represents the "true" ortholog, i.e. which zebrafish gene is most similar to the human gene we are interested in. 
*This is not at all straightforward!* (see, for just one example, https://doi.org/10.1093/bioinformatics/btaa468)
Gene duplications along the zebrafish lineage may result in complicated relationships among genes, especially with regard to divisions of function.

Simply combining values across zebrafish transcripts using an average may result in the loss of a lot of data and will likely not be representative of the zebrafish biology.
One thing we might do to make the problem somewhat simpler is to reduce the number of multimapped genes by requiring a certain level of support for each mapping from across the various databases included in `HCOP`.
This will not fully solve the problem (and may not even be desirable in some cases), but we present it here as an example of an approach one might take.
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,413 @@
...
+filtered_zebrafish_ensembl_df <- human_zebrafish_mapped_df %>%
+  # Count the number of databases in the support column for each prediction
+  dplyr::mutate(n_databases = stringr::str_count(support, ",") + 1) %>%
+  # Group by human gene symbols
+  dplyr::group_by(human_symbol) %>%
+  # Now filter for the rows with more than one database in support for each human gene symbol
+  dplyr::filter(n_databases > 1)
+
+head(filtered_zebrafish_ensembl_df)
+```
```
jashapiro, line 381: We might look to see before and after how many multimapped genes there are. I would expect it does not go to zero, and we should note that.

jashapiro (commented):
This is looking good! 

I made some suggestions for the multimapping discussion, where I think we can say a bit more about why this is *a whole thing* (I added one citation, but I could easily have added dozens!).

The other thing which is more practical is the question of how to get the data. I tend to think that the web interface is a bit more straightforward, but the annoying behavior of Safari insisting on decompressing `.gz` files on download is something we should keep in mind. I am not sure what happens on Windows machines! But I don't think that should preclude merging this PR. We should just be sure to file an issue about it so we can discuss the correct way to handle it. 

 As to your question here: https://github.com/AlexsLemonade/refinebio-examples/pull/246#discussion_r504081218, I have no preference as to the file structure... does @cansavvy ?

I also made a few minor changes here and there, including freezing the header date and switching to `session_info()` to parallel changes made in the template.

references.bib
```diff
@@ -482,12 +506,15 @@ @Article{Wickham2019
   doi = {10.21105/joss.01686},
 }
 
-@Manual{Wickham2020,
-  title = {devtools: Tools to Make Developing R Packages Easier},
-  author = {Hadley Wickham and Jim Hester and Winston Chang},
-  year = {2020},
-  note = {R package version 2.3.0},
-  url = {https://CRAN.R-project.org/package=devtools},
```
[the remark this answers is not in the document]
cbethell: None of the above actually. When I removed the devtools package installation from the notebook in this PR, I also removed the citation without accounting for the fact that it is also referenced in the getting started notebook. Good catch! I reverted this change in the last commit.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,413 @@
...
+The files available for download include files that have been mapped between human and another species using the [`hcop` package](https://github.com/stephenturner/hcop/) [@Turner2018].
+Ortholog species files with the '6 Column' output returns the raw assertions, Ensembl gene IDs and Entrez Gene IDs for human and one other species, while the '15 Column' output includes additional information such as the chromosomal location, accession numbers and the databases that support the assertions.
+
+To download a file, click the file name.
+For this notebook, you will want to download the file named `human_zebrafish_hcop_fifteen_column.txt.gz`.
+If you are using a different dataset, you can replace `zebrafish` in `human_zebrafish_hcop_fifteen_column.txt.gz` with the name of the species you have data for, and click on that file to download.
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/07499927a22ed7aa045245e40d692f8e49d7ba52/template/screenshots/download-hgnc-zebrafish-file.png" width=400>
+
+Now copy and paste the `human_zebrafish_hcop_fifteen_column.txt.gz` file into your `data/` folder.
```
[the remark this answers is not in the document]
cbethell: I made a note where relevant in the last commit and will file an issue for future changes to cover these _odd_ cases (as you put it) 👍

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cbethell: Let me know if what I added in the [last commit](https://github.com/AlexsLemonade/refinebio-examples/pull/246/commits/2717346bbcfcc5a55f001d39d027b10b40260a28#diff-6e1a150e702a9b814aaf52aa272f45dadb468d003ccbfcdbc6be1da2ccd5cb55R381) sufficiently addresses your comment here @jashapiro.

references.bib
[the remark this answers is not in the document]
jashapiro: In the future, we probably don't need to remove references from this file. Having a few unused extras floating around won't hurt anything!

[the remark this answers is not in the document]
cbethell: Gotcha and noted! 👍

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -380,6 +405,20 @@ filtered_zebrafish_ensembl_df <- human_zebrafish_mapped_df %>%
...
+filtered_zebrafish_ensembl_df %>%
+  # Group by human gene symbols
+  dplyr::group_by(human_symbol) %>%
+  # Count the number of rows in the dataframe for each symbol
+  dplyr::count() %>%
+  # Filter out the symbols without multimapped genes
+  dplyr::filter(n > 1)
+```
+Now we only have 1,803 multi-mapped genes, compared to the 4,192 that we began with.
+Although we haven't filter down to zero multi-mapped genes, we were able to filter down to a more reasonable number of _reliable_ mappings.
```
jashapiro, line 419: These additions look good. Just a small wording suggestion below.
```suggestion
Although we haven't filtered down to zero multi-mapped genes, we have hopefully removed some of the less _reliable_ mappings.
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+# How to run this example
+
+For general information about our tutorials and the basic software packages you will need, please see our ['Getting Started' section](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#how-this-tutorial-is-structured).
+We recommend taking a look at our [Resources for Learning R](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#resources-for-learning-r) if you have not written code in R before. 
+
+## Obtain the `.Rmd` file
+
+To run this example yourself, [download the `.Rmd` for this analysis by clicking this link](https://alexslemonade.github.io/refinebio-examples/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd).
+
```
cansavvy, line 27:
```suggestion
Clicking this link will most likely send this to your downloads folder on your computer. 
Move this `.Rmd` file to where you would like this example and its files to be stored.
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+You can open this `.Rmd` file in RStudio and follow the rest of these steps from there. (See our [section about getting started with R notebooks](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#how-to-get-and-use-rmds) if you are unfamiliar with `.Rmd` files.)
+Clicking this link will most likely send this to your downloads folder on your computer.
```
cansavvy, line 29:
```suggestion
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+We recommend taking a look at our [Resources for Learning R](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#resources-for-learning-r) if you have not written code in R before. 
+
+## Obtain the `.Rmd` file
+
+To run this example yourself, [download the `.Rmd` for this analysis by clicking this link](https://alexslemonade.github.io/refinebio-examples/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd).
+
+You can open this `.Rmd` file in RStudio and follow the rest of these steps from there. (See our [section about getting started with R notebooks](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#how-to-get-and-use-rmds) if you are unfamiliar with `.Rmd` files.)
+Clicking this link will most likely send this to your downloads folder on your computer. 
+Move this `.Rmd` file to where you would like this example and its files to be stored.
```
cansavvy, line 30:
```suggestion
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: In accordance with #302

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: In accordance with #302

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+# Session info
+
+At the end of every analysis, before saving your notebook, we recommend printing out your session info. 
+This helps make your code more reproducible by recording what versions of software and packages you used to run this. 
+
+```{r}
+# Print session info
+sessioninfo::session_info()
+```
```
cansavvy, line 451: It won't let me add a line because of the backticks here, but here we will need to add a new line and then `# References` and a new line. In accordance with #301

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+## Obtain the dataset from refine.bio
+
+For general information about downloading data for these examples, see our ['Getting Started' section](https://alexslemonade.github.io/refinebio-examples/01-getting-started/getting-started.html#how-to-get-the-data). 
+
+Go to this [dataset's page on refine.bio](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process).
+
+Click the "Download Now" button on the right side of this screen.  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-now.png" width=200>
```
cansavvy, line 73:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/download-now.png" width=200>  
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+Go to this [dataset's page on refine.bio](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process).
+
+Click the "Download Now" button on the right side of this screen.  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-now.png" width=200>  
+
+Fill out the pop up window with your email and our Terms and Conditions:  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-email.png" width=500>
```
cansavvy, line 77:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/download-email.png" width=500>  
```

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+For this example analysis, we will use this [CREB overexpression zebrafish dataset](https://www.refine.bio/experiments/GSE71270/creb-overexpression-induces-leukemia-in-zebrafish-by-blocking-myeloid-differentiation-process).
+@Tregnago2016 measured microarray gene expression of ten zebrafish samples, five overexpressing human CREB, as well as five control samples.
+
+## Place the dataset in your new `data/` folder
+
+refine.bio will send you a download button in the email when it is ready. 
+Follow the prompt to download a zip file that has a name with a series of letters and numbers and ends in `.zip`. 
+Double clicking should unzip this for you and create a folder of the same name.  
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/download-folder-structure.png" width=400>
```
cansavvy, line 93:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/download-folder-structure.png" width=400> 
```
#300

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+- A folder called "data" which contains:
+  - The `GSE71270` folder which contains:
+    - The gene expression  
+    - The metadata TSV  
+- A folder for `plots` (currently empty)  
+- A folder for `results` (currently empty)  
+    
+Your example analysis folder should now look something like this (except with respective experiment accession id and analysis notebook name you are using): 
+
+<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/e140face75daa6d2c34e30a4755c362e6039a677/template/screenshots/analysis-folder-structure.png" width=400>
```
cansavvy, line 116:
```suggestion
<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/analysis-folder-structure.png" width=400>
```
#300

cansavvy (commented):
I'm commenting here so we keep this module updated with the template before merging some minor changes that we'll need to do based on #300 #301 #302 

@cbethell Let me know how this suggestion tactic works for you or if you end up preferring me commit changes directly on #267 I don't mind either way.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
cansavvy: For these, its easiest to search and replace the commit id.

jashapiro (approved):
Looks good! Lets get it in!

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -0,0 +1,451 @@
...
+
+In order for our example here to run without a hitch, we need these files to be in these locations so we've constructed a test to check before we get started with the analysis. 
+Run this chunk to double check that your files are in the right place. 
+
+```{r}
```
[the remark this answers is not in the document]
cansavvy: Forget this comment since this is just about ready for merging.

PR #248, opened 2020-09-29
GitHub Actions: Styler 

### LINKED ISSUE #231: Automation: run `styler` with GitHub Actions
Related to #228 and the closed issue #154 that went a bit off track...

`styler` is run as part of the Snakemake workflow: https://github.com/AlexsLemonade/refinebio-examples/blob/40aa22a415ec5a92c3f31fc92965822ab4d03b4f/Snakefile#L17 and https://github.com/AlexsLemonade/refinebio-examples/blob/40aa22a415ec5a92c3f31fc92965822ab4d03b4f/scripts/render-notebooks.R#L85

You might think makes adding `styler` to GitHub Actions redundant and I would not disagree with you! _However_, in our quest for not making human pull request reviewers (and authors) do things that computers can do for us, it would be useful to break that ultimate goal into tasks that are relatively small in scope so we can get started.

I believe running `styler` in GitHub Actions would require us to do the following:

1. Install `styler`
2. Run `styler` on :all_the_things:
3. Commit the styled results and push

Here's an example I found: https://github.com/r-lib/actions/tree/master/examples#when-it-can-they-be-used

cansavvy:
## Purpose:
Closes #231 

This PR adds a step to the GHA workflow that styles all the Rmd files upon a PR or merge with master. 
It also includes a slight modification to `docker-build.yml` so that builds only happen when changes to the Dockerfile occur. 

`styler.yml` defined GHAs pull the rocker/tidyverse image and runs it on there since `styler` needs `knitr` for styling Rmds and we also need `git` for the committing step. The commit that says `Style Rmds` is where that occurred. 

It is likely that this `styler.yml` file will later be modified so it handles spell checking as well, but that will be in a future PR.

### DIFF
--- a/.github/workflows/styler.yml
+++ b/.github/workflows/styler.yml
@@ -13,8 +13,8 @@ on:
 jobs:
   # This workflow contains a single job called "build"
   styler:
-    # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    container:
+      image: ccdl/refinebio-examples:latest
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
--- a/.github/workflows/styler.yml
+++ b/.github/workflows/styler.yml
@@ -15,11 +15,15 @@ jobs:
   styler:
     runs-on: ubuntu-latest
     container:
-      image: ccdl/refinebio-examples:latest
+      image: rocker/r-ver:4.0.2
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
+
+      - name: Install styler
+        run: Rscript -e "install.packages('styler')"
+
       - name: Run styler
-        run: Rscript -e "styler::style_files(list.files(pattern = 'Rmd$', recursive = TRUE, full.names = TRUE))"
+        run: Rscript -e "styler::style_file(list.files(pattern = 'Rmd$', recursive = TRUE, full.names = TRUE))"
--- a/.github/workflows/styler.yml
+++ b/.github/workflows/styler.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install styler
-        run: Rscript -e "install.packages('styler')"
+        run: Rscript -e "install.packages(c('knitr', 'styler'))"
 
       - name: Run styler
         run: Rscript -e "styler::style_file(list.files(pattern = 'Rmd$', recursive = TRUE, full.names = TRUE));warnings()"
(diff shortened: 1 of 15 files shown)


jashapiro:
It looks like this is not quite working as expected. For the pull request, it seems not be getting the branch info, so it is not pushing changes back to the branch.

On checkout, we get:
```
 /usr/bin/git checkout --progress --force refs/remotes/pull/248/merge
  Note: switching to 'refs/remotes/pull/248/merge'.
  
  You are in 'detached HEAD' state. You can look around, make experimental
  changes and commit them, and you can discard any commits you make in this
  state without impacting any branches by switching back to a branch.
  
  If you want to create a new branch to retain commits you create, you may
  do so (now or later) by using -c with the switch command. Example:
  
    git switch -c <new-branch-name>
  
  Or undo this operation with:
  
    git switch -
  
  Turn off this advice by setting config variable advice.detachedHead to false
  
  HEAD is now at 84e5dbe Merge d5b1720e4d13d860c2486145f82dbb9e026aa6e6 into 7b90e967f96ed6846cb77c40bcdeb48bb3fc13ea
/usr/bin/git log -1 --format='%H'
'84e5dbe88e18e616d121d5323033e0a4ba11ed2a'
```

Then on commit, we get the following:
```
[detached HEAD 7db424c] Style Rmds
 12 files changed, 657 insertions(+), 469 deletions(-)
fatal: You are not currently on a branch.
To push the history leading to the current (detached HEAD)
state now, use
```

cansavvy:
> It looks like this is not quite working as expected. For the pull request, it seems not be getting the branch info, so it is not pushing changes back to the branch.

I saw this, I'm trying to use the `with:` `ref:` thing mentioned here: https://github.com/actions/checkout/issues/124

jashapiro:
> > It looks like this is not quite working as expected. For the pull request, it seems not be getting the branch info, so it is not pushing changes back to the branch.
> 
> I saw this, I'm trying to use the `with:` `ref:` thing mentioned here: [actions/checkout#124](https://github.com/actions/checkout/issues/124)

I was just going to point to that... seems to have worked!

cansavvy:
I think this is ready for review, but I don't know what the status of the docker-build stuff is, @jashapiro ? Do we still want the changes to docker-build.yml you had originally said and that we made here?

.github/workflows/styler.yml
```diff
@@ -0,0 +1,38 @@
...
+  styler:
+    runs-on: ubuntu-latest
+    container:
+      image: rocker/tidyverse:4.0.2
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
```
jashapiro, line 24: Do we want a comment to explain this?

jashapiro (approved):
This looks good to me! And it seems to do what we want, so let's add it in!

PR #250, opened 2020-09-29
Remove install2.r from dockerfile and test dockerhub push action

jashapiro:
## Purpose:
This PR is primarily designed to test #238 (automatic pushing to dockerhub) , but along the way, it closes https://github.com/AlexsLemonade/refinebio-examples/issues/235

## Strategy
I removed calls to `install2.R` and replaced them with the more verbose, but built-in `install.packages()` function. The reasoning here is that this should result in a somewhat smaller docker image, as `install2.R` seems to retain the cached versions of packages, rather than deleting them.

Did I miss any important packages?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -63,7 +63,6 @@ RUN R -e "BiocManager::install( \
 # Installs packages needed for plottings
 # treemap, interactive plots, and hex plots
 # Rtsne and umap are required for dimension reduction analyses
-# org.Mm.eg.db is required for gene mapping
 RUN Rscript -e  "install.packages( \
     c('ggfortify', \
       'ggsignif', \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,39 +33,50 @@ RUN pip3 install "setuptools==49.3.0" "six==1.15.0" "wheel==0.34.2"
 ###############
 
 # Commonly used R packages
-RUN install2.r --error --deps TRUE \
-    cluster \
-    GGally \
-    optparse \
-    R.utils \
-    RColorBrewer \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -33,39 +33,50 @@ RUN pip3 install "setuptools==49.3.0" "six==1.15.0" "wheel==0.34.2"
...
 # org.Mm.eg.db is required for gene mapping
-RUN install2.r --error --deps TRUE \
-    ggfortify \
-    ggsignif \
-    patchwork \
-    pheatmap \
-    Rtsne \
-    umap  \
-    VennDiagram
+RUN Rscript -e  "install.packages( \
```
jaclyn-taroni, line 67: Can you remove the outdated comment above this line?

jaclyn-taroni (approved):
👍

docker/Dockerfile
```diff
@@ -33,39 +33,50 @@ RUN pip3 install "setuptools==49.3.0" "six==1.15.0" "wheel==0.34.2"
...
+      'limma', \
+      'marray', \
+      'org.Mm.eg.db', \
+      'org.Dr.eg.db'), \
+     update = FALSE)"
 
 # Installs packages needed for plottings
 # treemap, interactive plots, and hex plots
 # Rtsne and umap are required for dimension reduction analyses
 # org.Mm.eg.db is required for gene mapping
```
jashapiro, line 66:
```suggestion
```

docker/Dockerfile
jashapiro: I was doing that as you reviewed! With those exact words.

Issue #239, opened 2020-09-27
Automation: Spell checking files


jaclyn-taroni:
Related to #228 - The PR checklist is long and it's hard to make sure that more than the _initial submission_ adheres to the checklist.

I'll note that I think this is lower priority than some other `automate all the things` items. My intention, in part, is to illustrate how processes are intended to help us and ultimately the folks that are going to use these examples.

### Context

One of the items in the pull request checklist is 

> Spell checked any Rmd file or md file.

Where [the instructions for how to run spell check](https://github.com/AlexsLemonade/refinebio-examples/blob/7b90e967f96ed6846cb77c40bcdeb48bb3fc13ea/CONTRIBUTING.md#how-to-spell-check) are

> In R, run the following:
> ```
> dictionary <- readLines(file.path("components", "dictionary.txt"))
> spelling::spell_check_files("<tech-section>_<file_name>.Rmd",
>                             ignore = dictionary)
> ```
> Any terms that should be recognized by the spell check, you can add to `components/dictionary.txt`.

The purpose of using a spell checking mechanism is, in part, to help us enforce [formatting of typical words/items](https://github.com/AlexsLemonade/refinebio-examples/blob/7b90e967f96ed6846cb77c40bcdeb48bb3fc13ea/CONTRIBUTING.md#formatting-of-typical-wordsitems). Here's a selection of the items listed there: 

>   - Use "refine.bio", NOT "refinebio"
>   - Use `.Rmd`,  NOT "Rmd" or ".Rmd"
>   - Use "tidyverse", NOT "Tidyverse"
>   - Use "TSV",  NOT tsv or `tsv` or .tsv
>   - Use "PNG", NOT png or `png` or .png (and etc.)

A large reason **why** we have those guidelines has to do with the _intended audience_ for these examples. Using these terms/formatting inconsistently can be unnecessarily confusing for someone being introduced to them for the first time.

### Problem

To understand how we might automate spell checking, I followed the spell check instructions for a number of Rmd files for the microarray examples. There were a number of misspellings according to `spelling::spell_check_files()`. As a result, I had some observations.

1.  "Rmd" was flagged as a misspelling consistently at the _beginning_ of notebooks. "`.Rmd`" is included in the custom dictionary, but "Rmd" is not: https://github.com/AlexsLemonade/refinebio-examples/blob/7b90e967f96ed6846cb77c40bcdeb48bb3fc13ea/components/dictionary.txt#L8

[...]

PR #253, opened 2020-09-29
Automate spell checking! - PR1

cansavvy:
## Purpose:
The first PR noted in https://github.com/AlexsLemonade/refinebio-examples/issues/239#issuecomment-700879194

This introduces spell checking to the Github actions workflow, however, it is set really leniently because I have not gone through to address the spelling errors (by fixing them or adding to `dictionary.txt`) that will be in the upcoming PR. 

I also updated CONTRIBUTING.md with instructions for how one can use spell check or retrieve the list of misspelled words that this will archive to GitHub actions.


components/spell-check.R
```diff
@@ -0,0 +1,22 @@
...
+# Run spell check
+sp_errors <- spelling::spell_check_files(files, ignore = dictionary) %>% 
+  data.frame() %>%
+  tidyr::unnest(cols = found)
+
+# Print out how many spell check errors
+write(nrow(sp_errors), stdout())
+
+# Save spell errors to file temporarily
+readr::write_tsv(sp_errors, file.path('components', 'spell_check_results.tsv'))
```
jashapiro, line 22: I would suggest either making this an option or outputting directly to the current directory. No guarantees of where this will be run from, so expecting a folder can seem dangerous.

components/spell-check.R
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env Rscript
+#
+# Run spell check and save results
+
+library(magrittr)
+
+# Read in dictionary
+dictionary <- readLines(file.path('components', 'dictionary.txt'))
```
jashapiro, line 8: Same thing here: I would make this an option, perhaps with this default location, but use `rprojroot` at a minimum.

.github/workflows/style-and-sp-check.yml
```diff
@@ -24,8 +24,30 @@ jobs:
...
+          path: components/spell_check_results.tsv
+
+      # If there are too many spell checks, this will stop everything
+      - name: Check on spell check results
+        if: ${{ steps.spell_check_run.outputs.sp_chk_results > 340 }}
+        run: |
+          echo Too many spelling errors, $result
+          exit 1
+        env:
+          result: ${{ steps.spell_check_run.outputs.sp_chk_results }}
```
jashapiro, line 50: I think we should print out the spelling errors here, since it will be easier to see them here than by downloading a zip file, etc. Which means I think we should make this run every time, and use the shell script conditional to kill the workflow.

Also, I moved the `env:` declaration up for my preferences (I want to see variables declared before use).

```suggestion
      - name: Check on spell check results
        env:
          result: ${{ steps.spell_check_run.outputs.sp_chk_results }}
        run: |
          cat components/spell_check_results.tsv
          if (( $result > 340 )); then
            echo Too many spelling errors, $result
            exit 1
          fi
```

.github/workflows/style-and-sp-check.yml
```diff
@@ -24,8 +24,30 @@ jobs:
...
-      - name: Install styler
-        run: Rscript -e "install.packages('styler')"
+      - name: Install packages
+        run: Rscript -e "install.packages(c('styler', 'spelling'))"
+
+      - name: Run spell check
+        id: spell_check_run
+        run: |
+          Rscript components/spell-check.R
+          results=$(Rscript "components/spell-check.R")
```
jashapiro, line 34: No need to run the script twice!
```suggestion
          results=$(Rscript "components/spell-check.R")
```

jashapiro (commented):
This looks good!
I made a few suggestions, like having the spelling output print to stdout in the "Check" step. I think that will be easier for people to use than downloading the zip file, etc. This change should be reflected in the instructions as well. (Keeping the results in the zip file is fine, too, but options are good!)

I also mad a kind of hacky way to avoid the "blank" lines in the spellcheck results, but it works.

Finally I had opinions™ about where files should live and options for the script, since you are suggesting people can (and should) run it on their own.

components/spell-check.R
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env Rscript
```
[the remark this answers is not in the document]
cansavvy: Should I still have the output spell check file locally go to `components`?

jashapiro:
Since this is in a PR, you should not need to do this: https://github.com/AlexsLemonade/refinebio-examples/pull/253/commits/4193ee373dde25e44d55f3db479868a436f89717

cansavvy:
> Since this is in a PR, you should not need to do this: [4193ee3](https://github.com/AlexsLemonade/refinebio-examples/commit/4193ee373dde25e44d55f3db479868a436f89717)

I didn't see the github action I thought I would see show up right away, it must have been on a delay and I was too impatient. Will undo that.

components/spell-check.R
cansavvy: I was thinking I should do this but then I didn't. Good you mentioned it then.

scripts/spell-check.R
```diff
@@ -0,0 +1,29 @@
...
+sp_errors <- spelling::spell_check_files(files, ignore = dictionary) %>% 
+  data.frame() %>%
+  tidyr::unnest(cols = found) %>%
+  tidyr::separate(found, into = c("file", "lines"), sep = ":")
+
+# Print out how many spell check errors
+write(nrow(sp_errors), stdout())
+
+# Save spell errors to file temporarily
+readr::write_tsv(sp_errors, file.path(root_dir, 'spell_check_results.tsv'))
```
jashapiro, line 29: If I am running this myself, I would expect the results to end up in my current directory, not the project root.

```suggestion
readr::write_tsv(sp_errors, 'spell_check_results.tsv')
```

.github/workflows/style-and-sp-check.yml
```diff
@@ -24,8 +24,31 @@ jobs:
...
+        id: spell_check_run
+        run: |
+          results=$(Rscript "scripts/spell-check.R")
+          echo "::set-output name=sp_chk_results::$results"
+
+      - name: Archive spelling errors
+        uses: actions/upload-artifact@v2
+        with:
+          name: spell-check-results
+          path: scripts/spell_check_results.tsv
```
jashapiro, line 40:
```suggestion
          path: spell_check_results.tsv
```

.github/workflows/style-and-sp-check.yml
```diff
@@ -24,8 +24,31 @@ jobs:
...
+          results=$(Rscript "scripts/spell-check.R")
+          echo "::set-output name=sp_chk_results::$results"
+
+      - name: Archive spelling errors
+        uses: actions/upload-artifact@v2
+        with:
+          name: spell-check-results
+          path: spell_check_results.tsv
+
+      # If there are too many spell checks, this will stop everything
```
jashapiro, line 42:
```suggestion
      # If there are too many spelling errors, this will stop the workflow
```

CONTRIBUTING.md
```diff
@@ -259,16 +259,17 @@ For example:
...
-```
-dictionary <- readLines(file.path("components", "dictionary.txt"))
-spelling::spell_check_files("<tech-section>_<file_name>.Rmd",
-                            ignore = dictionary)
-```
+Spell checks are run automatically using Github actions upon opening a PR for master or prior to merging to master.
+Github actions will abort if there are more than 2 spelling errors and you will need to fix those before continuing.
+You can obtain the list of spelling errors on Github by going to `Actions` and clicking the workflow of PR you are working on.
+Click on the `style-n-check` step and in the upper right hand corner, there is a button that says "Artifacts" which should list a file called `spell-check-results`.
+Click on `spell-check-results` to download a zip file that contains the list of misspelled words.
```
jashapiro, line 268:
```suggestion
Click on `spell-check-results` to download a zip file that contains the list of misspelled words.
Alternatively, click on the "Check on spell check results" step in the workflow log to see the misspellings.
```

jashapiro (approved):
Looks good. I just added a comment about how to see the results without downloading the zip file.

PR #254, opened 2020-09-30
Rebuild all html files using snakemake

jashapiro:
## Purpose:

This is one of the big goals of #228: Automating the rendering of html files in the docker container.

I think it is mostly working (not totally done yet, so a draft, but I wanted to get it on the books)

## Strategy

Since we always want to use the latest docker image, I made this part of the docker build automation. But I made it smart (or at least I tried to)!

It checks if the Dockerfile has changed from before the current push to the current one, and if there is a change, it rebuilds the docker image and pushes the new one to Dockerhub. See comments below about how this check is done. 

It then uses that docker image, or the latest downloaded from Dockerhub if there have been no changes since the last push, to render all of the html files as specified in the `Snakefile`. The snakefile currently does both styling and rendering, but with other automation in #248, we will probably want to remove that from here. 

Currently, rendering seems to take about 5 minutes, including downloading data and the docker image, which isn't too bad! Obviously if we have to rebuild the docker image it will be slower, but that should not be too frequent.

I am not totally sure how robust my checks for changes in the Dockerfile are. I am using `git diff-list` comparing the current version to what github reports as the "before" SHA. In a branch with many changes like the development branch here, that is usually comparing to the previous commit, but I think with the `master` branch it should always be the one before a pull request is merged. 

As you can see from the commit list, there were definitely some fits and starts getting this working. (It seems to be hard to set a boolean output value from within a `run:` block.) It may be worth making a new branch and cleanly moving in the changes... or we can leave the mess. 

## TODOs:
- [x] re-enable pushing to Dockerhub (disabled for testing)
- [x] push `.html` changes back to the branch (disabled for testing; they are currently committed but not pushed upstream)


jashapiro:
Should this action force rebuilding of all the html files? It currently does not, but given the way snakemake handles files, it probably ought to.

jaclyn-taroni:
> Should this action force rebuilding of all the html files? It currently does not, but given the way snakemake handles files, it probably ought to.

That is the behavior I would expect based on what I understand that we are trying to accomplish with this change.

.github/workflows/docker-build-push.yml
```diff
@@ -1,38 +1,80 @@
...
+      # download data
+      - name: Download data
+        run: bash scripts/download-data.sh
+      # render pages with snakemake in the docker image
+      - name: Render all pages to html
+        run: |
+          docker run \
+          --mount type=bind,target=/home/rstudio,source=$PWD \
+          ccdl/refinebio-examples \
+          snakemake --cores 2 --forceall
```
cansavvy, line 69:
```suggestion
          snakemake --cores 2 --forceall
```
I think at some point you had advised bumping it up to `4`, but I don't know what GHA's machinery is like so I'm suspecting you are keeping it at `2` for a reason.

cansavvy (commented):
I can't tell with all the files changed, did we want to shut off the `--style` option in `snakemake`? Or are we okay with that both style-and-sp-check.yml and the file at hand are both styling and perhaps being a bit redundant? 

Also can we determine an order with this rendering? Aka, how do we know that the rendering occurs with the spell-check and styling changes (if we remove the styling step as I mentioned above).

> It may be worth making a new branch and cleanly moving in the changes... or we can leave the mess.

Is this `mess` you are referring to just all the commits? 

I'm assuming you are leaving this `re-enable pushing to Dockerhub (disabled for testing)` for the very last thing?

.github/workflows/docker-build-push.yml
jashapiro: Yeah, I think we only get one cpu on github actions. 2 threads may speed things up a touch with little risk, so I did that.

jashapiro:
> Is this `mess` you are referring to just all the commits?

Yes, all the commits. I left the files that changed upon rendering as illustration for other discussions, and to allow reviewers to verify that nothing untoward was happening as a result of these changes. 

> I'm assuming you are leaving this `re-enable pushing to Dockerhub (disabled for testing)` for the very last thing?

I just hadn't checked this off. It was re-enabled here https://github.com/AlexsLemonade/refinebio-examples/pull/254/commits/4784ae260c506d476ce93909520fdf9e5115aaff after removing this branch as a trigger.

jashapiro:
> I can't tell with all the files changed, did we want to shut off the `--style` option in `snakemake`? Or are we okay with that both style-and-sp-check.yml and the file at hand are both styling and perhaps being a bit redundant?
> 
> Also can we determine an order with this rendering? Aka, how do we know that the rendering occurs with the spell-check and styling changes (if we remove the styling step as I mentioned above).

I had been considering that, but it takes almost no time relatively speaking, so I didn't. With automated styling going on, I don't think this should ever make changes though, so we could turn it off.

As far as ordering goes; the submission and any updates to the PR will initiate spell checks and styling. So any change should have happened before a merge to master. The merge to master will trigger this workflow, which will then rerender and commit its changes directly to the master branch. At least that is the plan. I have a fear that the fact that the master branch is protected may result in an error when we do this in practice.

cansavvy (approved):
Okay then. Seems like this is/should work as we want it to with the caveat that we may want to think about trimming out styling step with the `--style` option but we can revisit that later.

PR #257, opened 2020-10-01
PR2: Fix spelling errors and add things to dictionary.txt

### LINKED ISSUE #239: Automation: Spell checking files

### LINKED ISSUE #243: Style add/update: `IDs` NOT `ids` or `Ids`
In the most recent `gene-id-annotation` examples @cbethell, she went with `IDs` for the standard style (which we did not have addressed in the style guide at the time). 

We need to update the CONTRIBUTING.md and search and replace all the instances where `id` or `Id` or any other, now obsolete style of this is being used, most importantly any instances in the template. And then we need to re-render everything of course.

### LINKED ISSUE #244: Template fix: `softwares` -> `software`
Turns out the plural of `software` is `software`. https://github.com/AlexsLemonade/refinebio-examples/pull/233#discussion_r496044013 

This was in the template from the beginning so it is now everywhere in this repository. We will need to search and replace throughout the project.

cansavvy:
## Purpose:
The second PR mentioned on #239 where I fix all the spelling errors and add the words to the dictionary.txt that we need. 

As a side consequence, `softwares` (#244) and  `ID` (#243) and Rmd -> `.Rmd` inconsistencies have been resolved and added to the CONTRIBUTING.md if they were not previously there. 

Closes #243
Closes #244
Closes #239

## Strategy
A lot of iterations of running spell check, searching in the project for where those words occurred and determing whether they were legit words and should be added to the dictionary.txt or if they were legit errors and in that case I fixed them. 

For non-updated notebooks that had manual citations eg. `Yaari et al, _PLoS_` I deleted those because they will be done automatically, but I kept the paper links so that when we do get to updating that notebook we know what the paper and its reference should be. 

Unfortunately `spelling::spell_check_files()` isn't great with hyphens so you will note I put fake words like `vivo` even though I really wanted it to put `in-vivo` but it would not take that. 

### For reviewers: 
I wouldn't worry about examples that have not been updated yet. I've basically made them so they will pass but they will be further refined when we get to those examples. I would mostly take a look at dictionary.txt and CONTRIBUTING.md and see if that makes sense, as well as the spelling fixes you will see sprinkled throughout. 


components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
+actin
+ADT
+al
 al.
```
jashapiro, line 4: I don't think we want both of these, and only "al." is correct, as it is an abbreviation.
```suggestion
al.
```

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+AML
+AnaLysis
+ASXL
+Azacytidine
+arounds
```
jashapiro, line 9: Convince me this is correct?
```suggestion
```

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+DESeqDataset
+devtools
+DGE
+directionality
+DocToc
+Dorsoventral
+ECM
+edgeR
+edgeR's
+eg
```
jashapiro, line 41: This should always be "e.g." I suspect?
```suggestion
```

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+ENSDARG
+Ensembl
+ENSG
+Entrez
+et
 et.
```
jashapiro, line 47: Should not need both. But also, "et." is likely to be wrong: "et" is a word on its own, not an abbreviation.
```suggestion
et
```

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+subtypes
 tidyverse
+TPM
+TPMs
+transcriptional
+transcriptome
+transcriptomic
 TSV
+tximport
+ucture
```
jashapiro, line 122: Is this correct?

jashapiro (commented):
I made a number of suggestions to the spelling dictionary, which are best addressed individually below.

I also have a couple of higher level questions about how the spellcheck actually works: do we need plurals in cases of words like "subtype", or does adding that give us "subtypes" for free (if I wrote a spellchecker...I am not sure which decision I would make). Similarly, possessives: if we have "dataset", do we need "dataset's"?

Seeing some of those, I also want to

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+ColorBrewer
+ComplexHeatmap
+CREB
+Crouser
+cytosolic
+Danio
 dataset
 dataset's
+decitabine
 DESeq
```
jashapiro, line 30: I notice that most function names are not here, which leads me to think that `DESeq()` will not trigger a spelling error. Are we using `DESeq` on its own in a different way?

components/dictionary.txt
cansavvy: Ugh this is because of its problem with hyphens. I wanted to put here `work-arounds` but it would not work.

components/dictionary.txt
cansavvy: We can decide in the style guide. In most places we have an italicized `eg` but I am fine with changing it to what we deem appropriate.

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+ENSDARG
+Ensembl
+ENSG
+Entrez
+et
 et.
+FACS
+functionalize
+FPKM
+GEne
```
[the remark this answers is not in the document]
cansavvy: It does, but its actually an abbreviation in `WebGestaltR` but I don't know how else to get around it. https://github.com/AlexsLemonade/refinebio-examples/blob/e630a4eec112271038f7c1a1a645f533a119fb3b/02-microarray/pathway_analysis_microarray_02_ora_with_webgestaltr.Rmd#L132

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+genesets
```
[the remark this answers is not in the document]
cansavvy: That is fine.

components/dictionary.txt
cansavvy: See my comment above about: https://github.com/AlexsLemonade/refinebio-examples/blob/e630a4eec112271038f7c1a1a645f533a119fb3b/02-microarray/pathway_analysis_microarray_02_ora_with_webgestaltr.Rmd#L132 I hate it but don't know what else to do about it.

components/dictionary.txt
jashapiro: If we are good grammarians, it should be *e.g.* everywhere, as it is an abbreviation for *exempli gratia*.

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+HGNC
+histological
+Hochberg
+hypomethylating
+IDH
+Illumina
+IRF
+isoform
+isoforms
+jpeg
```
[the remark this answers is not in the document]
cansavvy: OH Yes, we actually shouldn't be saving JPEGs actually, but this slipped in from an old notebook that hasn't been reformatted, I will turn it to PNG and make `JPEG` a thing.

components/dictionary.txt
jaclyn-taroni: +1 for "e.g."

components/dictionary.txt
cansavvy: We use it as in talking about `DESeq` but I think the instances its used it should just be changed to `DESeq2`.

components/dictionary.txt
```diff
@@ -1,11 +1,133 @@
...
+QuSAGE
+qusage
+QN
+README
+rerio
+ribosomes
+RMA
+RPKMs
 RStudio
+SeT
```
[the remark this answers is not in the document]
cansavvy: See my comment above about: https://github.com/AlexsLemonade/refinebio-examples/blob/e630a4eec112271038f7c1a1a645f533a119fb3b/02-microarray/pathway_analysis_microarray_02_ora_with_webgestaltr.Rmd#L132 I hate it but don't know what else to do about it.

components/dictionary.txt
cansavvy: I think its unlikely for this misspelling to occur naturally so I was okay with this add.

components/dictionary.txt
cansavvy: I think this another instance where it doesn't look for punctuation. But this will be obsolete once the old citations from the old notebooks are updated anyway.

cansavvy: I'll keep `et` and `al` because spelling won't look at the `.` anyway.

components/dictionary.txt
cansavvy: Addressed in https://github.com/AlexsLemonade/refinebio-examples/pull/257/commits/19a0d74f278ccfd083fd8a1a9065634ab0415f66

CONTRIBUTING.md
```diff
@@ -261,9 +267,9 @@ Had no year associated with it, so it has keywords for its tag `pca-visually-exp
...
 
-Spell checks are run automatically using Github actions upon opening a PR for master or prior to merging to master.
-Github actions will abort if there are more than 2 spelling errors and you will need to fix those before continuing.
-You can obtain the list of spelling errors on Github by going to `Actions` and clicking the workflow of PR you are working on.
+Spell checks are run automatically using GitHub actions upon opening a PR for master or prior to merging to master.
+GitHub actions will abort if there are more than 2 spelling errors and you will need to fix those before continuing.
+You can obtain the list of spelling errors on GitHub by going to `Actions` and clicking the workflow of PR you are working on.
 Click on the `style-n-check` step and in the upper right hand corner, there is a button that says "Artifacts" which should list a file called `spell-check-results`.
 Click on `spell-check-results` to download a zip file that contains the list of misspelled words.
 Alternatively, click on the "Check on spell check results" step in the workflow log to see the misspellings.
```
cbethell, line 275: Re the instructions to run the R script and perform spell check locally at line 280 (it won't allow me to leave my comment there in this PR), I had to install the `spelling` package due to the following error:

`Error in loadNamespace(name) : there is no package called ‘spelling’
Calls: :: ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted`

while running in the Docker container.

If the intent is to leave this package out of the Docker container, then we should probably add a note here about likely needing to install the package if you go the local route first. (I could have also done something wrong but I rebuilt the docker image, as well as pulled the latest, and got the error both times).

cbethell (commented):
Looks good! Just one question re running the spell check script locally (it worked fine once I got the package installed, see below for more details!).

CONTRIBUTING.md
cansavvy: This is odd. I've been running this on my Docker container without issue, but I do see there is no explicit install of `spelling`. Let me pull the most recent docker image and start fresh and see whats up. (I usually make it a point to not install things manually to container but always install using the Dockerfile build).

cansavvy: You are correct, it is not on the Docker container. Good catch. I forgot that for github actions we have install specifically. Will add this to Dockerfile now. (Should push automatically I think based on #237 #252)

cbethell (approved):
LGTM! 🚀

PR #259, opened 2020-10-01
Fix automatic rendering

jashapiro:
I apparently combined two options that I can't combine in docker builds. So now we are just pushing to the registry, which means that later steps are downloaded from dockerhub. Which seems redudnant, but I can't figure out the fix right now. 

I also removed style and spellcheck from pushes to master, since those should be completed when Pull Requests are made.


cansavvy (approved):
Give it a try! 🤞

PR #264, opened 2020-10-02
Tidy up autorender change issues

### LINKED ISSUE #256: Changes when autorendering
I just enabled pushing html files back to the repository after rendering in #254 , and I noticed a few things in the first pushed diff: https://github.com/AlexsLemonade/refinebio-examples/commit/e8169fc36de1e23434c19db2599a260baa42a2d7

The first is that, even though it is not October as I write this, it is for github actions robot.  So I am probably the first to see all of the files' date headers flip over from September to October. Because any changes to one page will lead us to rerender all of them, and our current `.Rmd` files in this repo have a dynamically-generated date in the header, this means that anychanges pushed to one file will cause all files to get that file's date in the header.  I would suggest that we go back to manual dates at the top of the page, so we don't see those changes all the time.

It might be worth putting the rendering date in the `Session Info` block to separate the rendering date from the writing date but keep a record of both. This should not affect local development too much, as most html files should not be rerendered often in that context (see #245), but that remains something to keep in mind, as it could result in more inconsequential changes.

The second set of differences are mostly changes in the order of dependencies (versions are unchanged), which mostly seems due to the absence of `rstudioapi_0.11`. I suspect this is the difference between rendering from within `RStudio` and directly from the docker image.

Interestingly, there are also a few of places where there are _slight_ changes in results: 
- In 02-microarray/dimension-reduction_microarray_01_pca.html, `PC285` has somewhat different values. This seems completely inconsequential (Why are we printing all 285 PC values for each sample? that is a LOT for people to scroll past.)
- In 03-rnaseq/differential-expression_rnaseq_01.html, a few values are slightly different. While in the previous example I could see no change in included packages, here there is a slight version change in `matrixStats`: 0.56.0->0.57.0
- A number of figures have changed (I have not looked at them in detail), but they should be compared; differences there are likely to be insignificant.

cansavvy:
## Purpose:
Closes #256 

- I made `dimension-reduction_microarray_01_pca.Rmd` only print out ten PC columns instead all 285. Add some words about that. 

- I search and replaced all `sessionInfo()` with `sessioninfo::session_info()` It honestly looks much neater. 

- I searched and replace all the automatic date setters in the headers with `Oct 2020`. I know this isn't exactly when these were written, but many of the originals were written a couple of years ago and I'm not sure what date to use. Figure moving forward we can do this one by one. 

- The template just has a `{{Month Year}}` for us to fill in.

- Lastly I ran snakemake on the docker container because I think it's not quite automated yet??


jashapiro:
>  Lastly I ran snakemake on the docker container because I think it's not quite automated yet??

Just a quick comment to say that this should still be done by contributors even after automation! It makes review easier, and may catch errors before submission.

02-microarray/clustering_microarray_01_heatmap.Rmd
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Clustering Data - Microarray"
 author: "CCDL for ALSF"
-date: "`r format(Sys.time(), '%B %Y')`"
+date: "Oct 2020"
```
jashapiro, line 4: Style question: Before we were using full Month names. I personally liked it a bit better.

cansavvy:
> Just a quick comment to say that this should still be done by contributors even after automation! It makes review easier, and may catch errors before submission.
> 


Ah good to know. I was pretty sure we will still need to run it but there's been a lot of ideas and changes so I wasn't sure.

jashapiro (approved):
I agree that `sessioninfo::session_info()` looks better, and it will probably be easier to diff changes with one line per package. 

My only comment is the Oct/October question previously. 

Otherwise 🚀

02-microarray/clustering_microarray_01_heatmap.Rmd
cbethell: I agree with @jashapiro here, I personally like the full month names better (and believe I have _mostly_ been using them throughout or at least trying to).

envest: Full month names more accessible?

jashapiro:
I just realized that I restricted the gh-pages branch updates to .html files, but the links to .Rmd files are also served from Pages (the links are to alexslemonade.github.io). So before merging this we should change the add in the workflow. I am submitting this change to this branch now.

cansavvy:
> I just realized that I restricted the gh-pages branch updates to .html files, but the links to .Rmd files are also served from Pages (the links are to alexslemonade.github.io). So before merging this we should change the add in the workflow. I am submitting this change to this branch now.

Oh yeah. I hadn't thought about this. :+1:

PR #265, opened 2020-10-02
Render automatically in gh-pages branch

jashapiro:
This PR changes the github action that occurs on merge to add rendered files to the `gh-pages` branch. 

At the moment, it still uses the github actions users, so the `gh-pages` branch has been left unprotected. If (when) we protect that branch this will need to be updated with the proper token to have write access. In the mean time, this should test the ability of the action to keep things up to date and in sync with the master branch.

### For reviewers
The git steps are now a bit more extensive than just checkout and push, as they require moving changes between branches. They should be looked at with an eye toward unexpected behavior that might occur. Of note is that I used the `recursive` strategy with `theirs` to automatically handle conflicts by using the commit from the `master` (or more accurately the triggering) branch. This seems correct to me, but I might not have thought it all through.

The good thing is that nothing committed by this action will go into `master`, which is much safer than our previous action!


cansavvy (approved):
This seems reasonable to me. Might as well give it a try? 🚀 ¯\\_(ツ)_/¯

PR #266, opened 2020-10-02
Switch to the Docs bot

jashapiro:
This PR adds the token for the alex's lemonade docs bot to the github checkout/push, so it should be able to push to a protected repo.


kurtwheeler (approved):
LGTM!

Issue #245, opened 2020-09-28
Change to the PR process and to CONTRIBUTING.md


cansavvy:
When new examples are added to `_navbar.html` all examples are rerendered and clutter up the PR with the changes. 

From the discussion on #228: https://github.com/AlexsLemonade/refinebio-examples/issues/228#issuecomment-699561760
The plan is to split up example add PRs into 2 PRs (both to master). 

> * PR1: all the changes that are part of the new document and its requirements, with the exception of changes to `_navbar.html` This will prevent most of the extraneous changes in other pages, and simplify review.
> * PR2: Change `_navbar.html` to add links everywhere to the new page.
> 
> This has the advantage of allowing a live look at the rendered files before it is linked into navigation, and no major changes to the development cycle.

This new plan needs to be described in more detail in CONTRIBUTING.md and we can hash out some specifics on this issue.

cansavvy:
Now that the automation pieces are mostly (not quite all) merged, this would be a good thing to test out: see how it works and looks from the author side as well as reviewer side.

jashapiro:
In light of #256 and https://github.com/AlexsLemonade/refinebio-examples/pull/254#issuecomment-701372041, I have some new thoughts and ideas. 

If we remove `_navbar.html` from the `input:` directive for rendering a page, pages will only rerender by default if the underlying `.Rmd` file is changed. This was not the behavior we would want when triggering renders manually, as we wanted changes to the navbar to force changes to all pages, so people would not forget.

However, in #254 we will force all pages to rebuild on merge to master, which will incorporate any changes to `_navbar.html` regardless of whether `_navbar.html` is included as an `input` in each individual step. In other words, if we keep the `_navbar.html` changes in a PR associated with adding a page (and all content for that page), we will be able to preview it in the added page as part of the PR review, but merging that PR will result in rerendering of all pages, propagating that change to all "old" pages.

So what I am saying is that we can leave the instructions mostly as they are, with a single PR per page addition, but change `Snakefile` to accomplish the same overall goal of simplifying review to focus on the new content.

cansavvy:
Chatting with @jashapiro on Slack, we discussed that with these updated changes to how PRs should go, we should also recommend that only the files that you are changing be committed and pushed in your PR. This way the session_info fluctuations we see in the files diff part of PRs are out of the way, and ultimately the automated snakemake step will make sure everything is up to date.

cansavvy:
[...]

Issue #117, opened 2020-07-08
[Discussion/Idea] "Glossary" of recurrent items that are not central to analyses? 


cansavvy:
There's things that pop up time and time again in our notebooks but aren't necessarily central to the analysis: 

Some examples: 
- set.seed()
- the magrittr pipe
- sessionInfo()
- `if (!dir.exists("plots")) {}`

One option is to describe each of these things every time they come up, but what might be a nice alternative is having a link to a "glossary" of sorts where we explain what those items are. That way a user who understands what that's about can focus on the analysis, but a user who's confused can easily find more info and an explanation about what that thing is.

jashapiro:
I like this idea, but I'm going to suggest we think about how we use some of these (probably a separate issue, but here we are). In particular, I would probably vote to switch from `%>%` <- dplyr::`%>%` to just `library(magrittr)`. 

Second, is there a good reason not to skip the `if()` and let the `dir.create()` warning fly if the directory exists? If `dir.create` failed in the presence of a directory that exists I might have a different opinion, but I don't really see the need for the if statement here. I guess the question is really which is more confusing to a user: seeing a warning or the somewhat mysterious `if()` syntax.

(Also, note that this only catches if a _directory_ with the same name exists. If there is a _file_ with the name, the warning might be even more confusing, since you might think you already checked for a conflict.)

Happy to be wrong on either of these. (And definitely keep `set.seed()` and `sessionInfo()`!)

jaclyn-taroni:
Figured if I would offer up `dir.create(..., showWarnings = FALSE)` here.

jashapiro:
> Figured if I would offer up `dir.create(..., showWarnings = FALSE)` here.

I had that in my original comment, but worried about the second case resulting in odd failures... people trying to put a directory where a file was will get a mysterious failure later. Which maybe is an argument for the status quo: while it doesn't error out when trying to put a directory over a file, at least it warns at the relevant action, but (not) making a folder that exists is silent.

cansavvy:
This might be something we can more easily set up after we have more of the material set up. Then we can identify the most common repeat terms; and put links to the glossary at all their occurrences.

cansavvy:
`dev.off()` and RDS files in general are some topics that confuse beginners and should be added to this as well.

PR #268, opened 2020-10-05
Set the seed for those jitter plots in DE modules

cansavvy:
## Purpose:
This _should_ wrap up #256 and it helps us do some testing a la #245 

## Strategy
The DE modules have jitter plots. I should have set the seed so these don't change! (tisk tisk past Candace). 
This PR sets the seed for those three DE modules and I re ran those only. 

- Note that I include a TODO <del>broken link</del> here to a glossary (#117) since `set.seed()` is something that I think would be a good thing for the glossary. I might get a template for the glossary set up soon so we can slowly add to it and this link won't be broken for too long.


jashapiro:
I'd rather not have a broken link if we can avoid it. Maybe an html comment with the todo?

cansavvy:
> I'd rather not have a broken link if we can avoid it. Maybe an html comment with the todo?

We have a broken link on the RNA-seq section header as well. https://github.com/AlexsLemonade/refinebio-examples/blob/28fb568c27ab531af6b77d11c83f42f8ed5e15bd/03-rnaseq/00-intro-to-rnaseq.Rmd#L119

cansavvy:
I can change both/all to html comments though.

envest (commented):
Cool -- set.seed(:100:)

Something I'm curious about in general -- would it make sense to reset the seed every time the code does something involving randomness? For example, assuming the data stays the same, if one were to go back later and add an additional jitter plot somewhere in the code, all the jitter plots that come after that would change, right?

⚠️ opinionated preference ⚠️ when using ```geom_jitter()```, I like to set ```height = 0``` so the numerical biological data on the y axis is not altered. I think that situation may apply here. (I agree there are exceptions but that would be my default.) 👍

jashapiro:
> Something I'm curious about in general -- would it make sense to reset the seed every time the code does something involving randomness? For example, assuming the data stays the same, if one were to go back later and add an additional jitter plot somewhere in the code, all the jitter plots that come after that would change, right?

They would, but _oh the humanity_. Once per notebook solves most of the problem.

envest:
> > Something I'm curious about in general -- would it make sense to reset the seed every time the code does something involving randomness? For example, assuming the data stays the same, if one were to go back later and add an additional jitter plot somewhere in the code, all the jitter plots that come after that would change, right?
> 
> They would, but _oh the humanity_. Once per notebook solves most of the problem.

Agreed! Especially if the presentation of the results is automated and not being post-processed in Illustrator, e.g.

cbethell (commented):
Everything in this PR LGTM and tested fine locally (however, I won't approve just yet, since part of @envest's [comment](https://github.com/AlexsLemonade/refinebio-examples/pull/268#pullrequestreview-502341095) is still outstanding).

cansavvy:
> ⚠️ opinionated preference ⚠️ when using `geom_jitter()`, I like to set `height = 0` so the numerical biological data on the y axis is not altered. I think that situation may apply here. (I agree there are exceptions but that would be my default.) 👍

I went ahead and set height = 0 where we use `geom_jiitter()`. :+1:

03-rnaseq/differential-expression_rnaseq_01.Rmd
```diff
@@ -187,6 +187,14 @@ library(ggplot2)
 library(magrittr)
 ```
 
+The jitter plot we make later on involves some randomness.
```
cbethell, line 190: Actually, it appears that we are not making a jitter plot later on in this notebook.

cansavvy: We do, but it's made using `DESeq2::plotCounts()` function instead of `geom_jitter()` and it doesn't have an equivalent `height` argument (that I can find).

cansavvy: https://github.com/AlexsLemonade/refinebio-examples/blob/28fb568c27ab531af6b77d11c83f42f8ed5e15bd/03-rnaseq/differential-expression_rnaseq_01.Rmd#L381

cansavvy: But the output from this function, is still a jitter plot.

cbethell: Ahh of course, gotcha 👍

jashapiro: It might be worth referring to the actual function that will generate the jitter plot to avoid confusion.

cbethell (approved):
LGTM 👍

PR #272, opened 2020-10-06
Change inputs for Snakefile

cansavvy:
## Purpose:
Now that Snakemake is forced to run everytime before merging to master with GHA (#254), we want to alter when snakemake is run locally. 

Specifically, in the local sense we don't want snakemake to re-run ALL the files  if the _navbar.html, the reference.bib, or .csl file are changed. Those changes will be reflected locally in any Rmd files that are updated but will also be forced to be updated everywhere by GHA befoer merging to master. 

## Strategy
In the Snakefile I delete those three files:  _navbar.html, the reference.bib, or .csl  from the `input` this means snakemake will not evaluate those files when considering if the render script needs to be re-run. It will only look for changes in Rmd files. 

But these files are still provided to the actual script to be run.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -19,14 +19,12 @@ rule target:
 rule render_citations:
     input:
         rmd = "{basedir}/{basename}.Rmd",
-        bib = "references.bib",
-        csl = "components/genetics.csl"
     output:
        "{basedir}/{basename}.html"
     shell:
         "Rscript scripts/render-notebooks.R"
         " --rmd {input.rmd}"
-        " --bib_file {input.bib}"
-        " --cite_style {input.csl}"
+        " --bib_file references.bib"
+        " --cite_style components/genetics.csl"
         " --html {output}"
         " --style"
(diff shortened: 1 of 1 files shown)


cansavvy:
Just so I could prove this to myself, I did a quick edit to the `references.bib` file and ran snakemake, it said there was nothing to do, which is what we expect/want.

Snakefile
```diff
@@ -19,15 +19,12 @@ rule target:
 rule render_citations:
     input:
         rmd = "{basedir}/{basename}.Rmd",
-        nav = "{basedir}/_navbar.html",
-        bib = "references.bib",
-        csl = "components/genetics.csl"
     output:
        "{basedir}/{basename}.html"
```
jashapiro, line 23:
```suggestion
        "{basedir}/{basename}.html"
```

jashapiro (approved):
Looks good. I just found one tiny insignificant indentation issue that we might as well address..

PR #273, opened 2020-10-06
Add checking dockerfile for changes to docker-build step

jashapiro:
Since we are using the `docker-build.yml` step as a check, it would be nice if it ran faster. This PR should accomplish that by making it only run the time consuming docker build steps if the Dockerfile is different from the one in `master`, borrowing ideas from `docker-build-push.yml`

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -17,9 +17,6 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
-        with:
-          # get the full repo
-          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


jashapiro:
Based on limited testing from PRs above, this actually seems to do what we want! The docker build workflow runs if the Dockerfile is different from that in `master`, regardless of whether it changed recently. 

So this should be a pretty good speedup/savings in action run times.

The only concern is that as we now require  `style-n-check` to ensure that PRs are up to date with master, if the bot pushes changes those changes may not retrigger a check. There are a few ways we can get around this, but I think the first step will be to see if it actually comes up at all!

cansavvy:
Seems good to me?!

cansavvy (approved):
🚀

PR #293, opened 2020-10-14
Update dataframe rendering

jashapiro:
This small PR implements a change that @cbethell found to render data frames as fancy paged tables, rather than as plain text. I thought it deserved its own PR.

The only substantive change is in `scripts/render-notebooks.R`, but all html files are updated with the new format (this was mostly for testing, but I left it in.)


cansavvy (approved):
Took a look at the renderings. Seems like a simple change that leads to a nice improvement (the best kind of changes)! Great idea @cbethell ! and thanks for implementing it @jashapiro

Issue #291, opened 2020-10-14
Spell Check isn't failing github action checks when it should


cansavvy:
https://github.com/AlexsLemonade/refinebio-examples/pull/283/checks?check_run_id=1253245020

Not sure why its continuing when it should be failing. I'll look into this today.

cansavvy:
My guess is that it isn't treating it like its numeric again.

cansavvy:
It does print out the correct number of spell check errors but says it like this: `/__w/_temp/1a971902-fb24-4b8a-b30d-67fb42164f56.sh: 2: 7: not found` where `$result` appears to be `7`.  I think that means something in `if (( $result > 2 )); then` needs to be changed?

cansavvy:
I switched the if statement to be `if [ $result \> 2 ];` and this seems to be working as expected. 
https://github.com/AlexsLemonade/refinebio-examples/pull/292#issuecomment-708398830

cansavvy:
This appears to be not entirely resolved now that we are trying it on other branches. This is perplexing. 

Here's a new failed to fail on the ORA branch even with the new if statement mentioned in the above comment. 
https://github.com/AlexsLemonade/refinebio-examples/runs/1254289296?check_suite_focus=true

cansavvy:
Yay. GitHub actions is failing (like its supposed to)! I think what we may have learned here is Github actions likes it best if you use its own GHA `if:`s and it doesn't seem to like listening to bash or sh if statements.

PR #294, opened 2020-10-14
Attempt 2: Spell check still doesn't kill GHA

cansavvy:
## Purpose:
This PR is a second attempt to make GHA exit based on spell check since it stopped doing so and does not seem to be working consistently still. See #291 again


jashapiro (approved):
🚀

PR #298, opened 2020-10-14
Do some rearranging and polishing of the file path set up for getting started for each example

### LINKED ISSUE #269: Ask users to replace paths to data when defining variables
Screen Shot 2020-10-05 at 10 14 30 PM

Can we add `#Replace with path to your metadata file` here? Also define metadata_file name and data file name? 

I'm thinking that if there is only one spot the users will need to update if they are using a different dataset, instead of every spot downstream, less likely they are to make mistakes. Also they will be able to catch if they have pointed it to an invalid path.

We should add a note saying modify the file paths if using a different dataset, and link them back to this chunk, in the downstream chunks.

cansavvy:
## Purpose:
Closes #269 

Based on @dvenprasad  and I's discussion about designating file paths and how to designate them in a way that people don't miss that they've been designated if they use the "Jump to analysis code" but they still are only declared once so that the user only has to change them in one place if they switch datasets. 

https://github.com/AlexsLemonade/refinebio-examples/issues/269#issuecomment-704998909

## Strategy
Step 1) Make the file path changes to the template only according to the discussion on #269. 

Step 2) Ask people to review it to make sure it addresses #269 and that the wording seems good. - WE ARE HERE. 

Step 3) Implement this at each example analysis, then ask for another review. - WE ARE HERE

- Does this seem like a better flow? 
- Should I print out `data_file` and `metadata_file` to make it even more clear or is that overkill because we somewhat assume our users know how file paths work?
- Do the links appear to be in the right places? (I haven't been able to test them yet since the nb.html for the template is a bit wonky because of the `{{}}` bits.


template/template_example.Rmd
```diff
@@ -124,17 +124,30 @@ Your example analysis folder should now look something like this (except with re
...
+metadata_file <- file.path(data_dir, {{METADATA_ACCESSION FILENAME}}) # Replace with file path to your metadata
+```
+
+Now that our file paths are declared, we can use the `file.exists()` function to check that the files are where we specified above. 
+
+```{r}
+# Check if the gene expression matrix file is where the file path stored in `data_file` says
+file.exists(data_file)
+
+# Check if the metadata file is where the file path stored in `metadata_file` says
```
cbethell, line 149:
```suggestion
# Check if the metadata file is at the file path stored in `metadata_file`
```

If you use the above suggestion this line can also change.

cbethell (commented):
Looks good @cansavvy! I just have a few small suggestions below.

cbethell (approved):
LGTM! Seems to satisfy https://github.com/AlexsLemonade/refinebio-examples/issues/269#issuecomment-704998909 👍

dvenprasad (approved):
👍

cansavvy:
Just some information to help review, there are 14 Rmd files total with the changes including the template.Rmd. So there are 6 rna-seq examples and 7 microarray examples and 1 template. 

But note that the ortholog mapping examples (n = 2) do not read in the metadata file so they are slightly different. 

I've been using these numbers to check that the strings have made it in the correct number of times in the project. It looks like the changes are propagated correctly (from my end) and everything runs fine. But another pair of eyes to look through the monotonous changes would be good.

02-microarray/clustering_microarray_01_heatmap.Rmd
```diff
@@ -118,17 +118,30 @@ Your example analysis folder should now look something like this (except with re
 <img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/analysis-folder-structure.png" width=400>
 
 In order for our example here to run without a hitch, we need these files to be in these locations so we've constructed a test to check before we get started with the analysis.
-Run this chunk to double check that your files are in the right place.
+These chunks will declare your file paths and double check that your files are in the right place.
+
+First we will declare our file paths to our data and metadata files, which should be in our data directory.
+This is handy to do because if we want to [switch the dataset](#using-a-different-refinebio-dataset-with-this-analysis) we are using for this analysis, we will only have to change the file path here to get started.
```
cbethell, line 124: Unfortunately, this link to the #using-a-different-refinebio-dataset-with-this-analysis section does not link to the section as ideally as one would hope in the rendered html file, I am thinking this is because the section we are linking to is immediately after.

The link does change to `....html#using-a-different-refinebio-dataset-with-this-analysis` but the user would not be sent to the exact location (this may be a user error on my part so you may want to try to reproduce this issue).

Perhaps we can change this everywhere to:
```suggestion
This is handy to do because if we want to [switch the dataset](#using-a-different-refinebio-dataset-with-this-analysis) we are using for this analysis as outlined in the next section, we will only have to change the file path here to get started.
```

02-microarray/differential-expression_microarray_01_2-groups.Rmd
```diff
@@ -117,19 +117,33 @@ Your example analysis folder should now look something like this (except with re
...
+
+```{r}
+# Check if the gene expression matrix file is at the file path stored in `data_file`
+file.exists(data_file)
+
+# Check if the metadata file is at the file path stored in `metadata_file`
+file.exists(metadata_file)
+```
+
+If the chunk above printed out `FALSE` to either of those tests, you won't be able to run this analysis _as is_ until those files are in the appropriate place.
```
cbethell, line 146:
```suggestion
```

Looks like this is a duplicate line.

cbethell (commented):
Looks good, all of the files seem to have all of the necessary propagated changes. 
I just have one main comment below that you can let me know your thoughts on @cansavvy.

02-microarray/clustering_microarray_01_heatmap.Rmd
cansavvy: This is odd that this isn't working. It shouldn't matter that it's the next section. And I could see that maybe the numbering throws it off, but the numbering didn't throw off the other section link `#check-out-our-file-structure` so that feels inconsistent.

cansavvy: And the only reason I don't want to use numbering in the link is that some sections like ORA and others not yet written in advanced topics will have more rearrangements of this sections so the numbering may not be identical, so I'd hate to have to alter the numbering each time (which we shouldn't have to do??).

I copied and pasted from DocToc output just to make extra sure there's no typos that are breaking it. That's not it.

cansavvy: Nope, even using `#3_Using_a_different_refinebio_dataset_with_this_analysis` doesn't work even though that works normally: https://alexslemonade.github.io/refinebio-examples/02-microarray/clustering_microarray_01_heatmap.html#3_Using_a_different_refinebio_dataset_with_this_analysis

cansavvy: Well, I don't know how else to fix this and it is just thee next section, so I'll resort to sans-link just telling them there will be more info about switching datasets in the next section.

cbethell: Very strange situation, but your final plan to just tell them there will be more info about switching datasets in the next section sounds good to me 👍

PR #300, opened 2020-10-15
Update screenshots in the getting started section for each example

### LINKED ISSUE #199: Create images specifically for "How to Run This Example"
We have images that are included in the template for every analysis.

The image for the download structure appears to be a cropped version of something we include in our docs: 

Where there is extraneous information on the left hand size because that was illustrating a multiple experiment download.

And this is a screenshot on Mac for the analysis folder structure:

Where there is extraneous information at the top of the screenshot. It seems like we would benefit from using a consistent set of images that only illustrate the points we are trying to make and are as OS-agnostic as possible.

### LINKED ISSUE #271: Replace download now modal image
Screen Shot 2020-10-06 at 8 22 57 AM

Can we use one without the error message?

cansavvy:
## Purpose:
Closes #199 and closes #271 

These two issues have to do with updating images in the getting started sections of each analysis. 

## Strategy
I replaced/updated the images in a first commit, obtained the commit id for this, then search and replaced in the project all instances of the old commit id with the new images' commit id. 

I then reran snakemake so I could see how it rendered. 

Does it look like everything's covered?


jaclyn-taroni (approved):
:shipit:

PR #301, opened 2020-10-15
Add 'References' header to each example

### LINKED ISSUE #275: Move References section
The Reference section gets a bit lost below the (very long) Session Info section.

I would propose we move them above it, and give them their own header. Even if we don't move them up, we should give them a header!

We might want to make References and Session Info subheadings of some other heading, but I don't feel strongly about this.

cansavvy:
## Purpose:
Closes #275 by adding `References` headers to each example. 

## Strategy
I added `# References` to the end of each current analysis notebook and then re-ran with snakemake. 

It's interesting that the numbering is left off of the Referrences section? Not sure why this is. Do we not like this? Is this something I should dig into? -- Edit: I looked into this, it seems to be a default for `number_sections: true`. I don't feel like fighting against it particularly.

### DIFF
--- a/02-microarray/clustering_microarray_01_heatmap.Rmd
+++ b/02-microarray/clustering_microarray_01_heatmap.Rmd
@@ -359,3 +359,5 @@ This helps make your code more reproducible by recording what versions of softwa
 # Print session info
 sessioninfo::session_info()
 ```
+
+# References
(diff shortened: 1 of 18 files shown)


cbethell (approved):
LGTM!

Issue #282, opened 2020-10-07
Visual Unity: refinebio-examples font and colors made more like refine.bio


cansavvy:
This is _fairly_ straightforward to do. We already render our notebooks through `render-notebooks.R` and the `rmarkdown::html_document()` can be given a `css` argument. 

I already tested this out and have a basic framework that works set up [on this branch](https://github.com/AlexsLemonade/refinebio-examples/tree/cansavvy/refinebio-fashion). 

Second step is to make that styles.css file conform to the refine.bio style guide here: https://github.com/AlexsLemonade/refinebio-design/blob/master/styleguide-dev.pdf

This is less clear to me what the most straightforward way to do this is.

jashapiro:
Noticed in #246, we should make sure our code chunks don't wrap arbitrarily. While we should strive to keep our lines short, in a narrow window, we may still result in odd line breaks that impair readability of code. 

 To fix this I think we will need to define CSS for the `<code>` blocks with `white-space: pre` rather than `pre-wrap` which seems to be what they get now.

cansavvy:
Closed by #303

cansavvy:
Also superceded by #323

Issue #278, opened 2020-10-07
Visual Unity: Make _navbar.html more match-y with refine.bio


cansavvy:
To make it less looking like users have jumped to a different site when they go to refinebio-examples, we should try to spruce up the navbar to make it match the refine. bio color schemes and maybe have the refine.bio logo. 

Recall that only the [main `_navbar.html`](https://github.com/AlexsLemonade/refinebio-examples/blob/master/components/_navbar.html) needs to be changed and this should propagate to all the examples.

Here are all the refine.bio design items we will need to do this:
https://github.com/AlexsLemonade/refinebio-design

I don't know anything about CSS, but @dvenprasad said we should be able to add a bit to the _navbar.html to make it match-y.

dvenprasad:
Here's the styleguide https://github.com/AlexsLemonade/refinebio-design/blob/master/styleguide-dev.pdf

cansavvy:
We should be able to borrow the things I figured out with David's help on #282 for this too.

cansavvy:
I think this is all set, except for what's now tracked on #323

Issue #277, opened 2020-10-07
refinebio-examples feedback survey 


cansavvy:
As users are going through refinebio-examples we want them to be able to provide feedback. Our main points we want to know from them are: 
- Did they encounter any problems?
- Did they have the resources they need to get started?
- Did they find what the expected to find?

We can have a HubSpot survey linked in the navbar as well as at the end of each example. 

Here's a rough draft for what questions should be included: 

```
- What example analyses were you following today? (Presented as a checkbox list of the examples?)
- How useful did you find this example analysis? (a Likert 5 point scale of usefulness that pops up for each example they said they used)
- Did you encounter any problems? (See if there's a way for them to attach a screenshot). 
- What additional information would you have liked to seen? (Free text)
- How useful did you find the links and sources? (Likert 5 point scale)
- What other analyses would you like to see? (A multiple choice for new materials with Other as an option)
- What were you expecting to see on refinebio-examples that is not there? (free text)
- Provide your email if we can reach out to you for follow up questions in a 15 - 30 minute follow up interview. (email)
```

cansavvy:
Here's a rough draft: https://share.hsforms.com/1pQ8pPB70TuG37sVjr-KtXA336z0

cansavvy:
@dvenprasad I have a comment and two questions: 
Question 1: When people sign up for the emailing list on refine.bio, what HubSpot contact list does that correspond to? I made a guess on this workflow: https://app.hubspot.com/workflows/5187852/platform/flow/40104722/edit
Question 2: What refine.bio logo should be included on the form? (And should we remove one of the two that are on there now?)
Comment is, feel free to take a look at (and make edits to) the form whene you have time (the URL will not change so we can begin adding it to the examples in the meanwhile) https://app.hubspot.com/forms/5187852/editor/a50f293c-1ef4-4ee1-b7ee-c563afe2ad5c/edit/form

cansavvy:
Something we will have to keep in mind and make procedures about is that the `Which examples were you following` question will need to be updated as we update modules (or change the name of them).

cansavvy:
What should this feedback button say? @dvenprasad ?
- `Give us feedback` 
- `Feedback` 
- `Tell us how it's going` 
- `Survey`

I don't love any of these.

dvenprasad:
I went with `Give Feedback`.

cansavvy:
[...]

PR #303, opened 2020-10-15
Mechanics for CSS file and navbar add feedback URL

cansavvy:
## Purpose:
Part of #278 
Part of #282 
Part of #277 

These are updates that need to happen to the navbar and general html/CSS things. It also includes adding the survey URL to the navbar. 

## Strategy
I've added a CSS file that is used by `render-notebooks.R`, but it's not exactly perfected. I'm not exactly sure on how to fix it to make it more prettier and to be more visually unified with refine.bio

- In general, the CSS file here needs to be perfected in a way that makes our html renders look more like they are actually related to refine.bio, but I don't know how to do that. We can revisit the CSS file itself later for more polishing. 

- When you look at the drop down menu the text is hard to read, I don't know how to make it so the dropdown menus have a good contrast with the background when you are hovering over it. I tried `link.hover` it did not work. 

Here's a preview of what this looks like: (It's not exactly right yet). 
[getting-started.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/5392896/getting-started.html.zip)


scripts/render-notebooks.R
```diff
@@ -113,6 +113,7 @@ rmarkdown::render(tmp_file,
     toc = TRUE, toc_depth = 2,
     toc_float = TRUE, number_sections = TRUE,
     df_print = "paged"
+    css = file.path("..", "components", "styles.css")
```
cansavvy, line 116: I don't like this precarious way I've designated this file path but `html_document()` wants it relative to the Rmd and I'm not sure how to make this less precarious.

cansavvy:
I've changed the colors so they are readable and still probably better than all white background with black text. It's not perfect, but I think we can move forward with this and return to the CSS file for polishing later.

scripts/render-notebooks.R
jashapiro: I would assume a full path will always work? If so: 
```suggestion
    css = normalizePath(file.path("components", "styles.css"))
```
As you do this, it may make sense to move `references.bib` into `components` as well.

jashapiro:
Since this just popped up in Slack, I'm going to suggest that this either go back to draft status or be changed to a blank CSS file (which I think will just use the defaults) until we have some time/expertise to devote to making it look decent.

cansavvy:
> Since this just popped up in Slack, I'm going to suggest that this either go back to draft status or be changed to a blank CSS file (which I think will just use the defaults) until we have some time/expertise to devote to making it look decent.

I would agree with you, but what is not recorded here and there was no way for you to know (my bad) is that @dvenprasad has taken a whack at this and is about to commit some things soon.

dvenprasad:
Okay, made a few changes. It should feel closer to refine.bio now. 
The only thing I wasn't able to do was get the logo to sit properly in the navbar. It cut out. And when I tried to make the navbar accommodate it, it just overlapped with the content, instead of pushing it down.  So for now I've replaced the logo with text until someone else can take a pass at it. 
 
Screen Shot 2020-10-22 at 9 25 05 AM

cansavvy:
I think this looks good to me! I rendered it locally and it works and looks good. I filed an issue for the logo problem: https://github.com/AlexsLemonade/refinebio-examples/issues/323 

Otherwise I think this is ready for merging if everyone else agrees.

jashapiro:
> Okay, made a few changes. It should feel closer to refine.bio now.
> The only thing I wasn't able to do was get the logo to sit properly in the navbar. It cut out. And when I tried to make the navbar accommodate it, it just overlapped with the content, instead of pushing it down. So for now I've replaced the logo with text until someone else can take a pass at it.

Thanks Deepa!

I think we can fix some of this by adjusting offset margins; there is code in the generated page that does this:

https://github.com/AlexsLemonade/refinebio-examples/blob/a660e89593690cac0113fb4f1cb8709c159ef624/02-microarray/00-intro-to-microarray.html#L2715-L2740

Oddly, this appears after the inserted CSS; we may need to put this code in the `_navbar.html` fragment for it to render properly.

There is a second issue with when the window gets small, but not too small: 
Screen Shot 2020-10-22 at 10 01 28 AM

I don't know if this is related to the right side menu, or if it needs to be fixed some other way.

You also commented out the button in the navbar which shows up when the window gets down to mobile size; that was where the menus went but now they are gone 😱.

Another issue that I had been avoiding, but I think it is time: We are currently generating 100% self-contained files, but this seems unneccessary, and may slow downloads if people are on navigating among pages: I think we can use `lib_dir` and `self_contained` options to slim things down/reduce redundancy in generated output:  https://www.rdocumentation.org/packages/rmarkdown/versions/2.4/topics/html_document

jashapiro:
> I think this looks good to me! I rendered it locally and it works and looks good. I filed an issue for the logo problem: #323
> 
> Otherwise I think this is ready for merging if everyone else agrees.

With the #323 filed (and filing another to explore non-standalone pages), I think this can go in, but I think we should add back the menu button before merging. Screenshots below:

With menu button:
Screen Shot 2020-10-22 at 10 29 31 AM

Without menu button:
Screen Shot 2020-10-22 at 10 29 49 AM

dvenprasad:
Added back the mobile nav

jashapiro (approved):
Looks pretty good. There are still a few style glitches we may need to sort out, but this is good for now!

PR #307, opened 2020-10-16
Adding basic footer

### LINKED ISSUE #279: Visual Unity: Make refine.bio-examples footer
I don't know how difficult this is to do, but if we can add a footer that is like refine.bio's footer, that would help people cite us as well as help them navigate to other things: https://www.refine.bio/

This is definitely doable but how painful it is to make it exactly like refine.bio's footer I do not know. 

https://stackoverflow.com/questions/25329375/creating-a-footer-for-every-page-using-r-markdown

This _shouldn't_(?) have to change for every example like the `_navbar.html` does so it shouldn't add to upkeep burden. I think we should be able to add a footer to the header of each example and then the rest should be automatic. I haven't tested this though.

cansavvy:
## Purpose:
Closes #279 

Add a basic footer with additional info to the end of each htmls. 

## Strategy
I've added a `footer.html` file that is included in the render by the `render-notebooks.R` script (which is used by snakemake. 
It is successful in that when you re-render it is included in the render and the links work. Otherwise its very basic but we can take another whack at the details later: Filed here: #328 

Here's a file with the footer incorporated as it is right now.
[getting-started.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/5425215/getting-started.html.zip)


cansavvy:
This is what this looks like right now: 

Screen Shot 2020-10-22 at 11 10 39 AM

cansavvy:
This navbar could use more pizzazz and information. But I think for now just having the mechanics should be fine. I think I will file an issue about [my comment above](https://github.com/AlexsLemonade/refinebio-examples/pull/307#issuecomment-714562143) so we can return to this when someone who knows more about CSS would like to take a stab at it.

jaclyn-taroni (approved):
I agree with some of the comments on #279 - namely, I don't know if we need to link to refine.bio in the footer (the header's visible the whole time). 

My _strong suspicion_ is that @dvenprasad wanted to see CCDL and ALSF branding in the footer, but there's no details on that issue about what _from refine.bio's footer_ needs to be there.

But also this isn't final so, let's go 🚀

Issue #276, opened 2020-10-06
Add URLs for all references


jashapiro:
Some of the references in the `.bib` file do not have URLs associated with them. That adds some unneccessary friction to people trying to track down a reference. All of our references should have URLs, preferably DOIs, and we should add that requirement to `Contributing.md`

PR #308, opened 2020-10-19
Standardize references.bib and add URLs

cansavvy:
## Purpose:
#276 In this issue, Josh pointed out not all references have URLs printed out and this would help users out for finding sources. 

## Strategy
Step 1) I added URLs (and doi's where applicable) to all source that didn't have these yet. 
Step 2) I standardized the syntax by switching `" "` to `{ }` and etc. 
Step 3) Changed the CSL file so that URLs are always printed out no matter the source type (e.g. website, Article, Manual, etc.)
Step 4) Updated CONTRIBUTING.md to reflect these changes

I have NOT re-rendered all notebooks as to not clutter this up for reviewers. 

Here's one preview of what the references look like now: 
[getting-started.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/5402090/getting-started.html.zip)

- How do we feel about the `url: ` prefix in the render References section, should I get rid of it or change it?
- Did I miss any stray syntax that is out of line? 
- Any URLs still missing? 
- Do the CONTRIBUTING.md updates make sense?


components/genetics.csl
```diff
@@ -151,7 +151,7 @@
           </group>
         </else>
       </choose>
-      <text macro="access" prefix=". "/>
+      <text variable="url" display="block" prefix=" url: "/>
```
jashapiro, line 154: I think we want to leave this as is, and modify the `access` macro. This will allow us to display the DOI if it exists.

```suggestion
      <text macro="access" prefix=". "/>
```

I think the following update to the access macro should do the trick (I used https://editor.citationstyles.org/visualEditor/ to work this out)
```<macro name="access">
    <choose>
      <if match="any" variable="DOI">
        <text variable="DOI" prefix="https://doi.org/"/>
      </if>
      <else-if match="any" variable="URL">
        <text variable="URL"/>
      </else-if>
    </choose>
  </macro>
```

cansavvy: I may have put this in the wrong place? It's not working

cansavvy: I'm not familiar with how this stuff works, I just made changes until it did.

cansavvy: Would you mind committing this change where it's supposed to go? I'm a bit confused on where it should go.

jashapiro: Done. (It replaced the current access macro)

cansavvy: Thanks!

cansavvy: It works!

CONTRIBUTING.md
```diff
@@ -231,6 +231,19 @@ From the RMarkdown Cookbook [bibliographies chapter](https://bookdown.org/yihui/
...
+Automatic reference renders (like those recommended below) don't maintain consistent syntax (though they usually will still work).
+You may/will still need to do some manual edits to maintain consistency with the current `references.bib`-listed sources.
+
+**Adding sources rules:**  
+
+- All sources need to have URLs.
+- Where applicable, sources should have doi's.
+- `{}` are preferred to `" "`.
+- All lower case tags are preferred to upper case (eg. `title` NOT `Title`).
+- Spaces before and after `=` are preferred.
```
jashapiro, line 243: I don't really think we need these rules, as they seem overly prescriptive to me and don't affect rendering.

More helpful to suggest may be making sure the capitalization of tool and gene names will be preserved by judicious use of `{}`. 

I noted using this that `texmed` does bad stuff with titles. I just looked for alternatives, and https://www.doi2bib.org/ seems much better (and more general)

CONTRIBUTING.md
```diff
@@ -231,17 +231,30 @@ From the RMarkdown Cookbook [bibliographies chapter](https://bookdown.org/yihui/
...
 - _R packages_: Use this kind of command in the console: `toBibtex(citation("tidyverse"))`.
 Copy and paste the output to the `references.bib` file.
 
-- _Articles_: For PubMed articles, you can use [this website to find the article](https://www.bioinformatics.org/texmed/). Select the correct article that pops up and click `Export`. Then copy and paste the reference to the `references.bib` file.
+- _Articles_: For articles, you can use [this website to get a BibTex](https://www.doi2bib.org/ ).
+Then copy and paste the reference to the `references.bib` file.
 
 - _Websites_: Websites can be added using this template:
 ```
 @online{website,
```
jashapiro, line 254: I think this should be: 

```suggestion
@website{citekey,
```

Also, in sections below, we should use `@article` lower case and `@website` for consistency.

CONTRIBUTING.md
```diff
@@ -231,17 +231,30 @@ From the RMarkdown Cookbook [bibliographies chapter](https://bookdown.org/yihui/
...
 
+Automatic reference renders (like those recommended below) don't maintain consistent syntax (though they usually will still work).
+You may/will still need to do some manual edits to maintain consistency with the current `references.bib`-listed sources.
+
+**Adding sources rules:**  
+
+- All sources need to have URLs.
+- Where applicable, sources should have doi's.
+- `{}` are preferred to `" "`.
+- To preserve capitalization, use an extra `{}` around what should be capitalized.
```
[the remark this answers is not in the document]
jashapiro: Other things worth adding:
- All journal names should be given in full, not abbreviated
- Author names with abbreviations should include a period after the abbreviation. i.e. `Harry S. Truman` not `Harry S Truman ` and `Truman, H. S.` not `Truman, HS`. I tried to catch most of these, but maybe didn't get them all.

jashapiro (commented):
I did some more cleaning of the .bib file using https://marketplace.visualstudio.com/items?itemName=James-Yu.latex-workshop

This allowed me to catch a few more inconsistencies, (and some duplicates) which I committed straight to the .bib file. It also did some nice lining up of the fields, which is nice, but not strictly necessary.

I ended up sorting by the citekey, which I don't love, but given the fact that most author names are given as `First Last` rather than `Last, First` it seemed more consistent.

I added a few other notes about some of the nitpicky things I found.

CONTRIBUTING.md
cansavvy: Ah. I didn't know it could be lower case but that makes sense.

CONTRIBUTING.md
```diff
@@ -231,17 +231,32 @@ From the RMarkdown Cookbook [bibliographies chapter](https://bookdown.org/yihui/
...
+You may/will still need to do some manual edits to maintain consistency with the current `references.bib`-listed sources.
+
+**Adding sources rules:**  
+
+- All sources need to have URLs.
+- Where applicable, sources should have doi's.
+- `{}` are preferred to `" "`.
+- To preserve capitalization, use an extra `{}` around words that require special capitalization (genes, software packages, etc).
+- All journal names should be given in full, not abbreviated
+- Author names with abbreviations should include a period after the abbreviation. i.e. Harry S. Truman not Harry S Truman and Truman, H. S. not Truman, HS.
```
jashapiro, line 244:
```suggestion
- Author names with abbreviations should include a period after the abbreviation. i.e. `Harry S. Truman` not `Harry S Truman` and `Truman, H. S.` not `Truman, HS`.
```

PR #309, opened 2020-10-19
Update Formatting of Titles

jashapiro:
I am here adding a number of changes mostly to titles, where I am cleaning up formatting. 

In particular, there are a number of places where the titles had extra brackets, which will interfere with automatic capitalization.

In the reverse case, many package names that require particular capitalization did not have the appropriate brackets, so I added them.

I am not sure I got everything, and I encountered a few duplicates along the way. I might recommend coming back to this with a dedicated tool (i.e. [BibDesk](https://bibdesk.sourceforge.io/)) to help clean things up.

### DIFF
--- a/references.bib
+++ b/references.bib
@@ -1,6 +1,6 @@
 
 @Manual{Ahlmann-Eltze2020,
-    title = {ggupset: Combination Matrix Axis for 'ggplot2' to Create 'UpSet' Plots},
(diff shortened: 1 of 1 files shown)


cansavvy (approved):
This looks good to me! Looks like this also found and removed a couple duplicates that were probably introduced from merge conflicts.

PR #310, opened 2020-10-19
Add a syntax highlight thing

### LINKED ISSUE #287: Add syntax highlighting to code
Our currently rendered html files do not have syntax highlighting for R code, which makes me sad.

Can we implement something like https://highlightjs.org to automatically highlight our code?

https://highlightjs.org/usage/

cansavvy:
## Purpose:
Closes #287 highlighting code was requested, there's an argument that can do that for us. There's a lot of default themes available, but `haddock` seemed more "highlighty' than the default but not too bold. 

## Strategy
I added `highlight = "haddock"` to the `html_document()` function in `render-notebooks.R` script. 

Here's what that looks like: 
[differential-expression_microarray_01_2-groups.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/5404214/differential-expression_microarray_01_2-groups.html.zip)

### DIFF
--- a/scripts/render-notebooks.R
+++ b/scripts/render-notebooks.R
@@ -112,7 +112,8 @@ rmarkdown::render(tmp_file,
   output_format = rmarkdown::html_document(
     toc = TRUE, toc_depth = 2,
     toc_float = TRUE, number_sections = TRUE,
-    df_print = "paged"
+    df_print = "paged",
+    highlight = "haddock"
   ),
   # Save to original html output file name
   output_file = output_file


jashapiro (approved):
Looks good, nice find! I might like `pygments` better than `haddock`, but do I really care?

cansavvy:
> Looks good, nice find! I might like `pygments` better than `haddock`, but do I really care?
> 

I'm not that attached to haddock. If you like pygments, we can use that.

cansavvy:
> > Looks good, nice find! I might like `pygments` better than `haddock`, but do I really care?
> 
> I'm not that attached to haddock. If you like pygments, we can use that.

I tried pygment, but I like haddock better. So I guess I am attached.

Issue #297, opened 2020-10-14
Move to a staging then make it live strategy


cansavvy:
Currently, this repo's github pages are not "live" in that they are not linked to refine.bio and are not directly user-facing. 

However, once we do make this repo's material live #286  we will need to figure out our strategy for editing our material, and it would probably be advisable to have a checkpoint before a PR something becomes user-facing

This issue originated here: https://github.com/AlexsLemonade/refinebio-examples/issues/295#issuecomment-708617271

We will need to figure out the details of what this strategy would look like. It would probably involve a `staging` branch and altering the github actions in a way that we can still see the html previews but they won't be published directly upon merging PRs.

cansavvy:
This is also probably related to #285 and may be something that we need to discuss in tandem.

cansavvy:
I think this will involve a few steps:

1) Make a staging branch and set it so PRs are required. (and whatever other protections we currently have on the master branch should be added here too). 
2) Can we set the default for PRs to go to the `staging` branches (is this an option in settings?). 
3) Update github actions to treat `staging` like the `master` was being treated.

4) Determine processes around deployments:  (aka `staging` -> `master` PRs)
     - How often do we deploy changes?
     - Should we require two approvals for these PRs?

5) CONTRIBUTING.md needs to be updated to reflect this process and any other changes we'll need to make.

cansavvy:
I don't think we want each deployment to have its own version tag number, but we should decide how often we consider something to be a "new release" : https://github.com/AlexsLemonade/refinebio-examples/issues/285

jaclyn-taroni:
> Can we set the default for PRs to go to the `staging` branches (is this an option in settings?).

Yes: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch

cansavvy:
As far as this step: 

> 3\. Update github actions to treat `staging` like the `master` was being treated.

I think in most instances we just change the `on` criteria too be `staging` instead of master, but a question is do we want the re-rendering to happen again before `staging -> master` merges or only `some-branch -> staging` merges? I think the answer is run all the checks again before master because why not?

jaclyn-taroni:
> How often do we deploy changes?

This is not like a service where the API might go down and need manual intervention to go back up, so having a set schedule seems a little bit like overkill to me. Which is to say, my answer is as often as we need to.

> Should we require two approvals for these PRs?

[...]

PR #311, opened 2020-10-20
Update github actions to reflect staging branch

cansavvy:
## Purpose:
Now that the staging branch is our default, we need github actions to be updated and act accordingly. 

Goal 1: have gha run on staging PRs as it was running for master PRs
Goal 2: make one exception: staging PRs should NOT publish to `gh-pages` which is where the public facing content goes "live".

See this comment: https://github.com/AlexsLemonade/refinebio-examples/issues/297#issuecomment-713002280

## Strategy

1) I added `staging` as an `on:` criteria so that all checks are ran with PRs to `staging` branch. 

2) I tried splitting the commit to gh-pages step into two different `if` `runs` depending on whether it is the `staging` or `master` branch. 

- Does this look like it should work? I'm particularly unsure about the `if` statements but that's what we'll test here. 
- Is there any facets I missed regarding `staging` that should be addressed?

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -30,6 +30,7 @@ jobs:
           then
             pages_branch="gh-pages"
           elif [github.ref == 'refs/heads/staging']
+          then
             pages_branch="gh-pages-stages"
           fi
           echo $pages_branch
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -26,11 +26,11 @@ jobs:
 
       - name: Checkout pages branch and sync with changes
         run: |
-          echo ${GITHUB_REF#refs/*/}
-          if [${GITHUB_REF#refs/*/} == 'refs/heads/master']
+          echo $GITHUB_REF
+          if [ $GITHUB_REF == 'refs/heads/master' ]
           then
             pages_branch="gh-pages"
-          elif [${GITHUB_REF#refs/*/} == 'refs/heads/staging']
+          elif [ $GITHUB_REF == 'refs/heads/staging' ]
           then
             pages_branch="gh-pages-stages"
           fi
@@ -89,7 +89,7 @@ jobs:
 
       # If we are on the staging branch, do not publish to github pages
       - name: Commit changed html back to non-public pages
-        if: ${GITHUB_REF#refs/*/} == 'refs/heads/staging'
+        if: github.ref == 'refs/heads/staging'
         run: |
           git add -A
           git commit -m 'Render html, do not publish' || echo "No changes to commit"
@@ -97,7 +97,7 @@ jobs:
 
       # If we are on the master branch, publish to github pages!
       - name: Commit changed html to public pages
-        if: ${GITHUB_REF#refs/*/} == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         run: |
           git add -A
           git commit -m 'Render html and publish' || echo "No changes to commit"
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -5,6 +5,8 @@ name: Build, Render, and Push
 on:
   push:
     branches: [ staging, master ]
+  pull_request:
+    branches: [ staging, master ]
 
 jobs:
   # This workflow contains a single job called "build-all"
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -25,12 +25,14 @@ jobs:
           token: ${{ secrets.DOCS_BOT_GITHUB_TOKEN }}
 
       - name: Checkout pages branch and sync with changes
+        env:
+          merge_branch: github.ref
         run: |
-          echo $github.ref
-          if [$github.ref == 'refs/heads/master']
+          echo $merge_branch
+          if [$merge_branch == 'refs/heads/master']
           then
             pages_branch="gh-pages"
-          elif [$github.ref == 'refs/heads/staging']
+          elif [$merge_branch == 'refs/heads/staging']
           then
             pages_branch="gh-pages-stages"
           fi
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -24,23 +24,15 @@ jobs:
           # use alexslemonade-docs-bot
           token: ${{ secrets.DOCS_BOT_GITHUB_TOKEN }}
 
-      - name: github.ref
-        run: echo $github.ref
-        
-      - name: If master branch, use public pages branch
-        if: github.ref == 'refs/heads/master'
-        env:
-          pages_branch: gh-pages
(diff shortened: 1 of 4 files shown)


cansavvy:
To test this, I think we'll need to push a very minor html change and see if it ends up in `gh-pages` when we merge to `staging`. It should not.

.github/workflows/docker-build-push.yml
```diff
@@ -68,19 +68,26 @@ jobs:
...
+      - name: Commit changed html back to staging branch
+        if: github.ref == 'refs/heads/staging'
         run: |
           git add -A
-          git commit -m 'Render html' || echo "No changes to commit"
-          git push origin gh-pages || echo "No changes to push"
-
-
+          git commit -m 'Render html, do not publish' || echo "No changes to commit"
+          git push origin staging || echo "No changes to push"
```
jashapiro, line 85: If the goal here is to commit back to staging, we will need to change the checkout before committing. The way this is currently set, it will make the changes in `gh-pages`, but push `staging`, which will be unchanged. 

I have not yet looked at the settings for `staging`, but if we are to follow the settings for `master`, then that branch should be protected as well. In that case the bot will not be able to push to it because pushes can only come from approved PRs. The easiest way around this would be to create another branch, `gh-pages-stages` if you will, that we could then push to from the staging branch.

cansavvy: I've set `staging` to have the same settings that `master` does. 

`gh-pages-stages` feels a tad janky, but I guess its fine?

jashapiro: it was a kind of a joke for a rhyme, but if you start the branch name with `gh-pages`, it will use the same settings. I don't know that it needs to be protected at all though. Just don't include `staging` in the name or it will inherit the restrictions from `staging`

docker/Dockerfile
```diff
@@ -14,7 +14,8 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     zlib1g \
     libbz2-dev \
     liblzma-dev \
-    libreadline-dev
+    libreadline-dev \
+    libglpk40
```
cansavvy, line 18: This is to address the individual Dockerfile problem noted on #312

jashapiro: Cam we put this in a separate PR? I think it deserves its own testing.

.github/workflows/docker-build-push.yml
cansavvy: I didn't put protections on `gh-pages-stages` btw.

.github/workflows/docker-build-push.yml
```diff
@@ -68,19 +68,26 @@ jobs:
...
+          git commit -m 'Render html, do not publish' || echo "No changes to commit"
+          git push origin gh-pages-stages || echo "No changes to push"
 
+      # If we are on the master branch, publish to github pages!
+      - name: Commit changed html to public facing gh-pages
+        if: github.ref == 'refs/heads/master'
+        run: |
+          git add -A
+          git commit -m 'Render html and publish' || echo "No changes to commit"
+          git push origin gh-pages || echo "No changes to push"
```
jashapiro, line 93: This is only half the battle... We also need to change line 28 to checkout the correct branch, depending on `github.ref`. This probably means passing `github.ref` in as an ENV variable in the `Checkout pages branch and sync with changes` step, then using that in an if statement for which branch to checkout.

You could use the same logic to collapse the last two steps here, but I don't think that is necessary.

.github/workflows/docker-build-push.yml
jashapiro: It will inherit them because it has `gh-pages` in the name. This should not matter though given the current setup.

cansavvy: OH right.

.github/workflows/docker-build-push.yml
```diff
@@ -26,11 +26,11 @@ jobs:
...
-          if [${BRANCH_NAME} == 'refs/heads/master']
+          echo ${GITHUB_REF#refs/*/}
+          if [${GITHUB_REF#refs/*/} == 'refs/heads/master']
           then
             pages_branch="gh-pages"
-          elif [${BRANCH_NAME} == 'refs/heads/staging']
+          elif [${GITHUB_REF#refs/*/} == 'refs/heads/staging']
           then
             pages_branch="gh-pages-stages"
           fi
```
jashapiro: Can't make this a suggestion, but try going back to: 
```
 run: |
          echo $GITHUB_REF
          if [ $GITHUB_REF == 'refs/heads/master' ]
          then
            pages_branch="gh-pages"
          elif [ $GITHUB_REF == 'refs/heads/staging' ]
          then
            pages_branch="gh-pages-stages"
          fi
```
And change line back 92 to 
```
if: github.ref == 'refs/heads/staging'
```
and similarly line 100
```
if: github.ref == 'refs/heads/master'
```

.github/workflows/docker-build-push.yml
```diff
@@ -4,7 +4,7 @@ name: Build, Render, and Push
 # events only for the master branch
 on:
   push:
-    branches: [ master ]
+    branches: [ staging, master ]
```
jashapiro, line 7: Just for testing, do you want to try 
```suggestion
    branches: [ staging, master, cansavvy/gha-updates-pr-process ]
```
for a round or two?

cansavvy:
Looks like all is working now that #316 is merged. Should I revert the [testing part](https://github.com/AlexsLemonade/refinebio-examples/pull/311#discussion_r509389048)?

jashapiro:
> Looks like all is working now that #316 is merged. Should I revert the [testing part](https://github.com/AlexsLemonade/refinebio-examples/pull/311#discussion_r509389048)?

Yes... There is a minor question about the fact that this branch means that neither of the `if` statements are true, but we will see if it works when we merge. It ought to.

cansavvy:
> Yes... There is a minor question about the fact that this branch means that neither of the `if` statements are true, but we will see if it works when we merge. It ought to.

Agreed, but I'm not sure of a great remedy for this. Does this mean you think this is ready for merging then?

.github/workflows/docker-build-push.yml
```diff
@@ -4,7 +4,7 @@ name: Build, Render, and Push
 # events only for the master branch
 on:
   push:
-    branches: [ master ]
+    branches: [ staging, master, cansavvy/gha-updates-pr-process ]
```
jashapiro:
```suggestion
    branches: [ staging, master ]
```

jashapiro (approved):
Should be good to go! (there may be a suggestion in there that is now irrelevant, but I can't find it to delete it)

.github/workflows/docker-build-push.yml
cansavvy: I did this in 7baef02

jashapiro: As I said, irrelevant, but I couldn't delete it!

cansavvy: Oh weird.

Issue #281, opened 2020-10-07
Set up Google analytics for refinebio-examples pages


cansavvy:
I don't know how to do this, but its something @dvenprasad recommended. We want to know what people are clicking on and using the most. Feedback is helpful!

This seems like it should help: 
https://support.google.com/analytics/answer/1008015?hl=en

jashapiro:
We will probably want to make use of the include options as I described [here](https://github.com/AlexsLemonade/refinebio-examples/issues/279#issuecomment-706194235) to get this into the header.

cansavvy:
Okay. I tried to do this and it doesn't seem to be working. There's so many buttons and I'm quite lost: https://analytics.google.com/analytics/web/?authuser=1#/p249911149/realtime/overview?params=_u..nav%3Ddefault

jashapiro:
What is that link supposed to be taking us to? What did you try? What are your questions?

cansavvy:
> What is that link supposed to be taking us to? What did you try? What are your questions?

I had so many questions but it turned out I was doing it all wrong anyway, @dvenprasad says we set this up through ALSF's already standing Google Analytics account which I didn't realize was a thing.

cansavvy:
But a question that is still pertinent is will https://github.com/AlexsLemonade/refinebio/issues/2542 affect the domain name we provide for this?

jashapiro:
> But a question that is still pertinent is will [AlexsLemonade/refinebio#2542](https://github.com/AlexsLemonade/refinebio/issues/2542) affect the domain name we provide for tracking?

I would go with the assumption that we will be able to change it later if we need to.

cansavvy:
Okay. I've coordinated with Craig and we need to add this a Google Analytics script with tags and etc at the top of all our htmls. I'll file a PR with this.

cansavvy:
Wanted to check with Craig before I posted the blurb publicly. This is what needs to be posted at the top of each html. 
```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-C59ZQK86CL"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'G-C59ZQK86CL');
</script>
```

jashapiro:
[...]

PR #314, opened 2020-10-21
Add google analytics to renderings

cansavvy:
## Purpose:
I think this is the last bit to wrap up #281 on our end. 

We needed a chunk of code I got from Craig to be added to all our html headers. 
I used the `rmarkdown::includes()` function and it works quite nicely.


scripts/render-notebooks.R
```diff
@@ -107,13 +107,17 @@ new_lines <- append(lines, header_line, header_range[1])
 # Write to an tmp file
 readr::write_lines(new_lines, tmp_file)
 
+# Declare path to google analytics bit
+google_analytics_file <- normalizePath(file.path("components", "google-analytics.txt"))
```
jashapiro, line 111: It may not look like it, but this is an .html file, and should get that extension.

```suggestion
google_analytics_file <- normalizePath(file.path("components", "google-analytics.html"))
```

jashapiro (approved):
Looks good (assuming you got the 👍🏼 from Craig on the fragment). I'd just change the file name as indicated. Approving on the assumption you can do that easily 😉.

Issue #312, opened 2020-10-20
Have github actions quit completely if the docker build steps (or render steps) fail


cansavvy:
https://github.com/AlexsLemonade/refinebio-examples/runs/1265424176?check_suite_focus=true

I'm not sure why this is. I can push the image manually, but its odd the Docker image with `clusterProfiler` didn't make it to Dockerhub. 

It's been failing periodically because of this.

cansavvy:
So the build failed because of a dependency but it didn't actually stop the build it just kept moving on anyway. I didn't expect this. 

```
ERROR: dependency ‘enrichplot’ is not available for package ‘clusterProfiler’
```

cansavvy:
I don't like that the renders have been failing this whole time but it didn't seem to tell us. This seems like not what we want for our processes. Can we make this rendering completion a necessary check? Looks like currently it is not required (not sure why it's not).

cansavvy:
To be clear, I don't *think* this is a fail to push problem, I think this was a weird instance that something failed to install but the docker kept building anyway, which is why we didn't see an error until render. The error I posted above was buried in a lot of docker output text but if you look at that instance, it didn't fail the docker build with that error even though the packages didn't install.

jaclyn-taroni:
Change the title to reflect what needs to be solved now that you have an idea.

jashapiro:
I guess I assumed that getting an exit code 1 along the way would fail the a github actions step, but clearly not. We should be able to make that happen

jashapiro:
There seem to be 2 issues here: Dockerfile fix and GHA fix.

cansavvy:
> There seem to be 2 issues here: Dockerfile fix and GHA fix.

Dockerfile fix is easy, [this works](https://github.com/AlexsLemonade/refinebio-examples/issues/312#issuecomment-713025216). GHA fix is really what we are left with.

jaclyn-taroni:
> I guess I assumed that getting an exit code 1 along the way would fail the a github actions step, but clearly not. We should be able to make that happen

There is an exit code 1 for the render step here: https://github.com/AlexsLemonade/refinebio-examples/runs/1265424176?check_suite_focus=true

So I take it that getting an exit code 1 _does_ fail the GHA step.

> > There seem to be 2 issues here: Dockerfile fix and GHA fix.
> 
[...]

PR #316, opened 2020-10-21
Docker dep fix: Add lib package 40 thing that clusterprofiler needs

cansavvy:
## Purpose:
Fix the precipitating event of #312 - clusterprofiler needed a dependency to install even though it docker didn't stop building when it failed to install. 

This was the problem: 
```
#12 608.2 Error in dyn.load(file, DLLpath = DLLpath, ...) : 
#12 608.2   unable to load shared object '/usr/local/lib/R/site-library/igraph/libs/igraph.so':
#12 608.2   libglpk.so.40: cannot open shared object file: No such file or directory
#12 608.2 Calls: <Anonymous> ... asNamespace -> loadNamespace -> library.dynam -> dyn.load
#12 608.2 Execution halted
#12 608.2 ERROR: lazy loading failed for package ‘enrichplot’
#12 608.2 * removing ‘/usr/local/lib/R/site-library/enrichplot’

ERROR: dependency ‘enrichplot’ is not available for package ‘clusterProfiler’
```

## Strategy
Add back in `libglpk40` which appears to be the thing it needs.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,7 +14,8 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     zlib1g \
     libbz2-dev \
     liblzma-dev \
-    libreadline-dev
+    libreadline-dev \
+    libglpk40
 
 # libmagick++-dev is needed for coloblindr to install
 RUN apt-get -y --no-install-recommends install \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,8 +14,7 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     zlib1g \
     libbz2-dev \
     liblzma-dev \
-    libreadline-dev \
-    libglpk40
+    libreadline-dev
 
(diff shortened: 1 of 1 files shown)


jashapiro (commented):
Do we want to add `options(warn=2)` in this PR as suggested:  

https://github.com/AlexsLemonade/refinebio-examples/issues/312#issuecomment-713586320

cansavvy:
> Do we want to add `options(warn=2)` in this PR as suggested:
> 
> [#312 (comment)](https://github.com/AlexsLemonade/refinebio-examples/issues/312#issuecomment-713586320)
 
Seems like a good time for that, sure! Will add now.

jashapiro (approved):
Confirming that this docker image does build as expected.

cansavvy:
Yay. `option(warn=2)` made it fail like it should in https://github.com/AlexsLemonade/refinebio-examples/runs/1288528019?check_suite_focus=true (I deleted the linux dependency that gave the warnings in the first place). 

Now I've reverted that test in 46835b5 so it should pass again.

PR #319, opened 2020-10-21
Add some issue templates

### LINKED ISSUE #317: Make an issue template for these refinebio-example issues
Probably should look something like this.

```
What was the original problem and context/Background / State the problem
What would force you to write what you assume down 
What needs to be there for you to complete this task (in however long an interval after the issue is written)
What potential "gotchas" do we know of?
What are the recommended next steps?
```

### LINKED ISSUE #318: Add a new analysis-specific issue template
We could use a template for new analyses issues. 

Here's some questions it should address.
```
What should this new example show; what's the goal of the analysis? 
What kind of refine.bio dataset will we need? 
What steps should be included in this analysis? An outline would be nice if possible. 
What packages/methods do you recommend using or looking into for this analysis?
Do you have helpful references and resources we should consult for creating this example? 
```

cansavvy:
## Purpose:
Closes #317 
Closes #318 

The issues in this repo might be better if we had templates to help us out. I've added some rough drafts of what these issue templates might look like.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-analysis.md
+++ b/.github/ISSUE_TEMPLATE/new-analysis.md
@@ -7,9 +7,11 @@ assignees: ''
 
 ---
 
-### What are goals of this new example analysis?
+### What are the goals of this new example analysis?
 
 <!-- What should this new example show that's not currently shown in the existing samples -->
+<!-- What was the background and context that lead to this new example? -->
+<!-- Links to github comments or related issues are also helpful --> 
 
 ### What kind of dataset will this need?
 
--- a/.github/ISSUE_TEMPLATE/new-analysis.md
+++ b/.github/ISSUE_TEMPLATE/new-analysis.md
@@ -1,0 +1,24 @@
+---
+name: New analysis example
+about: Use this issue template for filing a new analysis issue
+title: 'New Analysis Example:'
+labels: new analysis example
+assignees: ''
+
+---
+
+### What are goals of this new example analysis?
+
+<!-- What should this new example show that's not currently shown in the existing samples -->
+
+### What kind of dataset will this need?
+
+<!-- Is there a particular species, sample size, experimental model, etc. that's needed -->
+
+### What steps should be included in this analysis?
+
+<!-- An outline of what steps should probably be taken would be nice if possible -->
+
+### What packages/methods do you recommend using or looking into for this analysis?
+
+<!-- Do you have helpful references and resources we should consult for creating this example? -->
--- a/.github/ISSUE_TEMPLATE/other-issue.md
+++ b/.github/ISSUE_TEMPLATE/other-issue.md
@@ -1,8 +1,6 @@
 ---
 name: Other issue
 about: Use this issue template to describe a current issue with an analysis or documentation (that is not a new example analysis)
-title: 'Other issue'
-labels: other issue
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/other-issue.md
+++ b/.github/ISSUE_TEMPLATE/other-issue.md
@@ -1,0 +1,27 @@
+---
+name: Other issue
+about: Use this issue template to describe a current issue (that is not a new example analysis)
+title: 'Other issue:'
+labels: other issue
+assignees: ''
+
+---
+
+### Background
+
+<!-- What was the background and context that lead to this problem? -->
+<!-- Links to github comments or related issues are also helpful --> 
+
+### Problem
+
+<!-- Why does the current state of things not work?-->
+
+#### What potential "gotchas" do we know of?
+
+<!-- Are there any potentially tricky things or unknowns that the person who takes on this issue should know? -->
+
+### What are the recommended next steps?
+
+<!-- If applicable, what steps would you need to take to get started? -->
+
+<!-- What other info is needed to complete this task (in however long an interval after the issue is written) -->
--- a/.github/ISSUE_TEMPLATE/other-issue.md
+++ b/.github/ISSUE_TEMPLATE/other-issue.md
@@ -1,7 +1,7 @@
 ---
 name: Other issue
-about: Use this issue template to describe a current issue (that is not a new example analysis)
-title: 'Other issue:'
+about: Use this issue template to describe a current issue with an analysis or documentation (that is not a new example analysis)
+title: 'Other issue'
 labels: other issue
 assignees: ''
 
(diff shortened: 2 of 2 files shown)


cansavvy:
@cbethell would you like to take a look at these issue templates and see what you think should bee added or clarified? 

If you imagine yourself reading an issue to implement it, what info would you need to know and do you think it would be covered in the prompts here? 

On the flip side, if you were to file an issue about something, would this have enough guidance to make you feel confident in filling it out?

.github/ISSUE_TEMPLATE/new-analysis.md
```diff
@@ -0,0 +1,24 @@
+---
+name: New analysis example
+about: Use this issue template for filing a new analysis issue
+title: 'New Analysis Example:'
+labels: new analysis example
+assignees: ''
+
+---
+
+### What are goals of this new example analysis?
```
cbethell, line 10:
```suggestion
### What are the goals of this new example analysis?
```

.github/ISSUE_TEMPLATE/other-issue.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Other issue
+about: Use this issue template to describe a current issue (that is not a new example analysis)
```
cbethell, line 3:
```suggestion
about: Use this issue template to describe a current issue with an analysis or documentation (that is not a new example analysis)
```

We may want to add this here because `issue` alone may be a bit misleading in that one may think this template is to describe an existing issue/ticket rather than an issue with an existing analysis. What do you think?

.github/ISSUE_TEMPLATE/other-issue.md
```diff
@@ -0,0 +1,27 @@
...
+title: 'Other issue:'
```
cbethell, line 4: If we make the change above, perhaps we should also propagate the change here?
```suggestion
title: 'Other issue re existing analysis:'
```

Or something like that

.github/ISSUE_TEMPLATE/other-issue.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### What potential "gotchas" do we know of?
+
+<!-- Are there any potentially tricky things or unknowns that the person who takes on this issue should know? -->
+
+### What are the recommended next steps?
+
+<!-- If applicable, what steps would you need to take to get started? -->
+
+<!-- What other info is needed to complete this task (in however long an interval after the issue is written) -->
```
cbethell, line 27:
```suggestion
<!-- What other info is needed to complete this task (in however long an interval after the issue is written) -->
<!-- Do you have helpful references and resources we should consult? -->
```

I believe this would be helpful information under this section as well so maybe include it as one of the prompts?

.github/ISSUE_TEMPLATE/other-issue.md
cbethell: Note that I added "or documentation" here because I realized just "Use this issue template to describe a current issue with an analysis" may be too specific. However, this is not completely necessary? What do you think?

.github/ISSUE_TEMPLATE/new-analysis.md
```diff
@@ -0,0 +1,24 @@
...
+
+<!-- What should this new example show that's not currently shown in the existing samples -->
```
cbethell, line 12: The idea for some new analyses may result from comments on PRs or issues related to existing analyses. Should we also had a section or prompts here on the background for a new analysis issue? Maybe:

```suggestion
<!-- What should this new example show that's not currently shown in the existing samples -->
<!-- What was the background and context that lead to this new example? -->
<!-- Links to github comments or related issues are also helpful --> 
```

Or perhaps we can add a `### Background` header like in the other template in this PR? Not sure which option would be best here as I believe both would get the job done.

cbethell (commented):
I believe the headings (questions) on both templates in this PR do a good job of covering what we need to be documented in an issue in this repo!

I just left a few comments on some wording and suggested prompts we can add in, let me know what you think.

.github/ISSUE_TEMPLATE/other-issue.md
cansavvy: I agree. Will use your wording.

.github/ISSUE_TEMPLATE/other-issue.md
cansavvy: I think the description should be enough.

jaclyn-taroni: Do you think we need the `Other issue:`? It seems like these will not necessarily have a unifying theme and appending `Other issue:` to the start will make automagically finding them to mention them on GH or connect things via ZenHub more difficult.

.github/ISSUE_TEMPLATE/other-issue.md
```diff
@@ -0,0 +1,28 @@
+---
+name: Other issue
+about: Use this issue template to describe a current issue with an analysis or documentation (that is not a new example analysis)
+title: 'Other issue'
+labels: other issue
```
jaclyn-taroni, line 5: Similar question about the title - how does automatically labeling these with `other issue` help us?

cansavvy: Probably doesn't. I looked at the openpbta issue templates and this type of things was there, but we can delete as many of these categories as is useful.

.github/ISSUE_TEMPLATE/other-issue.md
cansavvy: Ah I see. I'll delete this.

.github/ISSUE_TEMPLATE/other-issue.md
jaclyn-taroni: That's a very different beast! These are supposed to help us on a project that will very much be internal. So in addition to what will result in good issues that have the required context and are actionable, we need to think about how we're going to _find stuff_ and get a high-level look at the state of the project.

cansavvy: Yes I understand better what the intent is for these now. 👍

cansavvy:
@cbethell Thanks for the review, I think I've incorporated your suggestions. See what you think. Note that if we try out the templates we merge here and after we use them we decide we don't like them or they need to be tweaked, we can always do that.

cbethell (approved):
LGTM @cansavvy 🚀

PR #313, opened 2020-10-20
Updating CONTRIBUTING.md with instructions about staging -> master set up

cansavvy:
## Purpose:
We need some rules/guidance written down in CONTRIBUTING.md about how the `staging` and `master` branches are to be used. 

This should be the last bit to close #297? Let me know if I'm wrong. Of course I can also see us revisiting this again as we continue to try to put it into practice (This has already happened as I tried out previous strategies on #320 and #329).

## Strategy
Tried to describe the branches and what kinds of PRs happen here so that way contributors know how this is planned to work. These instructions have fluctuated a bit since this PR originally opened and we've been testing out these strategies on #320 and #329. See those PRs for some of the discussion that led to this configuration. 

The headlines of this process (once we confirm it) will be reflected in a checklist form when we make a "make it live" PR template #322. 

#### Specific questions: 

- <del>I have a bigger maybe somewhat controversial question about how "never really user-facing" changes should be merged. Should these still go through the staging -> master process? The only reason I wonder if maybe they should go direct to master is that it may be confusing on where a contributor should look for the most recent instructions (like this CONTRIBUTING.md). But I suppose `staging` is now default so this should be where the most up-to-date CONTRIBUTING.md (or other development pertinent, but not user facing items) should be? I would hate for someone to accidentally be using an out of date `CONTRIBUTING.md` or other development pertinent item. </del> I think I've answered my own question, because staging is default, I think developers should generally be looking at staging, which will always have the latest for these items. 

- Should I be more prescriptive/make a standard for a prefix or branch name that "make it live" PRs should have? I think something that makes it clear what is happening is good, but I also don't want to be over-prescriptive.
 
- [Should we give some guidance on how large/small "make it live" PRs should ideally be](https://github.com/AlexsLemonade/refinebio-examples/pull/313#issuecomment-715367215).

- Not sure about [how hotfixes should ultimately make it into staging (as well as master)](https://github.com/AlexsLemonade/refinebio-examples/pull/313#discussion_r510914009).

#### General questions: 
- Does this cover everything? 
- Are we concerned about any part of this process?

PR #320, opened 2020-10-21
Staging -> Master

cansavvy:
## Purpose:
This should bring master up to date with the most recent PRs: 

#314 and #311 

It's also a good test for how the gha's mechanics are working for this kind of merge.


cansavvy:
I expected this would be checking out `gh-pages` but it is still checking out `gh-pages-stages` Upon merging will this be the correct branch though?? @jashapiro 
`https://github.com/AlexsLemonade/refinebio-examples/pull/320/checks?check_run_id=1288999328

cansavvy:
I think maybe it should be `github.base_ref` since here both the branch and the base branch are something that initiates github actions from the `on:` criteria?

jashapiro:
> I expected this would be checking out `gh-pages` but it is still checking out `gh-pages-stages` Upon merging will this be the correct branch though?? @jashapiro
> `https://github.com/AlexsLemonade/refinebio-examples/pull/320/checks?check_run_id=1288999328

I think this is correct. This was triggered by the push to `staging` not the PR. (This also means that `github.base_ref` will not be defined)

Most of the time this should not come up, as we should not generally be committing directly to `staging` as it appears you did here to merge from `master`?

cansavvy:
> > I expected this would be checking out `gh-pages` but it is still checking out `gh-pages-stages` Upon merging will this be the correct branch though?? @jashapiro
> 
> > `https://github.com/AlexsLemonade/refinebio-examples/pull/320/checks?check_run_id=1288999328
> 
> 
> 
> I think this is correct. This was triggered by the push to `staging` not the PR. (This also means that `github.base_ref` will not be defined)

Ah right. 


> Most of the time this should not come up, as we should not generally be committing directly to `staging` as it appears you did here to merge from `master`?


We would be committing directly to staging and then to master? This is more representative of what the process will look like.

jashapiro:
> We would be committing directly to staging and then to master?

We would be committing to `staging` via a PR, then merging to `master` from there. Or more likely, we would be creating a different new branch which consists of particular files/commits that have been tested in staging (to avoid irrelevant changes that might have gone into staging for another purpose)?. I think this process still needs discussion in #313.

cansavvy:
> I think this process still needs discussion in #313.

Besides the discussions of other instances of this kind of PR (that are happening elsewhere (#313), is there anything to this particular PR that needs adjusting before we merge?

jashapiro:
I am not going to approve this PR; I think we need to follow a staging->branch->master strategy. Otherwise this PR keeps getting new commits as more changes get pushed in, so an approval might not reflect the current contents.

Right now there is at least one conflict, and resolving it does not make sense within the branch (and should actually not be allowed).

cansavvy:
> I am not going to approve this PR; I think we need to follow a staging->branch->master strategy. Otherwise this PR keeps getting new commits as more changes get pushed in, so an approval might not reflect the current contents.
> 
> Right now there is at least one conflict, and resolving it does not make sense within the branch (and should actually not be allowed).

Okay. Yes. I’ve been noticing these conflicts. I think what you say makes sense. Will readjust this and clarify on #313

cansavvy:
Will reconfigure this PR. As is, the conflict resolutions make this not an ideal situation (can't be committing directly to staging!). So will try a "cherry pick" and make a new branch strategy.

PR #321, opened 2020-10-21
Stop staging from pushing to dockerhub

jashapiro:
## Purpose:
Addresses some considerations related to #297  arising from unexpected consequences of #311 
Namely, that we don't really want to overwrite the docker image on Dockerhub that is tagged as `latest` to be something from staging.

## Strategy

I abandoned my fancy "did it change" code, so we build the docker image every time. This is mostly because I was worried about a change somehow hiding and resulting in not building the required image. (The case I can envision is two merges in rapid succession, the first has changes in the Dockerfile, the second doesn't (relative to the first), so it skips the build step and then ends up pulling the "old" image from Dockerhub)

The costs of using the wrong build seem higher than the cost of building more often than is strictly necessary.

Following that, I change the build step from "Build and Push" to "Build and Load".

Then in a separate step we push to Dockerhub if we are in the `master` branch. 

We could push a `staging` tagged docker image to Dockerhub, but I have elected not to do that; if there is a reason we might want to, I would consider it.

Note, in the current setup, the push step will be skipped until this is merged to master, but I have [tested it](https://github.com/AlexsLemonade/refinebio-examples/runs/1289291981?check_suite_focus=true) with a different tag.

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -58,14 +58,14 @@ jobs:
           tags: ccdl/refinebio-examples:latest
       # for staging, we just want to build, but not push
       - name: Build and Load Docker image
-        if: github.ref == 'refs/heads/staging'
+        if: github.ref == 'refs/heads/jashapiro/dont-push-staging-docker'
         uses: docker/build-push-action@v2
         with:
           push: false
           load: true
           context: docker
           file: docker/Dockerfile
-          tags: ccdl/refinebio-examples:latest
+          tags: ccdl/refinebio-examples:staging
       # download data
       - name: Download data
         run: bash scripts/download-data.sh
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -65,7 +65,7 @@ jobs:
           load: true
           context: docker
           file: docker/Dockerfile
-          tags: ccdl/refinebio-examples:staging
+          tags: ccdl/refinebio-examples:latest
       # download data
       - name: Download data
         run: bash scripts/download-data.sh
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -37,40 +37,32 @@ jobs:
           git checkout $pages_branch
           git merge -s recursive --strategy-option=theirs ${{ github.event.after }}
 
-      # Test if Dockerfile has changed
-      # sets steps.check_docker.outputs.changed to 1 if the Dockerfile has changed, 0 otherwise
-      - name: Check Dockerfile for changes
-        id: check_docker
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-push.yml
```diff
@@ -37,47 +37,32 @@ jobs:
...
-          push: true
+          push: false
+          load: true
           context: docker
           file: docker/Dockerfile
           tags: ccdl/refinebio-examples:latest
+      # push the Docker image if this is master
+      - name: Push Docker image
+        if: github.ref == 'refs/heads/master'
+        run: docker push ccdl/refinebio-examples:latest
```
cansavvy, line 61: Because users (at this point) are never instructed to use the docker image, and it's mainly for development purposes, isn't it more important that the docker image is up to date at the "staging" stage?

cansavvy (commented):
One question. But otherwise this makes sense.

.github/workflows/docker-build-push.yml
jashapiro: I'm not sure what you mean here... This process will also apply to changes to the Docker image, so we need to use the Dockerfile that is part of staging, which may be ahead of the "released" version. I would expect the image on Dockerhub to correspond to the current public pages. If we want to have a `staging` tagged file as well, that would be easy to implement. 

But I think this gets a bit into the process for development: I can imagine where two people working independently would end up on different image versions, regardless of what is listed as 'latest'. We've certainly seen this happen, and I think restricting the `latest` version to be parallel to `master` is the most straightforward to explain.

jaclyn-taroni: What is someone returning to the project to develop, while others have been actively developing (e.g., people are still merging to staging), supposed to pull from Dockerhub? I think that is the nature of @cansavvy's question.

cansavvy: > I think restricting the latest version to be parallel to master is the most straightforward to explain.

Why is this more straightforward than having the docker image updated for `staging` and `master`? 

For development purposes it makes sense to me to have the docker image updates available to pull as soon as they are approved and being used for rendering. 

But we can also make a `staging` tag in preparation for a future time in which we may actually do instruct users to use the docker image.

jashapiro: Okay, I think that makes sense. In that case, maybe we want to switch things around: keep `latest` as the tag that corresponds to the staging branch, and make a `release` tag or similar that corresponds to `master`.

cansavvy: That seems fine to me.

cansavvy (approved):
This looks good to me.

jashapiro:
Oops... merged without editing the title. Oh well. This PR does not do what it says it does.

PR #324, opened 2020-10-22
Bring refine.bio fashion to staging too. 

cansavvy:
### Purpose: 

#303 had predated the existence of the `staging` branch and it was merged to `master` instead. 

In reality, we probably should have changed the target base to `staging` so we could practice this 2 PRs to "live" model more, but its been merged and here we are. 

Now we need to make sure staging has that information from #303. So this is a staging -> master merge which I don't think should be a typical part of our process.


cansavvy:
Me thinking about this, led to my comment here about how hotfixes should work: https://github.com/AlexsLemonade/refinebio-examples/issues/297#issuecomment-714569999

cansavvy:
I didn't want to commit conflict resolves directly to master, so I'm trying a different strategy #325. 

Closing this now.

PR #326, opened 2020-10-22
Update diagrams showing how microarray/RNA-seq work 

cbethell:
## Purpose:
What issue(s) does your PR address?

This PR closes #255.

## Strategy
What was your strategy for this new or edited analysis?

Per the [original comment on #255](https://github.com/AlexsLemonade/refinebio-examples/issues/255#issue-712223762), I recreated the diagrams showing how microarray/RNA-seq work based on the figures currently in master:

[microarray diagram](https://github.com/AlexsLemonade/refinebio-examples/raw/13fd8c2d8469f2ac35e1608b4a952a5f5fe9698c/components/figures/microarray-overview.jpg)
[RNA-seq diagram](https://github.com/AlexsLemonade/refinebio-examples/raw/13fd8c2d8469f2ac35e1608b4a952a5f5fe9698c/components/figures/rnaseq-overview.png)

The intention here was to create higher resolution diagrams, and a more accurate single-color microarray diagram that corresponds with the data in refine.bio.

## Concerns/Questions for reviewers:
What things should reviewers look out for?

- Do the images in this PR satisfy the suggestions made on issue #255? cc: @cansavvy and @dvenprasad
- I converted the `microarray-overview.jpg` file to a PNG file (for consistency with the other images in the directory). I can change this file back to JPG if necessary.
- I added a sentence below each diagram as a brief description of what's going on in them. Please let me know if these descriptions are not clear or need more context.

### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (not applicable)

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [x] Comments and documentation are up to date.
* [x] All links have been checked and are properly formatted.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar).
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds.

PR #329, opened 2020-10-23
Making staging changes live

cansavvy:
## Purpose
We've merged a bunch of changes to staging now. We should try to get these into master. 

This is the predecessor to #320 which we realized wasn't ideal because of conflict resolutions and that whenever more changes are happening to `staging` the PR was continually being changed. 

I don't believe there's any changes that shouldn't be made live, so this stage-merge branch contains all the changes in staging. 

I described my thoughts on this PR process in #313 and this is where I'm testing this out. 

_To be more specific this includes_: 
- CSS file and refine.bio fashion to htmls #303 
- Basic footer add #307 
- Google analytics adds #314 
- Issue templates #319  
- Diagram updates #326 
- Github actions updates #311 
- References.bib moved


jashapiro (approved):
Looks like it does what it says... 

But as I am sure you know it is hard to tell! This should get easier as we implement squash merges to master.

Issue #336, opened 2020-10-27
The multiple pull request templates do not show up 


cansavvy:
### Background

We added the PR templates in #334 but now they aren't really showing up, this is not a shock, but I still want to make sure its track. 

### Problem

I tried following https://github.blog/2018-01-25-multiple-issue-and-pull-request-templates/  and not shockingly, it did not work as seemed to indicate it should (they made it sound like it was so simple but I don't think that it is THAT simple). 

### What are the recommended next steps?
I'm going to do more Dr. Google asking and digging into github docs, looking for examples of where someone might have gotten multiple PR templates to work.

cansavvy:
Oh I see what that blog meant about the URLs, the templates can populate your PR if you add an ending with `expand=1&template=<TEMPLATE_FILENAME>` : https://github.com/AlexsLemonade/refinebio-examples/compare/cansavvy/survey?expand=1&template=hotfix-pr.md

cansavvy:
When you open a PR, it is populated with the links to the templates and if you click on Preview you can click on the link directly and it will bring you to the template! :tada: I think this might be as good as we are going to get.

jashapiro:
> When you open a PR, it is populated with the links to the templates and if you click on Preview you can click on the link directly and it will bring you to the template! :tada: I think this might be as good as we are going to get. 

Cool that this works! Thinking about the common workflows though... we might want to make the "standard" PR the default template, but include links to the hotfix and publish templates as part of the intro?

Saves some clicks (going to preview mode, especially) in the most common case.

cansavvy:
> we might want to make the "standard" PR the default template, but include links to the hotfix and publish templates as part of the intro?

I think having the general `Other` PR template as the default makes sense and having links to the others makes sense.

PR #337, opened 2020-10-27
Try a "main PR" strategy with links to the real PR templates

cansavvy:
## Background
An attempt for closing #336 that I *think* should work but I'm not sure. 

Apparently Github doesn't have the same kind of interface for choosing between PR templates as it does for issue templates: https://github.community/t/multiple-pull-request-templates/1850/8 (GitLab does apparently but that's not helpful for us). 

Here's an article (like many others) that explains that Github doesn't have an interface for this but GitLab does: https://seesparkbox.com/foundry/better_pull_requests_merge_requests_with_templates

## About the problem: 

You can use multiple PR templates but it involves piecing together the URL with a `expand=1&template=<TEMPLATE_FILENAME>` type of deal, but this is inconvenient and annoying which is not what templates are supposed to be about. 

Try out this link to see what I mean: https://github.com/AlexsLemonade/refinebio-examples/compare/?expand=1&template=hotfix-pr.md
The templates are here, but there's no easy way to get to them. 

## Potential solution here: 
But, my idea is if Github takes a "main PR template" still, then I can just have links to the others and we can have a not exactly pretty, but somewhat functional way to choose a template. 

I think if we use `<a href="?expand=1&template=hotfix-pr.md"> Hotfix PR </a>` it *should* append to what ever your current URL is? Which maybe is what we want????

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,13 +2,13 @@
 ### Use a link to choose a PR template meant for the type of changes you are proposing:
 
 For any stage of a adding a new analysis example:
-[New analysis PR](https://github.com/AlexsLemonade/refinebio-examples/compare/survey?expand=1&template=new-analysis-pr.md).
+New analysis PR - `?expand=1&template=new-analysis-pr.md`
 
 For publishing changes to Github pages:
-[Publishing PR](https://github.com/AlexsLemonade/refinebio-examples/compare/survey?expand=1&template=publish-pr.md).
+Publishing PR - `?expand=1&template=publish-pr.md`
 
 For either stage of a hotfix PR -- something that is straightforward and needs to be user-facing quickly:
-[Hotfix PR](https://github.com/AlexsLemonade/refinebio-examples/compare/survey?expand=1&template=hotfix-pr.md).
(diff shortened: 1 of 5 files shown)


cansavvy:
So I tried this: `<a href="?expand=1&template=publish-pr.md"> Publishing PR </a>` type strategy if I just typed it in, used preview to click it. It flashes the contents of `publish-pr.md` up on the PR -- it knows what I mean but then it disappears. Bah humbug.

jaclyn-taroni (approved):
Let's just try it! Other things that came up in 1:1 

- One template to rule them all where each of the four are HTML commented out blocks
- One template to rule them all where certain questions are _condensed_ and then certain prompts are removed for certain types of PRs

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+
+### Use a link to choose a PR template meant for the type of changes you are proposing:
+
+For any stage of a adding a new analysis example:
```
cbethell, line 4:
```suggestion
For any stage of adding a new analysis example:
```

cbethell (approved):
I agree with the comment above, let's try it out!

Issue #340, opened 2020-10-28
[Discussion] Organizational structure of pathway analysis material


cansavvy:
### Background

I'm writing down some thoughts from a discussion with @jaclyn-taroni:

Currently, we have plans for making a Pathway Analysis Introduction notebook (#214) which will be the first kind of module-specific introduction notebook. This strategy was discussed (mostly by me it looks like) in https://github.com/AlexsLemonade/refinebio-examples/issues/141 where the idea of what to do with README info was brought up. For most modules it wasn't a concern, but for pathway analysis it is. 

For pathway analysis the intro notebook is meant to give some background information that will help inform the user about what kind of pathway analysis they may want to use (based on their dataset). And each pathway analysis example links back to that introduction notebook

### Problem

As of now, both RNA-seq and microarray sections would have their own pathway analyses and introduction notebook. It's unlikely the Introduction notebook between the technologies would differ much in their content, meaning they would probably be pretty redundant. 

No matter what, there will probably be some redundancy involved as has been discussed elsewhere : #223 and #175 

What we need to find is what is the best balance between user navigability and ease of maintenance. In other words, redundant material is harder to maintain but it's easier for a user to find what they are looking for if it is structured more like the other modules.

Three options we can think of as of now:
Option 1) Continue with plan as is, keep intro notebooks noting that microarray and rna-seq pathway intro notebooks will be almost identical. 
     - Pros: the structure will be consistent with being "by technology" which may be nice if users largely focus on one technology (perhaps more likely one dataset that happens to use a particular technology) over the other. 
     - Cons: this will mean two introduction to pathway analyses notebooks exist that are largely identical. We deal with this elsewhere in this repo, but do we think its worth it to deal with redundancy again here? 

Option 2) Abbreviate the introduction notebook material into a section that can be in each pathway analysis example and continue with the microarray and rna-seq examples. 
     - Pros: this continues with the philosophy of making each example "self-contained" and doesn't disturb the "by technology" organizational structure too much. 
     - Cons: this is still being rather redundant and will require abbreviating the pathway analyses intro material some. These may not be insurmountable barriers, but they still involve the redudancy issue (which appears to be common to this repo). 

3) Make a whole new "bigger" cross-tech section of pathway-analysis aka 02-microarray, 03-rna-seq, 04-pathway-analysis where an introduction could be more similar to the rna-seq and microarray introductions and would require less redundancy -- but how would this be user-wise?
[...]

PR #342, opened 2020-10-29
Add timeframe reminder to issue template

cansavvy:
### Purpose

Communicating timeframes (when relevant) is helpful and important. I added a reminder bit in our issue template so we can better remember to communicate timeframes when they apply. 

### Issue addressed

Related to the kind of comment here: https://github.com/AlexsLemonade/refinebio-examples/issues/340#issuecomment-718815055

If we don't find this reminder is helpful,  then we can remove it later!

### DIFF
--- a/.github/ISSUE_TEMPLATE/other-issue.md
+++ b/.github/ISSUE_TEMPLATE/other-issue.md
@@ -8,13 +8,13 @@ assignees: ''
 ### Background
 
 <!-- What was the background and context that lead to this problem? -->
-<!-- Links to github comments or related issues are also helpful --> 
+<!-- Links to github comments or related issues are also helpful -->
 
 ### Problem
 
 <!-- Why does the current state of things not work?-->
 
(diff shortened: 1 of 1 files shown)


cbethell (approved):
Looks and seems good! 

I do have one question and that is, is there a reason we're not updating the new analysis issue template as well? I can imagine a timeframe may be a bit more difficult to estimate in that case but can also be helpful! Otherwise this LGTM!

cansavvy:
>a timeframe may be a bit more difficult to estimate in that case but can also be helpful! 

Like you kind of said here, I think timeframe estimates are less likely to be possible/accurate for new analyses example adds and I think the decision making happens slightly differently. Decisions in regards to analysis examples seem to happen on more of a per comment,  in a review type basis (rather than an issue basis) -- like what happening on draft PRs like in #339 so it seems less helpful to put it on the issue. 

But, of course where and if it applies, communicating timeframes for some of those decisions is also still good!

Issue #284, opened 2020-10-09
Microarray: Pathway Analysis - GSEA


cansavvy:
As a part of the pathway analysis module, we should add a GSEA example for microarray data. 

There's a lot of docs from the broad we can reference: https://www.gsea-msigdb.org/gsea/index.jsp

We can use the differential expression results from either `differential-expression_microarray_01_2-groups.Rmd` or `differential-expression_microarray_02_several-groups.Rmd`.

We can borrow some inspiration from https://github.com/AlexsLemonade/training-modules/blob/master/pathway-analysis/02-gene_set_enrichment_analysis.Rmd

cbethell:
Per discussion with @cansavvy, the current plan for this issue is as follows:

1. Load in packages (`clusterProfiler` to run GSEA, `msigdbr` and species annotation package)
2. Import the results from this repo's `differential-expression_microarray_01_several-groups.Rmd` (note that this is tentative to change based on how the visualization of the GSEA results for this dataset looks later on in the example)
3. Like in the updated ORA example, we can include the [“getting familiar with `clusterProfiler`'s options"](https://github.com/AlexsLemonade/refinebio-examples/blob/36b81e881afd50f2760122495de142039233183f/02-microarray/pathway-analysis_microarray_02_ora.Rmd#L184) section
4. Like in the training modules GSEA example, include a [section on isolation of hallmark genes](https://github.com/AlexsLemonade/training-modules/blob/3dbc6f3f53c680ec6aa2f513851c1cd4635cc31c/pathway-analysis/02-gene_set_enrichment_analysis.Rmd#L90) with a note saying something along the lines of: "if you want a more curated dataset, run this chunk" (whether or not we keep this chunk may depend on processing time and length of the notebook; we may replace the chunk with a sentence or two re using hallmark genes to generate a curated dataset)
5. Perform gene identifier conversion (using the `mapIds()` function)
6. Determine genes of interest list (using log fold change and possibly adjusted p-value as cutoffs -- decision will likely be made based on number of significant results found and visualizations)
7. Perform gsea using `GSEA()` function from `clusterProfiler` providing the genes of interest list to the `geneList` argument -- note that discretion may also be used here with the number of permutations based on processing time (we want to keep processing time to a minimum for our users but without losing an important aspect of the analysis)
[...]

PR #345, opened 2020-10-30
Pr 1 of 2: Add Microarray Pathway Analysis - GSEA example

cbethell:
## Analysis Purpose

This PR addresses issue #284

## Pull Request Stage

The draft PR relevant to this PR is #339.

## Strategy

As suggested in [one of the comments on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/339#discussion_r513798463), the decision was made to break up the analysis notebook into two PRs.

For this PR, the first of the two, focus was placed on the preparation steps leading to the gene set enrichment analysis step (the second PR will be focused on the GSEA step and visualizations).

Therefore, this PR includes:
- Context around the differential expression results file being imported
- A section on `Getting Familiar with `clusterProfiler's` options 
- Gene identifier conversion from Ensembl to Entrez IDs
- A step to filter out duplicate IDs to account for warning messages thrown by `clusterProfiler::GSEA()` (which will be used in the latter half of the notebook) -- brought up by [this comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/339#discussion_r513799154)

The context re what GSEA does at the top of the notebook, as well as the context in the `Getting familiar with clusterProfiler's options` section can take an extra close look.

The step filtering out duplicates can also take an extra close look as I am not sure that the decision made was best suited for this purpose (the reason I filtered out duplicated IDs is because we have filtered out multi-mapped IDs in a previously updated analysis and mentioned that we may want to do that depending on our downstream case, and in this case, our downstream analysis does not like duplicated IDs -- of which there were n = 2 duplicate IDs in this PR).

Overall, does the guided instructions and context provided in this PR appear to be suffice for our users? Should there be more/less context added?

### Content checks
* [x] Seed is set (not applicable)

### Formatting Checks

### Add datasets to S3


02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,314 @@
...
+If you want a less curated dataset, run the chunk above without specifying a `category` to the `msigdbr()` function.
+
+In our differential expression results data frame, `dge_df` we have Ensembl gene identifiers. 
+So we will need to convert our Ensembl IDs into either gene symbols or Entrez IDs for GSEA.
+
+## Gene identifier conversion
+
+We're going to convert our identifiers in `dge_df` to Entrez IDs, but you can, with the change of a single argument, use the same code to convert to many other types of identifiers!
+
+The annotation package `org.Dr.eg.db` contains information for different identifiers [@@Carlson2019-zebrafish].
```
cansavvy, line 224:
```suggestion
The annotation package `org.Dr.eg.db` contains information for different identifiers [@Carlson2019-zebrafish].
```

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,314 @@
...
+    column = "ENTREZID", # Replace with the type of gene identifiers you would like to map to
+    keytype = "ENSEMBL", # Replace with the type of gene identifiers in your data
+    multiVals = "first" # Where there are multiple mappings, the first mapped value will be returned
+  )) %>%
+  # If an Ensembl gene identifier doesn't map to a Entrez gene identifier, drop that
+  # from the data frame
+  dplyr::filter(!is.na(entrez_id))
+```
+
+This message is letting us know that sometimes Ensembl gene identifiers will map to multiple Entrez IDs.
```
cansavvy, line 258: Specificity about the message is good, in case for some reason a different message prints out or they change the method and no longer see this message. 
```suggestion
This `1:many mapping between keys and columns` message means that some Ensembl gene identifiers map to multiple Entrez IDs.
```

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,314 @@
...
+  dplyr::group_by(entrez_id) %>%
+  # Count the number of rows in the data frame for each Entrez ID
+  dplyr::count() %>%
+  # Filter out the IDs with multiple mappings
+  dplyr::filter(n > 1)
+```
+
+It looks like we have two Entrez IDs that were mapped to multiple Ensembl IDs.
+For the purpose of performing GSEA later in this notebook, we will filter out the duplicated Entrez IDs.
+For more about how to explore this, take a look at our [microarray gene ID conversion example](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html).
```
cansavvy, line 281: I wonder if we want to spend as much time on the gene ID mapping since we have separate examples for that? We could just skip right to using `multiVals = "filter"` since this is essentially what you are doing here but manually. 

That being said, we may not want to drop data here and I don't think we are as particular about which gene ID is used, so maybe we should just simplify,  use `multiVals = "first"`, tell them to see the mapping examples for more info and then move on?
https://www.rdocumentation.org/packages/AnnotationDbi/versions/1.30.1/topics/AnnotationDb-objects

cansavvy (commented):
I think this looks very good! I just have a comment/question about maybe trimming down the mapping section, let me know what you think. 

In a different note: this new PR process seems working well review-wise, how do you feel it's working on your end?

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: I believe I implemented `multiVals = "first"` as you suggest here in the last commit. I used an `inner_join()` to join the expression data as to not have NAs in the `entrez_id` column (which may pose an issue later when running `GSEA()`). Please let me know if this is what you intended @cansavvy or if I should make any further changes here.

cbethell: I will note however, that this still leaves us with two duplicated Entrez IDs that map to multiple Ensembl IDs resulting in the following warning message when running `GSEA()` later in the notebook: `There are duplicate gene names, fgsea may produce unexpected results.`

cbethell:
@cansavvy I believe I addressed your comments so this is ready for another look. Note my [comment here](https://github.com/AlexsLemonade/refinebio-examples/pull/345#discussion_r515511334) however, as using `multiVals = "first"` still leaves us with  duplicated Entrez IDs that map to multiple Ensembl IDs.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cansavvy: We should probably show users that these duplicates exist then and how to deal with them. The fact that there are only two instances of this is kinda of annoying but its probably good we have this come up so we can show users how to deal with it -- in other datasets or species its possible it will come up more (or maybe not at all). 

I think we should incorporate two steps (that should maybe be their own chunk). 

1) Show users how to test for if there are multiple entrez ids. A TRUE/FALSE like `any(duplicated())` would probably work. 

2) Show one way that you can decide on which entrez id's data to keep (here's where there could be a lot of ways to do this but we will just have to pick one that we think will be generally useful in most contexts. 
I think an okay way to do this would be to keep the data for the entrez gene id with the higher t value (or lower p value) since it will be of greater interest. 

May be good to get a @jashapiro opinion on this.

cansavvy:
I think this is almost ready! Just need to make some decisions about the multiple entrez ids: https://github.com/AlexsLemonade/refinebio-examples/pull/345#discussion_r516126323

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
jaclyn-taroni: > I think an okay way to do this would be to keep the data for the entrez gene id with the higher t value (or lower p value)

Picking whatever entry has the larger absolute value for the stat we use for ranking makes sense to me. (Are we still using LFC?) A caveat we should point out is that the genes that have duplicate identifiers could be enriched in a particular pathway/gene set and you may get an overly optimistic view of how perturbed that pathway is using this approach.

cansavvy: >  Are we still using LFC?

There is log fold change. And based on the [draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/339/files) LFC will be used for GSEA.

cbethell: We are using the `t` value now (per your [comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/339/files#r513817275) @jaclyn-taroni).

In the [last commit](https://github.com/AlexsLemonade/refinebio-examples/pull/345/commits/6fa5e4319a5707199f22036d69a656198bd083a9), I added two steps, one to check for duplicate identifiers and the other to sort by `t` and remove the lower duplicate value.

Let me know if you think this is the best approach given the suggestions above @cansavvy and @jaclyn-taroni.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -67,7 +67,7 @@ In the same place you put this `.Rmd` file, you should now have three new empty
 ## Obtain the gene set for this example
 
 In this example, we are using differential expression results table we obtained from an [example analysis of zebrafish samples overexpressing human CREB experiment](https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_01_2-groups.html) using [`limma`](https://bioconductor.org/packages/release/bioc/html/limma.html) [@Ritchie2015].
-The table contains Ensembl gene IDs, log fold-changes, and adjusted p-values (FDR in this case).
+The table contains Ensembl gene IDs, t-statistic values, and adjusted p-values (FDR in this case).
```
cansavvy, line 70: Did we change the output table? What does the output table actually have? I thought both were included?

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -270,6 +270,25 @@ Let's see a preview of `dge_mapped_df`.
...
+
+```{r}
+filtered_dge_mapped_df <- dge_mapped_df %>%
+  # Sort so that highest t-statistic values are at the top
+  dplyr::arrange(dplyr::desc(t)) %>%
+  # Filter out the duplicated rows
+  dplyr::filter(!duplicated(entrez_id))
+```
+
+Note however, that a caveat in using this approach is that the genes that have duplicate. identifiers could be enriched in a particular pathway/gene set and we may get an overly optimistic view of how perturbed that pathway truly is.
```
cansavvy, line 291: Did you intend to make this two sentences or is this `.` a typo?

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,318 @@
...
+Looks like we do have duplicated Entrez IDs.
+We do not want duplicated gene identifiers for the GSEA steps later, so let's keep the Entrez IDs associated with the higher t-statistic value.
+
+```{r}
+filtered_dge_mapped_df <- dge_mapped_df %>%
+  # Sort so that highest t-statistic values are at the top
+  dplyr::arrange(dplyr::desc(t)) %>%
+  # Filter out the duplicated rows -- this will keep the first row with the
+  # duplicated value thus keeping the row with the highest t-statistic value
+  dplyr::filter(!duplicated(entrez_id))
```
cansavvy, line 288: I think `distinct()` is a more direct version of the `dplyr::filter(!duplicated())` you have here. 

Second question, can we prove this to ourselves a bit? Perhaps as simply as printing out one of the duplicated entrez IDs and their t values before and after?  (I don't want to add too much length to these steps, but I also think its good to make data removal steps proved and clear).

cansavvy (commented):
Just a few little comments/requests and then I think this is ready from my end.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
jaclyn-taroni: Did you find any evidence to support my thought that _t_ would be more "standard?" That was mostly based on recollection.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: Yes both are included, however there are other values that are included in the table that are not listed here (i.e. `B`, `AveExpr`, and `P.Value`).

I figured if we were not mentioning all of the columns in the table, we would mention only those that we would be using in the notebook. However, I can adjust this to mention all 8 columns that we have in the output table.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: That's definitely a typo, good catch!

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cansavvy: In that case, let's do a slight wording change:

```suggestion
The table contains summary statistics including Ensembl gene IDs, t-statistic values, and adjusted p-values (FDR in this case).
```

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: The reason I opted to use `dplyr::filter(!duplicated(entrez_id))` is because `dplyr::distinct(entrez_id)` returns only the column `entrez_id` while `dplyr::distinct()` returns all the rows containing duplicate identifiers (since their `t` values etc. are different). Perhaps my implementation of `dplyr::distinct()` is incorrect in this case?

Also, I agree with your second point here. While developing, I used the following to find the duplicate ids and their associated data as a sanity check:

`dge_mapped_df %>%
  dplyr::filter(duplicated(entrez_id))`

However, this returns just one of the rows with each of the duplicate ids (I manually searched the before and after data frames for the associated data using the exact `entrez_id` value returned)

Screen Shot 2020-11-04 at 12 48 04 PM

cansavvy: > `dplyr::distinct(entrez_id)` returns only the column `entrez_id`

Ah. There's an `.keep_all = TRUE` argument for `distinct()` that you need to use to not drop columns. 

> Also, I agree with your second point here. While developing, I used the following to find the duplicate ids and their associated data as a sanity check:
> dge_mapped_df %>% dplyr::filter(duplicated(entrez_id))

I was trying to find a straight forward way of doing this without having to make a separate duplicate entrez ids object, but I didn't come up with anything that's great. I was hoping `duplicated()` had an option to return values directly but it doesn't. I also looked to see if tidyverse had a reverse of `distinct()` but it doesn't seem to. If we installed another package (which I don't want to do) we could use `janitor::get_dupes()` but I don't find that worth  having users install another package for. 

So we are left with doing the manual preview you used here -- which I think may be the simplest route for users to follow and still get the point across. OR, this kind of thing: 
```
dup_entrez_ids <- dge_mapped_df %>% 
  dplyr::filter(duplicated(entrez_id)) %>%
  dplyr::pull(entrez_id)
```
where you then have to use `dup_entrez_ids` to retrieve things, but you'd still have to do an arrange. 

I think we should just stick with your simple and effective use of an example entrez id like `336702`.

cansavvy (commented):
I think after you decide about `distinct()` and the other info in the comment I just left, this looks good by me. :+1: Request a review from Jackie when you are ready for her to take a look.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: In the most recent commit, I went ahead and replaced the `dplyr::filter(!duplicated(entrez_id))` step with `dplyr::distinct(entrez_id, .keep_all = TRUE)` as you suggested @cansavvy, and left the subsequent steps as is because I also believe it is not worth having users install another package for. I do wish `distinct()` had a reverse function but I believe what we currently have is the next best simple yet effective solution.

cbethell:
@jaclyn-taroni this PR is ready for a second review whenever you are ready!

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+The data we're interested in here comes from zebrafish samples, so we can obtain just the gene sets relevant to _D. rerio_ by specifying `species = "Danio rerio"` and only the Hallmark gene sets by specifying `category = "H"` to the `msigdbr()` function.
+
+```{r}
+dr_hallmark_df <- msigdbr(
+  species = "Danio rerio", # Replace with species name relevant to your data
+  category = "H"
+)
+```
+
+If you want a less curated dataset, run the chunk above without specifying a `category` to the `msigdbr()` function.
```
jaclyn-taroni, line 215: "less curated" seems like an implicit value judgment to me, which we generally want to avoid. If you run this without specifying `category`, I believe `msigdbr()` will return all the MSigDB gene sets for zebrafish. I'd write that instead.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: Linking a comment that answers this question from PR 2 of 2 #347 here:

"""
The decision to create the pre-ranked gene vector based on the t-statistic rather than log fold change was explored based on [this comment from the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/339#discussion_r513817275) and eventually made based on what is recommended in [Discovering statistically significant pathways in expression profiling studies](https://www.pnas.org/content/pnas/102/38/13544.full.pdf) and the explanation from a [biostars forum](https://www.biostars.org/p/200715/) which says "Try to understand how they relate to the question you're interested in e.g. if you're most interested in effect size then the fold change is what you should use but if you're more interested in statistical significance then look for one of the statistics taking into consideration the assumptions they make e.g. t-test" as I believe we are encouraging users to look at statistical significance here.
"""

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+
+In our differential expression results data frame, `dge_df` we have Ensembl gene identifiers. 
+So we will need to convert our Ensembl IDs into either gene symbols or Entrez IDs for GSEA.
```
jaclyn-taroni, line 218: This, as far as I can tell, at no point _shows_ `dr_hallmark_df` or mentions that it contains gene symbols or Entrez IDs. (If this is a remnant of following what we have training, recall that the GSEA notebook was the second thing folks saw during the same instruction period where this package was introduced in the first notebook.)

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+For the purpose of performing GSEA later in this notebook, we keep only the first mapped IDs.
+For more about how to explore this, take a look at our [microarray gene ID conversion example](https://alexslemonade.github.io/refinebio-examples/02-microarray/gene-id-annotation_microarray_01_ensembl.html).
+
+Let's see a preview of `dge_mapped_df`.
+
+```{r}
+head(dge_mapped_df)
+```
+
+Now let's check to see if we have any Entrez IDs that mapped to multiple Ensembl IDs.
```
jaclyn-taroni, line 273: _Why_ are we checking this at this point? A reader probably won't know.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+
+Now let's take a look at the rows associated with one of the duplicated Entrez IDs.
+
+```{r}
+dge_mapped_df %>%
+  dplyr::filter(entrez_id == "336702")
+```
+We can see that the associated values vary for each row although the Entrez ID is the same.
+
+We do not want duplicated gene identifiers for the GSEA steps later, so let's keep the Entrez IDs associated with the higher t-statistic value.
```
jaclyn-taroni, line 296: I would explain this _sooner_ but it's probably good to repeat oneself!

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+```
+
+Looks like we do have duplicated Entrez IDs.
+Let's find out which Entrez IDs have been duplicated.
+
+```{r}
+dge_mapped_df %>%
+  dplyr::filter(duplicated(entrez_id)) %>%
+  dplyr::pull(entrez_id)
+```
```
jaclyn-taroni, line 286: I think my preference would be the approach where we assign these ids to a vector per https://github.com/AlexsLemonade/refinebio-examples/pull/345#discussion_r517591714. But you'd have to do:

```R
dup_entrez_ids <- dge_mapped_df %>% 
  dplyr::filter(duplicated(entrez_id)) %>%
  dplyr::pull(entrez_id)

dup_entrez_ids
```

to illustrate the point. My rationale:

* This isn't that much more complicated than the example where you use `entrez_id == "336702"` in my opinion.
* It's more general than `entrez_id == "336702"`, as in it is something users could potentially use if they adapt this to their own dataset, etc.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,343 @@
...
+
+Let's check to see that we removed the duplicate Entrez IDs and kept the row with the higher t-statistic value.
+
+```{r}
+filtered_dge_mapped_df %>%
+  dplyr::filter(entrez_id == "336702")
+```
+
+Looks like we were able to successfully get rid of the duplicate gene identifiers and keep the observation with the higher t-statistic value!
+Note however, that a caveat in using this approach is that the genes that have duplicate identifiers could be enriched in a particular pathway/gene set and we may get an overly optimistic view of how perturbed that pathway truly is.
```
jaclyn-taroni, line 316: Move to ~line 296 and _expand upon_. At this point, you haven't explained GSEA yet (https://github.com/AlexsLemonade/refinebio-examples/pull/339/files#diff-0d0fd22250c391b41f655f78697b0bcbd9626a54072e31fc31cbd01c6faf295dR295)! So I'd also think about, now that we need to explain scientific decision making during the gene identifier conversion, if we want to move up that explanation to before the gene ID conversion.

jaclyn-taroni (commented):
This looks pretty good! Before it goes in, I think we need to make sure we're explaining and showing folks everything they need to do as they're doing it. Regarding the point about moving up the GSEA explanation, I think we can work that over on #347.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: Good point, will make the above change in the upcoming commit 👍

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: I moved this line up to ~line 296 in this PR and left a comment on #347 to address the latter part of this comment: https://github.com/AlexsLemonade/refinebio-examples/pull/347#discussion_r518297547

cbethell:
@jaclyn-taroni, I believe I addressed your above review comments with exception of moving the GSEA explanation up tp before the gene ID conversion. I left a comment on #347 (https://github.com/AlexsLemonade/refinebio-examples/pull/347#discussion_r518297547) to address that as suggested.

Please let me know if I misinterpreted any of your comments or need to make additional changes to satisfy your comments.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,355 @@
...
+
+We can see that the associated values vary for each row.
+
+As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSEA steps later, so let's keep the Entrez IDs associated with the higher t-statistic value.
+Note however, that a caveat in using this approach is that the genes that have duplicate identifiers could be enriched in a particular pathway/gene set and we may get an overly optimistic view of how perturbed that pathway truly is due to the removal of the duplicate gene identifiers and their associated values.
+
+```{r}
+filtered_dge_mapped_df <- dge_mapped_df %>%
+  # Sort so that highest t-statistic values are at the top
+  dplyr::arrange(dplyr::desc(t)) %>%
```
jaclyn-taroni, line 314: I missed this in my first review. I would expect this to be based on _absolute_ value, i.e., whichever value is most likely to be highly- or lowly-ranked (or further away from the center of the ranking depending on how you'd like to talk about it) https://github.com/AlexsLemonade/refinebio-examples/pull/345#discussion_r517329319

cbethell: Gotcha, that does make the most sense here. I overlooked this in the first comment on #345 but believe I implemented it in the [most recent commit](https://github.com/AlexsLemonade/refinebio-examples/pull/345/commits/d64e208aaa6f4f23e0460665437cedf149449856). Please let me know if I missed an important step in the implementation @jaclyn-taroni.

cbethell: Worth noting, when running the [`GSEA()` step](https://github.com/AlexsLemonade/refinebio-examples/pull/347/files#diff-3238529b18500df7a3f9997a49913031d45ac77f0d919ae4a7a3aa1be17d3889R363) included in PR #347 throws the following error when the vector is sorted based on absolute value:

`Error in GSEA_internal(geneList = geneList, exponent = exponent, minGSSize = minGSSize, : geneList should be a decreasing sorted vector...`

jaclyn-taroni: I'm not saying we should sort the vector we pass `GSEA()` by absolute value (provided I am following you correctly), only that we should select the duplicate instances with a greater absolute value.

jaclyn-taroni (approved):
I returned my review early so I could respond to your comment. See what you think of what I wrote. I expect that we will need to continue to refine that language over on #347 depending on how and where we introduce GSEA.

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
cbethell: Ah gotcha! Makes even more sense now thinking about it 👍

02-microarray/pathway-analysis_microarray_03_gsea.Rmd
```diff
@@ -0,0 +1,355 @@
...
+
+```{r}
+dge_mapped_df %>%
+  dplyr::filter(entrez_id %in% dup_entrez_ids)
+```
+
+We can see that the associated values vary for each row.
+
+As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSEA steps later, so let's keep the Entrez IDs associated with the higher absolute value of the t-statistic.
+Note however, that a caveat in using this approach is that the genes that have duplicate identifiers could be enriched in a particular pathway/gene set and we may get an overly optimistic view of how perturbed that pathway truly is due to the removal of the duplicate gene identifiers and their associated values.
```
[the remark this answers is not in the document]
cbethell: I think this language looks good. I agree that we can refine it over on #347 but in the meantime this provides more detailed context 👍

Issue #346, opened 2020-11-04
[Investigation/Discussion] Methods/packages for identifying co-expressed gene modules


cansavvy:
### Background

We were discussing other use cases for ORA for RNA-seq and co-expressed gene modules was something @jaclyn-taroni mentioned as an idea: https://github.com/AlexsLemonade/refinebio-examples/issues/344#issuecomment-721812075

This issue is about what co-expressed gene module finding methods we may want to consider.

### Problem

What methods currently are available and recommended for finding coexpressed gene modules. 

I know of [WGCNA](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-9-559) but there's probably other methods out there that are newer and maybe one that we should recommend to our users instead of WGCNA. This issue is about taking a look at what methods and packages like WGCNA there are out there. 

### What are the recommended next steps?

I'm going to do some investigating online and I will post what I find here for discussion, but if anyone has any leads or recommendations on methods/packages for identifying coexpressed genes, please post here and we can discuss!

Obviously a scientifically sound method is important, but for our contexts, R-compatibility, good documentation are also key components in this evaluation

After we determine what and if there's a method we should make a new analysis example for, I will create a new analysis issue that is more specific based on that method that we may determine here.

cansavvy:
This review paper seems relevant: https://www.sciencedirect.com/science/article/abs/pii/S0010482519302574?via%3Dihub and compares WGCNA to THD-Module Extractor, DiffCoEx, and MODA.

cansavvy:
Here's an idea that might be appealing to our users: https://github.com/hidelab/diffcoexp DiffCoExp tries to find groups differentially co expressed genes. It claims it works with RNA-seq but it also sounds like it uses similar methods to WGCNA, so we'd need to look into whether the methods it borrows from WGCNA are the same ones that are problematic/suboptimal for RNA-seq. (I don't know at this time).

cansavvy:
Another contender I just found: `petal` is an R package that its authors describe as better suited for RNA-seq data specifically because it does not assume normality. 
[Petereit et al, 2016.pdf](https://github.com/AlexsLemonade/refinebio-examples/files/5494793/Petereit.et.al.2016.pdf)
Github: https://github.com/julipetal/petalNet

cansavvy:
^ Looks like it hasn't been updated in ~5 years : (

cansavvy:
Here's a brief summary of other ideas of come across: 

[...]

Issue #306, opened 2020-10-16
Restructure ORA example instruction -- more concise


cansavvy:
Related to this discussion: https://github.com/AlexsLemonade/refinebio-examples/pull/283#discussion_r505907273

The microarray ORA example needs some reorganization the instruction and steps used. 

Specifically, we should make it a bit more terse and specific but also look out for where we can highlight other decisions/methods that could be used. Should these other methods/decisions be their own example -- either an rna-seq example or a part of a set of ORA? OR can we just highlight with comments and further instruction how to alter the ORA example to fit one's needs; ultimately getting across that the decisions we made are somewhat arbitrary and are not an implicit rule (Josh's words).

cansavvy:
Another more specific direction as a part of this issue is that some of the background information: see @jaclyn-taroni's  comment here: https://github.com/AlexsLemonade/refinebio-examples/pull/283/files#r506455366 can be moved to the pathway analysis introduction file (#214). 

Because this is the first pathway analysis module to be (re)added its conceivable that as we move along we will hammer out a strategy for this background information as well as what the "prescriptive but not a rule" instruction discussed here should look like.

cansavvy:
This and the intro paragraph determined in #349 can be addressed at the same time.

cansavvy:
An idea related to these changes from @envest: https://github.com/AlexsLemonade/refinebio-examples/pull/356#discussion_r527240743 Use a table. I think a lot of the info needs to be trimmed down here, but we can see if a table of sorts makes explanations more brief. (But we also might just be able to get rid of the explanations entirely and link out to somewhere that explains it well (a strategy we've used a lot when we are trying to be brief but still give info).

cansavvy:
Found a nice source for explaining hypergeometric tests in context of genes: https://dputhier.github.io/ASG/practicals/go_statistics_td/go_statistics_td_2015.html

jaclyn-taroni:
In my opinion, this was fully addressed by #377.

Issue #344, opened 2020-10-30
New Analysis Example: RNA-seq Pathway analysis -- ORA 


cansavvy:
### What are the goals of this new example analysis?

We show how to use microarray (limma) results to do ORA: https://github.com/AlexsLemonade/refinebio-examples/issues/206

It might be good to show how to do ORA with RNA-seq (DESeq2 output) -- but we should keep in mind that these will likely not differ too much. This will start out as an exploration of  _how_ different these examples might be? This will help inform the organizational discussion on #340 

If we find that these RNA-seq vs microarray examples are going to be SO identical, we may want to reconsider having separate technology examples (related to #223 and #175) and this is something we can comment about on #340 

Alternatively, if we do want to maintain separate technology examples, we can take the strategy of illustrating different aspects/strategies in this ORA example as compared to what is shown in the microarray example. Aka, maybe we could decide on a gene list using a different method and make different plots??

### What kind of dataset will this need?

This will need RNA-seq differential expression results, which currently for RNA-seq we have one option already prepared: 03-rnaseq/differential-expression_rnaseq_01.html (until #242 is completed that is).

This will involve using this file specifically: SRP078441_differential_expression_results.tsv which is based on AML patients. 
We haven't used this for pathway analysis yet. If it turns out to be an insignificant dud, we may have to look into completing #242 and trying the results from that instead. 

### What steps should be included in this analysis?

The steps should for the most part follow what is being used for microarray: https://github.com/AlexsLemonade/refinebio-examples/blob/staging/02-microarray/pathway-analysis_microarray_02_ora.Rmd

We'll have to change the steps to use human though instead of zebrafish and also consider what other alternative decisions we may want to show users as possibilities. 

### What packages/methods do you recommend using or looking into for this analysis?

Same as the microarray example, we should still be able to use clusterProfiler.

jaclyn-taroni:
In my opinion, I think we can get started on this before we have a decision on #340. (@cansavvy might share that opinion, but it's hard to tell based on this issue.) Specifically, we can start to list all the ways this could potentially differ from the microarray example. 

[...]

PR #350, opened 2020-11-11
WIP: WGCNA module draft

cansavvy:
## Analysis Purpose

It's a spin off issue from #306 and the discussion on #346. 

Ultimately I was trying to see if we could squeak in a quick WGCNA run for ORA in one module, but as you can see here, it's not really realistic for the. WGCNA run to be that short particularly since we also need to do some DESeq2 transformations before we can get to WGCNA. 

## Pull Request Stage

## Strategy

I've been trying to sort through what seemed like the important points of the [WGCNA tutorials](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/) and: 
- Make them tidyverse and ggplot2 
- Put more explanation about what things mean for the user

Also made sure we are following the [RNA-seq guidance that's written in the FAQ of WGCNA](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/faq.html#:~:text=Can%20WGCNA%20be%20used%20to,Yes.&text=Whether%20one%20uses%20RPKM%2C%20FPKM,were%20processed%20the%20same%20way.).
 
- In the "Explore WGCNA" results section you will see I'm doing a number of things, making multiple plots. I in no way expect we keep all those things, but instead I want to know which things we find most useful and we can delete the others and I'll expand a tad more on the useful things. Here's a summary of the things I'm proposing (some of which are inspired by plots I see in WGCNA tutorials): 
    - Show users how to save the which gene is which module key (this seems necessary/useful)
     - A metadata labeled, sample x eigengene heatmap
    - A module eigengene x module eigengene heatmap

- You'll see I briefly mention/explain `TOMtype` parameter. Do we think this is unnecessary/out of scope for our users? 
- I've been looking at the tutorials and papers and forums and I can't for the life of me find an intuitive explanation for what "power" for soft-threshold network construction actually is (not info on how to use it, but like what does it mean?? What does WGCNA do with it). So if anyone has insight on that, I couldn't figure it out. 
- I have set the seed in the main WGCNA function `blockwiseModules()` but it sounds like a global set the seed might work instead? I suppose global is preferred or no? 

### Content checks
* [ ] All `{{BLANKS}}` have been replaced with the correct content.

### Formatting Checks

### Add datasets to S3


cansavvy:
[network-analysis_rnaseq_01_wgcna.html.zip](https://github.com/AlexsLemonade/refinebio-examples/files/5525630/network-analysis_rnaseq_01_wgcna.html.zip)

Here's the rendered html if you wanna see what the draft plots look like without checking out the branch.

jashapiro:
I'll start with the elephant, which is that I share your concerns about this being a bit of a bear (mixed animal metaphors for the win). I'm not 100% sure how I feel about that on the whole. It does seem like something that might end up in "advanced" rather than the main set of analyses.

I am also not sure whether you are intending to keep this described as ORA or if that is just leftovers from the template. I see it as quite distinct from a standard ORA, and I would not refer to it as such. Maybe you are planning to use these modules as the sets for ORA, but that would seem rather circular to me.

Skipping to the end, I feel like we would want some kind of rough conclusion for users to get out of this analysis. Is there a way we can reduce the number of modules that are returned to get the ones we might care most about? What do we do with these modules once we find them?  I think the relationship of the eigengenes to the metadata labels gets closest to answering this question, so it seems like the best place to focus effort.

To make this easier to interpret, we might try to come up with ways to reduce the number of Eigengenes that we get out. That might involve adjusting and examining some parameters and thresholds, which I imagine is work that might not make it into an example notebook, but should be done. The thing we need to think about is how to provide a balance between a single example that looks pretty good and the fact that parameters and results should be examined carefully.

Answering your question about `power`: I think based on my reading that the `Power` is the exponent to which they raise the correlations to before thresholding & dividing the correlation network into modules. This separates the really high correlations from the just kind of high correlations. At least that is my read. (0.99^14  vs 0.98^14 is a much larger difference than 0.99 vs. 0.98, and I imagine that helps with the thresholding steps).  So yeah, the first definition of power you learned, not statistical power, which is what I would have assumed based on no context (except that the values make no sense as statistical power).

A couple of specific notes:

- `power =  14` may be too low for this example, as it makes the threshold just below 0.8, not above it. Not that this seems like it should be a hard rule. This and other parameters probably require tweaking, which is too much to cover here, but we should present an example that looks "good". If you raise the power, do you get more or fewer eigengenes?

- It would probably make sense to use numeric module labels rather than the color names, especially as you are not plotting those colors at all. (Also, I think "grey" means no module as the equivalent to module 0, which is worth noting).

- Handy tidyverse function: `tibble::enframe()`
You can replace this:
```
gene_modules_df <- data.frame(gene_modules) %>%
  tibble::rownames_to_column("Gene")
```
with:
```
gene_modules_df <- tibble::enframe(gene_modules, name = "gene", value = "module")
```
To me that is more clear what is going on, and does it all in one neat step.
Though really, for this use case, just `table(gene_modules)` would be sufficient. I don't think we need to tidy all the things.

cansavvy:
@jashapiro, This was my attempt to "simplify" the WGCNA tutorials, but I think you are correct that it is still too much for a non-advanced topics user to go through. I tried to give just enough parameter guidance as to not be totally cavalier with the analysis but I think if someone were to really get into this the right way, they should dive into the parameters more than the user of this section may be willing or able to do. So... I think moving this to advanced-topics makes sense. 

To clear up the confusion about is this ORA or not: 
Although this is not supposed to be ORA anymore, the idea was wee would take one of the gene modules from this analysis to do ORA -- in its own separate example notebook #344. (The ORA comments in the notebook leftover are remnants from the older version of this draft which _was_ ORA but I wanted to get big picture thoughts before doing anymore polishing)

So, this means as far as timeline, we need to discuss the further work on this since the priority for this sprint is getting the pathway analyses and this was a tangent that was supposed to feed into an ORA, pathway analysis example. That being said it doesn't make sense to have an advanced topics example feed into a "main" example, so we will have to reevaluate where we are getting a gene set from for the ORA example #344. 

In other notes: 
- Thanks for your interpretation on power. I had a suspicion it related to exponents but couldn't figure out for sure what it meant. 
- I like your other suggestions and will implement them after we figure out "big picture" where this example is going.

jashapiro:
> To clear up the confusion about is this ORA or not:
> Although this is not supposed to be ORA anymore, the idea was wee would take one of the gene modules from this analysis to do ORA -- in its own separate example notebook #344. (The ORA comments in the notebook leftover are remnants from the older version of this draft which _was_ ORA but I wanted to get big picture thoughts before doing anymore polishing)

Is this a thing that people do? Because as I said, it seems kind of circular. I can see taking modules from one data set and using them for ORA in another data set, but if you are using modules defined in the same data set, I would assume that all the genes will be significant or none of them will be, as the modules are defined by the correlations between genes...

jaclyn-taroni:
The question I was after is "Does this coexpression module significantly overlap with known pathways?" So given a module that is associated with metadata, for example, are p53 pathway genes overrepresented?

> That being said it doesn't make sense to have an advanced topics example feed into a "main" example, so we will have to reevaluate where we are getting a gene set from for the ORA example 

I'm not outright opposed to having an advanced example feed into a "main" example. We're not sure people will ever look at more than one example and our point for getting on this tangent was basically that any list of genes of interest will work.

cansavvy:
> I'm not outright opposed to having an advanced example feed into a "main" example. We're not sure people will ever look at more than one example and our point for getting on this tangent was basically that any list of genes of interest will work.

I suppose that's fair. We usually link to the notebook where the data come from though, in case the user wants to produce the data themselves, so I guess we'd have to ask if it's okay to link a user to a more advanced topic or will that be overwhelming. It's obviously unclear to us at this stage how users will be navigating these kinds of examples with predecessors so perhaps it's not a problem.

jashapiro:
> The question I was after is "Does this coexpression module significantly overlap with known pathways?" So given a module that is associated with metadata, for example, are p53 pathway genes overrepresented?

Ah, that makes more sense... 

I am also not opposed to using an advanced example to feed back in, but it does seems a bit of a slippery slope. It could add confusion as to the "why" for a particular analysis.

cansavvy:
> I am also not opposed to using an advanced example to feed back in, but it does seems a bit of a slippery slope. It could add confusion as to the "why" for a particular analysis.

I think the plan is to move forward as is, but when we add a more simple method of getting gene cluster example, we can replace it in ORA. See here: https://github.com/AlexsLemonade/refinebio-examples/issues/344#issuecomment-725679614

cansavvy:
So it seems we probably want to end up with a gene list from a module that is most related to a metadata variable of interest. This seems most pertinent to what users would want to do as well as our downstream ORA application. This is somewhat related to my rough draft of a heatmap of eigengenes with metadata labels, but should probably be a little more methodical than that. 

[One of the WGCNA tutorials](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/FemaleLiver-03-relateModsToExt.pdf) suggests doing this kind of analysis through more correlations and a plot like this: 
Screen Shot 2020-11-11 at 5 43 51 PM

If we want to do this type of analysis, this means we'd actually have to have a meaningful numeric variable or more which means this dataset won't work and I'll have to do some digging to find a dataset that would work. (So meanwhile I'm not going to worry about playing with power if we switch datasets completely). 

However, I'm not crazy about more correlations, so if people have other suggestions of how to identify modules of most interest, I'm open ears. I'll do some digging and see if there's other legit ideas out there.

jaclyn-taroni:
You can test the module eigengenes for differential expression if you have a categorical variable that's of interest.

cansavvy:
> You can test the module eigengenes for differential expression if you have a categorical variable that's of interest.

I was thinking about that but didn't know if that'd be considered a "valid" method.

jashapiro:
> I was thinking about that but didn't know if that'd be considered a "valid" method.

It absolutely would be! Think about the eigengenes as a smaller set of genes that you can use to reduce multiple testing burdens. SO you can treat them as "gene" with measured expression, and then ask if the expression is different between the groups you care about. 

In the example you showed above, they are comparing lots of different continuous traits to eigengene "expression", and plotting all the correlation coefficients, but the much simpler case is when you have only one trait and you want to see which of the eigengenes is correlated with it. In the very simplest case, you can do something like a t-test comparing expression between two groups of samples. If you go from 20,000 genes to 100 eigengenes, that can give you a huge boost in power (statistical power here for realsies) because you don't have so much multiple testing to correct for.

jaclyn-taroni:
I wish GitHub had ☝️ as a reaction. @jashapiro is exactly right. WGCNA is both somewhat biologically motivated if we think that genes that are coordinated in expression might participate in a coherent process and technically motivated if we want to reduce the number of variables we need to test.

cansavvy:
So for differential expression do I make a new DESeq2 object from the eigengenes and use DESeq2 for differential expression per usual? Or do I do an ANOVA with a [base R function](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/anova)?

jaclyn-taroni:
I would probably keep it simple per Josh's suggestion.

cansavvy:
I'm probably going to open up a second Draft PR since much of this discussion has lead to quite a bit of changes and it'd probably be best to start fresh.

Issue #343, opened 2020-10-29
New Analysis Example: Microarray Pathway Analysis - GSVA


cansavvy:
### What are the goals of this new example analysis?

ORA and GSEA are certainly popular pathway analyses methods, but GSVA requires a bit less cutoffs and decision making so having this method as an example would probably be helpful for our users. 

Having a per sample pathway analysis results is a different question that GSVA can answer but the others can't so much. 

### What kind of dataset will this need?

We may want to use the same. original dataset we used in either GSEA or ORA so we have a comparison of pathway analyses?: GSE71270 (zebrafish CREB study) or GSE37418 (human medulloblastoma subtype).  

### What steps should be included in this analysis?

We can borrow some inspiration from https://github.com/AlexsLemonade/training-modules/blob/master/pathway-analysis/03-gene_set_variation_analysis.Rmd, keeping in mind that the narrative will need to change somewhat like other examples we've adapted from training to refinebio-examples: See #306 

1) Import `library(GSVA)` (add this to the Dockerfile)
2) Set up gene expression data as a matrix that that 
3) Import gene lists and decide about Hallmark or not (this decision should be made considering the discussion happening on https://github.com/AlexsLemonade/refinebio-examples/pull/339#discussion_r513813505 -- we'll want to. make sure users understand the implications of multiple testing corrections and how smaller gene sets can help with this.
4) Use `GSVA::gsva()` to perform GSVA, probably start out with largely the same parameters used in training but adjust if/when things look wonky. 
5) Display a preview of significant results in one way or another. Somewhat related to this discussion https://github.com/AlexsLemonade/refinebio-examples/pull/339#discussion_r513808472
6) Make some sort of visualization of the GSVA scores. Not sure what makes the most sense here? Plotting the top results and maybe a jitter plot by group? 
7) Write results to a TSV. 

### What packages/methods do you recommend using or looking into for this analysis?

[Probably `GSVA`](http://bioconductor.org/packages/release/bioc/html/GSVA.html) unless there are other package suggestions we should consider.

cbethell:
Based on a discussion with @cansavvy, the plan in the [original comment above](https://github.com/AlexsLemonade/refinebio-examples/issues/343#issue-732467563), and the [training modules example](https://github.com/AlexsLemonade/training-modules/blob/master/pathway-analysis/03-gene_set_variation_analysis.Rmd) for inspiration, the tentative plan for tackling this ticket is as follows:

[...]

PR #352, opened 2020-11-12
WIP: Add Microarray Pathway Analysis - GSVA example

cbethell:
## Analysis Purpose

This draft PR addresses issue #343.

## Pull Request Stage

## Strategy

I began by following the [plan](https://github.com/AlexsLemonade/refinebio-examples/issues/343#issuecomment-724255953) posted on issue #343 which is as follows:

1) Import `library(GSVA)` (will add this to the Dockerfile in an upcoming PR)
2) Read in gene expression data (the Homo sapiens medulloblastoma dataset which we also used in `differential-expression_microarray_01_several-groups.Rmd`)
3) Import gene list from broad institute url -- the original plan here was to use the recommendation from [GSVA vignette](https://bioconductor.riken.jp/packages/3.2/bioc/vignettes/GSVA/inst/doc/GSVA.pdf)  to read in the file (and isolate hallmark gene sets) but I implemented `qusage::read.gmt()` instead because the vignette requires that users load in a separate library and use the `data()` function to retrieve the gene sets which I believe may be a bit more complex that using  `qusage::read.gmt()` (context on making sure users understand the implications of multiple testing corrections and how smaller gene sets can help with this will be included in an updated PR). I will also note that when I attempted to use the method in the GSVA vignette to retrieve the hallmark gene sets, I was met with an error saying that the package is not available for the R version being used. 
4) Gene identifier conversion — mapping to human Entrez ids
5) Remove duplicate identifiers — using the highest absolute mean expression value
6) Use `GSVA::gsva()` to perform GSVA, using largely the same parameters used in [training](https://github.com/AlexsLemonade/training-modules/blob/3dbc6f3f53c680ec6aa2f513851c1cd4635cc31c/pathway-analysis/03-gene_set_variation_analysis.Rmd#L142) with **exception** of the `parallel.sz = 4` argument due to the following error message I got:  `Error in serialize(data, node$con, xdr = FALSE) : error writing to connection`. I therefore opted to use the default parameter which is `parallel.sz = 1L` (I removed this from the step and instruct users to use `?gsva` to see more options but should I still include `parallel.sz = 1L` although it is the default?)
7) Make some sort of visualization of the GSVA scores - Creating a heatmap using the GSVA scores and the associated metadata for annotation and a violin plot the scores by gene set/pathway.
8) Write results to a TSV. 

The added code in this PR (mainly the code related to reading the hallmark gene sets, the removal of duplicate gene identifiers using the higher absolute mean expression value, the main GSVA steps, and the visualization steps (include the preparation of the annotation data frame for the heatmap and the preparation of the data for the violin plots).

I will note here that the metadata does not contain a whole lot of information for this particular dataset, so the annotation on the heatmap only includes `refinebio_sex`. Comments on whether or not this is suffice for this use case, and suggestions for alternatives if this is not suffice, would be particularly appreciated!

I added the prompt **DRAFT** mainly before places where the context is rough and will be refined in upcoming PRs and **REVIEW** before code which could benefit from review at this stage.

Please let me know if anything above is unclear!
I am planning to break up this PR (likely into 2 separate PRs) for the purpose of refining the content in this PR and getting more detailed review comments so feel free to leave any high-level comments on this PR for implementation in the upcoming PRs.
cc: @cansavvy, @jaclyn-taroni, @jashapiro 

See the [rendered html](https://cbethell.github.io/open-pbta-output/pathway-analysis_microarray_04_gsva.html) for your reference.

### Content checks
* [ ] Seed is set (if applicable)

### Formatting Checks

### Add datasets to S3

* [ ] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar). (Will add in upcoming PR)
* [ ] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds. (Will add in upcoming PR)


02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+mapped_df %>%
+  dplyr::filter(entrez_id %in% dup_entrez_ids) %>%
+  dplyr::select(Ensembl, entrez_id, mean, dplyr::everything())
+```
+
+Now we can filter out the duplicate gene identifiers using the mean expression value.
+
+```{r}
+filtered_mapped_df <- mapped_df %>%
+  # Sort so that the highest absolute mean expression values are at the top
```
jaclyn-taroni, line 396: Why absolute value?

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+  dplyr::select(Ensembl, entrez_id, mean, dplyr::everything())
+```
+
+Looks like we were able to successfully get rid of the duplicate gene identifiers and keep the observations with the higher absolute mean expression value!
+
+As mentioned earlier, we need a matrix for GSVA.
+Let's now convert our data frame into a matrix and prepare our object for GSVA.
+
+```{r}
+matrix <- filtered_mapped_df %>%
```
jaclyn-taroni, line 418: This name has to change because it will conflict with `matrix()`. It's also not very descriptive!

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+  # First let's remove the columns we no longer need
+  dplyr::select(-c("Ensembl", "mean")) %>%
+  # Now we need to store our gene identifiers as rownames
+  tibble::column_to_rownames("entrez_id") %>%
+  # Now we can convert our object into a matrix
+  as.matrix()
+```
+
+Note that if we had duplicate gene identifiers here, we would not be able to set them as rownames.
```
jaclyn-taroni, line 427: I would have the first PR go to here.

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+
+Note that the `gsva()` function documentation says we can use `kcdf = "Gaussian"` if we had log-CPMs, log-RPKMs or log-TPMs, but we would use `kcdf = "Poisson"` on integer counts.
+See `?gsva` for more options.
+
+```{r}
+# Let's explore what the output of gsva() looks like
+head(gsva_results)
+```
+
+## Visualizing Results
```
jaclyn-taroni, line 473: I would probably pick one of these - this notebook is pretty long and it doesn't have all of the text we need yet! My inclination would be to leave the violin plot because I don't think we have many instances of this kind of wrangling in the examples and you could direct folks to the heatmap example(s) we already have.

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+
+### Perform GSVA
+
+**REVIEW**
+
+```{r}
+gsva_results <- gsva(matrix,
+  hallmarks_list,
+  method = "gsva",
+  # Appropriate for our transformed data on log2-like scale
```
jaclyn-taroni, line 450: Because this is a SCAN processed microarray dataset, you probably can say:
```suggestion
  # Appropriate for our log2-transformed microarray data
```

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
+---
+title: "Gene set variation analysis - Microarray"
+author: "CCDL for ALSF"
+date: "October 2020"
```
jaclyn-taroni, line 4:
```suggestion
date: "November 2020"
```

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+Now let's take a look at the rows associated with the duplicated Entrez IDs.
+
+```{r}
+mapped_df %>%
+  dplyr::filter(entrez_id %in% dup_entrez_ids)
+```
+
+We can see that the associated values vary for each row.
+
+As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSVA steps later, so let's keep the Entrez IDs associated with the larger mean expression value per sample.
```
jaclyn-taroni, line 371: This seems like a somewhat standard workflow for RNA-seq data because you might expect lowly expressed genes not to show up in your data, so you'd drop the instance of the identifier with low expression as estimated by the mean value across samples. For GSEA (not the pre-ranked version), the default way to collapse identifiers is to take the max value for multiple probes that map to  but you can apply the same idea to Ensembl IDs that map to the same Entrez ID ([ref](https://www.gsea-msigdb.org/gsea/doc/GSEAUserGuideFrame.html)). Other ways people do it (also at that ref) are to take the mean value or median value (should be less sensitive to outliers than the mean). Either way, it makes more sense to me to work with the _individual values_ (e.g., not based on a summary measurement across all samples) given that you're doing GSVA which is on a per sample basis. If you pick the max value, you're preferentially going to pick something with a higher rank and that caveat should get documented.

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+  ggplot2::theme_bw()
+
+# Display plot
+violin_plot
+```
+
+**DRAFT**
+
+What do you notice about these distributions?
+How might you use this information to inform your interpretation of GSVA scores?
```
jaclyn-taroni, line 535: I know this text is marked as a draft, but I want to make a larger point about the purpose of this plot. 

This looks like it's taken from the training material. In the training material, we generate a distribution of scores using random gene sets of different sizes to specifically illustrate the point about how the number of genes in a pathway/gene set influences the scores. I'm not sure what this violin plot is intended to illustrate if we're not going to include information about the gene set sizes. I don't know that we want to include that same experiment from training here (in fact I think we _don't_ want to include it). Instead, a more common use case for a violin plot is to look at the differences between groups. If you want to do that, you probably need to pick a different dataset with more metadata based on your comment when you filed the PR. Another idea would be to include the gene set size in the labels (e.g., `HALLMARK_ANGIOGENESIS (<number of genes>)`).

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
```diff
@@ -0,0 +1,631 @@
...
+```{r}
+gsva_results <- gsva(matrix,
+  hallmarks_list,
+  method = "gsva",
+  # Appropriate for our transformed data on log2-like scale
+  kcdf = "Gaussian",
+  # Minimum gene set size
+  min.sz = 5,
+  # Maximum gene set size
+  max.sz = 50,
```
jaclyn-taroni, line 455: This seems pretty low to me. Why did you pick this value?

jaclyn-taroni (commented):
The general structure of this looks good! The specifics of different choices need a bit of justification/work (the system is working)!

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
cbethell: Good question, this decision was made based on a [similar decision](https://github.com/AlexsLemonade/refinebio-examples/pull/345#discussion_r517329319) made in the GSEA example where it was suggested that we keep the higher absolute value for the purpose of ranking.

Although GSVA does not require a pre-ranked gene list like GSEA, this decision seemed to still make sense here as the row/observation with the larger absolute value would most likely be more interesting that one with a smaller value (whether it be a positive or negative value).

jaclyn-taroni: I bet you will respond to https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264 shortly, but some thoughts:

* By using the mean value across samples, you're making a decision based on all the samples but the GSVA scores are on a per-sample basis (part of the gist of the linked comment ☝️ )
* In general, the values that come out of refine.bio for quantile normalized microarray data (what you have here) are going to range from small negative values to higher positive values, where the small negative numbers are genes that are not highly expressed in a particular sample.

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
cbethell: Ah, this was actually unintentionally left over from development.
I had to do a bit of debugging which I thought the smaller parameters would assist with (as it seemed that it had timed out or had a memory issue) but now this should would just fine with the larger parameters so I will update this to use the paramaters `min.sz = 15` and `max.sz = 500` (tested it and it only takes a second or two longer than it does now).

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
cbethell: Gotcha, that all makes sense. I will use the default way and take the max value and be sure to document the caveat (as well as link to the [reference](https://www.gsea-msigdb.org/gsea/doc/GSEAUserGuideFrame.html)).

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
cbethell: Gotcha 👍

02-microarray/pathway-analysis_microarray_04_gsva.Rmd
cbethell: I did think that the gene set size experiment from training would not be well suited here but was not sure of the best alternative here, that said, I plan to hopefully find another dataset with more metadata. However, if that ends up being a bit more time-consuming than we would want, I'll go ahead and included the gene set size in the labels.

cbethell:
This can be closed now that #343 has been addressed with closed PRs #359 and #362.

Issue #349, opened 2020-11-10
Finalize pathway analyses text, including introductory paragraphs


cansavvy:
### Background

Based on implementing option 2: https://github.com/AlexsLemonade/refinebio-examples/issues/340#issuecomment-723272256, we need to draft up a paragraph with the essential info of how someone decides which pathway analysis they should use. 

### Problem

Currently the pathway analyses don't have background info about how to choose a pathway analysis based on your question and dataset. The original plan was to have an intro doc, but we discussed it on #340 and each pathway analysis will have its own brief intro info paragraph. 

### What are the recommended next steps?

The info will include the same kinds of concepts as are covered in the training slides but of course tailored for a self-teaching context as well as being more brief: https://docs.google.com/presentation/d/1WPuuN6KviEswaWVU0yBRe4XvFjPE0mT3Epd1Xr6TjMQ/edit#slide=id.p
Basically give the users guidance about which analysis they will want to use based on their questions and dataset's contents and link out to those other pathway analyses. 

1) Post a outline here. We should try to make it as brief as possible and link out to helpful sources. 
2) Draft a PR but only implement the paragraph in one of the analyses so the reviewer and author don't have to redundantly review and edit in each pathway analysis
3) In a second PR put that same paragraph in the other pathway analyses.

### Is there a particular timeframe for this issue?

Before going live probably.

cansavvy:
Here's the outline I'm thinking: 

- About this pathway analysis
   - Contains brief example-specific description 
- What is pathway analysis?
     - General info on what pathway analyses are
     - Link to Ten Years of Pathway Analysis: Current Approaches and Outstanding Challenges
- How to choose a pathway analysis?
    - Table that summarizes the current pathway analysis examples with these columns: 
         - Analysis name (with link)
         - What is required for input
         - What output looks like 
         - ✅ Pros
         - ⚠️ Cons

Unfortunately the `How to choose a pathway analysis?` will require updates whenever a new example is added. But the rest shouldn't have to be changed too much.

cansavvy:
Here's a general draft. If it looks decent, I will file a PR for the more detailed feedback.  


### About this pathway analysis
This example is one of the pathway analysis module set.



### What is pathway analysis?

We refer to any technique that uses predetermined sets of genes that are related or coordinated in their expression in some way (e.g., participate in the same molecular process, are regulated by the same transcription factor) to interpret a high-throughput experiment as pathway analysis.
[...]

Issue #215, opened 2020-09-14
Microarray Analysis Updates: Pathway Analysis - 03_qusage_meta_analysis.Rmd


cansavvy:
We need the main update items listed on #173 to happen here : 

> - Have the getting started template material added to it (# 146)
> - Update to adhere to the style guide (#148)
> - Add the dataset added to the S3 bucket. (See https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-s3-bucket)
> - Add to the Snakefile and render with snakemake on the Docker container. 

This notebook currently is the first one that uses a custom function. We may want to evaluate what extra background content is needed for this. 

This is also the first notebook that downloads a file other than a refine.bio file. We may want to discuss what this should look like.

PR #355, opened 2020-11-16
Delete intro Rmd and renumber

cansavvy:
## Purpose

This PR does some file deletions and renumbering for the pathway analysis. 

I changed the navbar and Snakefile file accordingly. This is in preparation for #349 but also just some tidying that could have be done at anytime.

This also involved doing some file neatening up since the 
- I deleted the intro file -- it is no longer need as #349 will be implemented next. 
- I deleted the qusage files since we can retrieve those from v1.0 whenever we need for #215 https://github.com/AlexsLemonade/refinebio-examples/blob/v1.0/
- Deleted `pathway-analysis_microarray_01_ortholog_mapping_kegg.Rmd` since it is obsolete with no plans of reviving it since ortholog_mapping does the job. 

### Issue addressed

Prepares us for #349 

## Gotchas the reviewer should know about

A general note: we will need to make sure the upcoming merges have the right numbering too. (Or we can file another small PR later after both GSEA and GSVA are in and before merges to master take place).


cbethell (approved):
LGTM! 

Per your general note:
> we will need to make sure the upcoming merges have the right numbering too

I will be sure to update any upcoming PRs appropriately.

PR #356, opened 2020-11-16
Add pathway analysis intro paragraph to microarray ORA

cansavvy:
## Purpose

This is the first part of #349 and I'm trying it out on the microarray ORA example. It's not much different than what is posted on #349, but I've obviously added example specific info as well as links. 

A lot of the wording has been borrowed from the original pathway analysis README: https://github.com/AlexsLemonade/refinebio-examples/blob/v1.0/pathway-analysis/README.md#what-is-pathway-analysis

**Note** The Files Changed is showing weird spacing things, but there are no actual changes after line 47 so you don't need to review beyond that point. 

## Issue addressed

First notebook for addressing #349 

## Remaining concerns and questions

- Does it contain enough information that a user could make a decent decision about what pathway analysis works best for their needs? 

- Besides the Khatri et al 2012 paper (which is linked to) do we have other references we want to send the user to for more info? 

- Along those same lines, should I add one more column to the table with the original paper references to those methods or do we just save that for the example's own descriptions?

PR #359, opened 2020-11-18
Pr 1 of 2: Add Microarray Pathway Analysis - GSVA example

cbethell:
## Analysis Purpose

This PR addresses issue #343.

## Pull Request Stage

The relevant draft PR is #352 

## Strategy

Per [this comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524294855), the decision was made to break up the notebook in that PR into two separate PRs, the first one stopping right before the execution of GSVA.

I mostly copied and pasted the code from the draft PR, refining the context/user instructions along the way.

Per [this comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524324537), I swapped out the dataset that was used in the draft PR for a [new dataset](https://www.refine.bio/experiments/GSE7696/glioblastoma-from-a-homogenous-cohort-of-patients-treated-within-clinical-trial) with more metadata, mostly for the purpose of plotting later in the notebook.

I also, per [this comment on the draft](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264), attempted to adjust the logic that was used in the draft PR to collapse or remove the duplicate gene identifiers.

From that comment, I understood that because GSVA is ran on a per sample basis, we would want to also make the decision of which duplicate gene identifier and associated data to keep based on _each individual sample's values_. That being said, I attempted the following logic to do so:

```
samples <- colnames(mapped_df[,-c(1:2)])

filtered_mapped_df <- mapped_df %>%
  dplyr::filter(entrez_id %in% dup_entrez_ids) %>%
  dplyr::pull(entrez_id, Ensembl) %>%
  as.data.frame() %>%
  dplyr::distinct(.)

for (sample in samples) {
  filtered_mapped_df[sample] <- mapped_df %>%
    dplyr::select(sample, entrez_id) %>%
    dplyr::filter(entrez_id %in% dup_entrez_ids) %>%
    dplyr::group_by(entrez_id) %>%
    dplyr::arrange(desc(sample)) %>%
    dplyr::distinct(entrez_id, .keep_all = TRUE)
}
```

which seemed to give me what we would want here:

Screen Shot 2020-11-17 at 6 15 46 PM

but also comes along with the following warning messages:

Screen Shot 2020-11-17 at 6 09 40 PM

Therefore, I made the decision to use a similar logic to what is in the draft PR, but instead using the maximum expression value to decide which observation to keep and it gives the _same results_ as the logic above.

Since this was [suggested](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264) by you on the draft PR @jaclyn-taroni, it would probably be ideal if you could take the first look at this particular section (starting around line 364) -- to provide any further insight in what's going right/wrong here.

The logic around collapsing duplicate gene identifiers as well as the added context re GSVA, etc. could all use an extra close look.

This may be hard to tell in this PR, but does this dataset seem to be better suited for the purpose this notebook (keeping in mind [#352 comment](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524324537))?

**Note**: I plan to update the introductory paragraphs in this notebook once #356 is approved and merged.

### Content checks

### Formatting Checks

### Add datasets to S3


cbethell:
@jaclyn-taroni, I am requesting your review specifically to get your feedback on the logic around collapsing duplicate gene identifiers as I mention in this PR's [original comment](https://github.com/AlexsLemonade/refinebio-examples/pull/359#issue-522808369) under "Strategy" (not necessarily for a full in-depth review of everything in the PR).

Please let me know what you think of the decision I made in this PR based on your [suggestion on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264) and my documented attempts seen above in the original comment.

jaclyn-taroni:
Would the following do it (not fully tested):

```R
mapped_df  %>% 
  dplyr::group_by(entrez_id) %>% 
  dplyr::summarize_all(max)
```

I expect it could be slow, but you could imagine filtering to the duplicated identifiers (which you did above), performing this step and then putting it all back together.

cbethell:
> Would the following do it (not fully tested):
> 
> ```r
> mapped_df  %>% 
>   dplyr::group_by(entrez_id) %>% 
>   dplyr::summarize_all(max)
> ```
> 
> I expect it could be slow, but you could imagine filtering to the duplicated identifiers (which you did above), performing this step and then putting it all back together.

Looks like that worked! I was able to filter to the duplicate gene identifiers, incorporate the code above, and put it all back together in the [last commit](https://github.com/AlexsLemonade/refinebio-examples/pull/359/commits/95a55ede9acf9e3b0214a5d28b6f79b9dccdbf2b). I tried to implement the code as efficiently as possible but of course it's still open to review and suggestions. Thank you!

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,428 @@
...
+Now let's combine the rest of the mapped data!
+
+```{r}
+filtered_mapped_df <- mapped_df %>%
+  dplyr::filter(
+    # First let's get the data associated with the gene identifiers that aren't duplicated
+    !entrez_id %in% dup_entrez_ids |
+      # Now let's get the data associated with the duplicated gene identifiers that we
+      # collapsed in `max_dup_df`
+      entrez_id %in% max_dup_df$entrez_id &
```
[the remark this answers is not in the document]
cansavvy: As an example, I think this means you really are going to keep things based on what `max(c("ENSG00000107018", "ENSG00000107014"))` returns, which will just be based on character values of the id and not what you determined in `max_dup_df`.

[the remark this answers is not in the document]
cansavvy: I think you will want to keep the first part of your filter (remove duplicated IDs) and then bind on `max_dup_df`. May want to do some kind of arrange after that so the dups aren't all at the end or beginning.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,428 @@
...
+        Ensembl %in% max_dup_df$Ensembl
+  )
+```
+
+Let's check to see that we removed the duplicate Entrez IDs from the rest of the mapped data.
+
+```{r}
+filtered_mapped_df %>%
+  # Let's sort by the Entrez IDs we know were duplicated
+  dplyr::arrange(dplyr::desc(entrez_id %in% dup_entrez_ids)) %>%
```
cansavvy, line 388: If we are just doing a preview, I think this can be a lot simpler: 
```suggestion
  dplyr::filter(entrez_id %in% dup_entrez_ids) %>%
```

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,428 @@
...
+
+Great, looks like we were able to successfully get rid of the duplicate gene identifiers while joining the rest of the mapped data!
+
+As mentioned earlier, we need a matrix for GSVA.
+Let's now convert our data frame into a matrix and prepare our object for GSVA.
+
+```{r}
+filtered_mapped_matrix <- filtered_mapped_df %>%
+  # First let's remove the column we no longer need
+  dplyr::select(-c("Ensembl")) %>%
```
cansavvy, line 401: I don't *think* you need this `c()`? 
```suggestion
  dplyr::select(-"Ensembl") %>%
```

cansavvy (commented):
@cbethell ! I think this looks good. I just have a few points where we need to do a tad bit of polishing/revision. 

1) I think the instruction around the "keep max values per sample" bit can be more clear. It can pretty easily be confused with "let's keep the row (in this case ENSG) with the max value". 
2) Part of the confusion about 1, I think is related to, or a symptom of the filtering step you have right now, which doesn't seem to fit with what I understand [Jackie's comment](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264)  The first part, where you make `max_dup_df` tracks with that idea, but the `filtered_mapped_df` creation part is confusing and I think not quite what we want. 
3) My last very minor comment has to do with `rownames` vs `row names` vs `rownames()`. This is something we've addressed in parallel issues in CONTRIBUTING.md but not for this instance specifically but I'm not sure if maybe we've missed this before? ¯\\_(ツ)_/¯

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,428 @@
...
+Now let's take a look at the rows associated with the duplicated Entrez IDs.
+
+```{r}
+mapped_df %>%
+  dplyr::filter(entrez_id %in% dup_entrez_ids)
+```
+
+We can see that the associated values vary for each row.
+
+As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSVA steps later, so let's keep the Entrez IDs associated with the maximum expression value per sample.
```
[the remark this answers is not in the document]
cansavvy: We may also want to remind the user that these values are representing probe IDs ultimately, even though the Ensembl column shows transcript IDs.  This is a tad different philosophically from RNA-seq where we've read the transcript itself; here we really are just dealing with probe differences here.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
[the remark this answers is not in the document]
cansavvy: I think to minimize some confusion, you can get rid of the `Ensembl` column earlier. Basically once your values are mixed up in `max_dup_df` the `Ensembl` column doesn't mean anything to us anymore anyway.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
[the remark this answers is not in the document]
jaclyn-taroni: Yes, but it's a little bit more nuanced that that because we're using BrainArray with SCAN. Quoting the [`SCAN.UPC`](https://bioconductor.org/packages/release/bioc/vignettes/SCAN.UPC/inst/doc/SCAN.vignette.pdf) documentation:

> By default,SCANmaps Affymetrix probes to “probeset” identifiers provided by themanufacturer.   However,  these  mappings  may  be  outdated  and  include  problematicprobes (for example, those that may cross hybridize).  In addition, multiple probesetsmay be assigned to a single gene.  As an alternative, the BrainArray resource (seehttp://brainarray.mbni.med.umich.edu/Brainarray/Database/CustomCDF/genomic_curated_CDF.asp) provides regularly updated mappings that exclude problematic probes and mapto genes rather than probesets.  

The target audience for this probably doesn't need all that information, so you can use terms like Ensembl gene identifier (not transcript identifiers, those would start with `ENST`).

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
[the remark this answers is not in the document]
cbethell: Yes I see your point here. I could see how long it would take to run the steps on the whole mapped data frame rather than just on the data frame with the duplicated gene identifiers. That would be the ideal option here I believe (if not I'll try binding on `max_dup_df` again, it didn't work for me in development).

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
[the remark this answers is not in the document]
cansavvy: My use of the word transcript is confusing/vague/a bit wrong. I meant to differentiate it from rna-seq reads, but no, these are not transcript IDs.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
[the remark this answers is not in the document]
cansavvy: >  Run the steps on the whole mapped data frame rather than just on the data frame with the duplicated gene identifiers. That would be the ideal option here I believe. 

Can you explain why you think this is ideal? (What if a user has less memory and a much bigger dataset -- I think this idea may be unnecessarily memory using)

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: I changed this to `head()` now that we arrange by the `dup_entrez_ids` in the previous step.

cbethell:
@cansavvy I addressed all of your comments and attempted to clarify the context and instructions around the "keep max values per sample" steps based on the suggestions in your comments. Please let me know if anything is still unclear!

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -347,11 +347,14 @@ mapped_df %>%
 
 We can see that the associated values vary for each row.
 
-As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSVA steps later, so let's keep the Entrez IDs associated with the maximum expression value per sample.
-Note that using the maximum expression value to collapse duplicate gene identifiers is just one of the expression values that can be used here.
-See the [Gene Set Enrichment Analysis (GSEA) User guide](https://www.gsea-msigdb.org/gsea/doc/GSEAUserGuideFrame.html) for more information on other expression values that can be used [@GSEA-user-guide]. 
+As we mentioned earlier, we will want to collapse any duplicate gene identifiers in preparation for the GSVA steps later.
+GSVA is executed on a per sample basis so let's keep the maximum expression value per sample associated with the duplicate gene identifiers.
+Note that using the maximum expression value to collapse duplicate gene identifiers means preferentially choosing a value or group of a higher rank downstream.
```
cansavvy, line 352: I'm not sure what this sentence means. Can explain what you mean here?

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -347,11 +347,14 @@ mapped_df %>%
...
+However, this is just one of the expression values that can be used to collapse duplicate gene identifiers.
```
cansavvy, line 353: What is one of the values?

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -347,11 +347,14 @@ mapped_df %>%
 
 We can see that the associated values vary for each row.
 
-As we mentioned earlier, we will want to remove duplicated gene identifiers in preparation for the GSVA steps later, so let's keep the Entrez IDs associated with the maximum expression value per sample.
-Note that using the maximum expression value to collapse duplicate gene identifiers is just one of the expression values that can be used here.
-See the [Gene Set Enrichment Analysis (GSEA) User guide](https://www.gsea-msigdb.org/gsea/doc/GSEAUserGuideFrame.html) for more information on other expression values that can be used [@GSEA-user-guide]. 
+As we mentioned earlier, we will want to collapse any duplicate gene identifiers in preparation for the GSVA steps later.
+GSVA is executed on a per sample basis so let's keep the maximum expression value per sample associated with the duplicate gene identifiers.
```
cansavvy, line 351:
```suggestion
GSVA is executed on a per sample basis so let's keep the maximum expression value per sample associated with the duplicate Ensembl gene identifiers.
```

When there's multiple gene id types, it can help to remind us which one is the duplicate here that we are talking about.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -364,48 +367,42 @@ max_dup_df <- mapped_df %>%
...
-      # collapsed in `max_dup_df`
-      entrez_id %in% max_dup_df$entrez_id &
-        Ensembl %in% max_dup_df$Ensembl
-  )
+  # First let's get the data associated with the gene identifiers that aren't duplicated
+  dplyr::filter(!entrez_id %in% dup_entrez_ids) %>%
+  # Now let's bind the rows of the data we collapsed in `max_dup_df`
+  dplyr::bind_rows(max_dup_df) %>%
+  # Arrange so that the Entrez IDs that were duplicated are at the top
+  dplyr::arrange(dplyr::desc(entrez_id %in% dup_entrez_ids)) %>%
```
cansavvy, line 381: Is there a reason we want the duplicates on the top? I think mixing in is fine? Or leaving as is?

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -364,48 +367,42 @@ max_dup_df <- mapped_df %>%
...
-Let's check to see that we removed the duplicate Entrez IDs from the rest of the mapped data.
+Let's preview `filtered_mapped_df` to check that we combined the collapsed duplicate Entrez IDs data with the rest of the mapped data.
 
 ```{r}
-filtered_mapped_df %>%
-  # Let's sort by the Entrez IDs we know were duplicated
-  dplyr::arrange(dplyr::desc(entrez_id %in% dup_entrez_ids)) %>%
-  # Now lt's display the first six observations of our new object
-  head()
+head(filtered_mapped_df)
```
cansavvy, line 389: If you are arranging early so the dups are at the top, we can leave the original `filtered_mapped_df`alone, but then here you can use `filtered_mapped_df %>% dplyr::filter(entrez_id %in% dup_entrez_ids)` to get the same effect.

cansavvy (commented):
This is getting there! 😅 I think your `filtered_mapped_df` is more like what we want now 👍 

But I think the explanation of the dup handling may need a bit of tweaking. What if you walk through one of the samples' max value selection to explain? I think this is a thing that will be easier to communicate from just seeing the example being done. 

So perhaps you can print out `mapped_df %>% dplyr::filter(entrez_id %in% dup_entrez_ids)` and `max_dups_df` side by side and explain an example in one of the samples about which values will be kept in the final product? 

I'm also not sure if the verb `collapse` is what we want here. I don't know if it tells us much about what's happening and I think about collapsing values in the sense of taking numbers and combining them to make a new number using some sort of math. Here we are choosing one number over another but not doing any calculation per se.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: The reason behind this was in preparation of the next step, where we can use the `head()` function to show that we retained the collapsed duplicate data as well as the rest of the mapped data.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: Gotcha 👍

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: This was based on [this comment from the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524299264), from which I gathered that choosing the max expression value here would result in choosing the values of higher ranking in the downstream GSVA results. I can re-word this sentence for clarification.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: Actually, I believe the Entrez ID is the duplicate here, multiple Ensembl gene identifiers mapped to the same Entrez ID.

cansavvy: Ah right. I got confused too. We’ll all the more reason to specify.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cansavvy: I think “collapsing” identifiers in general is a phrase that makes sense, but I think once we’ve stated the idea, we can be more specific about what we are doing. You use the phrase “collapse identifiers” quite a few times back to back so I’m trying to think if there more specific, clear descriptions we can replace some of those repeats with.

cansavvy: This could also be a thing that I’m being weird about and the next reviewer might have a different idea or thoughts about it. 🤷‍♀️

cbethell:
> But I think the explanation of the dup handling may need a bit of tweaking. What if you walk through one of the samples' max value selection to explain? I think this is a thing that will be easier to communicate from just seeing the example being done.
> 
> I'm also not sure if the verb `collapse` is what we want here. I don't know if it tells us much about what's happening and I think about collapsing values in the sense of taking numbers and combining them to make a new number using some sort of math. Here we are choosing one number over another but not doing any calculation per se.

Per your above comments @cansavvy, I substituted the word `collapse` and tried to clarify some of the context around keeping the max values per sample in the [last set of changes](https://github.com/AlexsLemonade/refinebio-examples/pull/359/commits/12f54879d795215839481b8bc375adbf952e7784).

I did not walk through one of the samples' max value selection to explain just yet because this notebook is already pretty long and with a 2nd PR to come, I believe we would want to make sure that adding lines is completely necessary before doing so. With that being said, if the wording I've included in the latest updates is still unclear, please let me know and we can probably go ahead and discuss the most flexible way to select a single sample (that would work for any dataset) in order to walk through its max value selection.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: I add an example in here.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: Let me know what you think of the latest changes in that case.

cansavvy:
> I did not walk through one of the samples' max value selection to explain just yet because this notebook is already pretty long and with a 2nd PR to come, I believe we would want to make sure that adding lines is completely necessary before doing so.

I *think* (hard to say for sure without trying it out) that you should be able to add this explanation and trim down some of the other wording and it *shouldn’t* add too many more lines. Plus I think being clear should be a high priority. So even if it’s slightly longer but we’ve made it super obvious what’s happening I think that’s okay.

> With that being said, if the wording I've included in the latest updates is still unclear, please let me know and we can probably go ahead and discuss the most flexible way to select a single sample (that would work for any dataset) in order to walk through its max value selection.

I’ll review this tomorrow and if needed, I’ll try giving you better/more specific suggestions. My brain is slowed down for the day.

cansavvy (commented):
I should officially post here that @cbethell and I did some pair-coding of sorts and now I think this is ready for someone else to look it over.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+This example is one of pathway analysis module set, we recommend looking at the [pathway analysis introduction](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_00_intro.html) to help you determine which pathway analysis method is best suited for your purposes.
+
+In this example we will cover a method called Gene Set Variation Analysis (GSVA) to calculate gene set or pathway scores on a per-sample basis.
+Gene set variation analysis (GSVA) constitutes a starting point to build pathway-centric models of biology.
+Rather than contextualizing some results you have received from another analysis like DGE, GSVA is designed to provide an estimate of pathway variation for each of the samples in an experiment.
+This means that GSVA scores will depend on the samples included in the dataset when you run GSVA; if you added more samples and reran GSVA, you would expect the scores to change [@Hanzelmann2013].
+
+GSVA determines the relative enrichment of gene sets across samples using
+a non-parametric approach.
+GSVA transforms a gene by sample gene expression matrix into a gene set by sample pathway enrichment matrix [@Hanzelmann-github].
```
jaclyn-taroni, line 23:
```suggestion
GSVA determines the relative enrichment of gene sets across samples using a non-parametric approach.
GSVA transforms a gene by sample gene expression matrix into a gene set by sample pathway enrichment matrix [@Hanzelmann-github].
```

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+---
+
+# Purpose of this analysis
+
+This example is one of pathway analysis module set, we recommend looking at the [pathway analysis introduction](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_00_intro.html) to help you determine which pathway analysis method is best suited for your purposes.
+
+In this example we will cover a method called Gene Set Variation Analysis (GSVA) to calculate gene set or pathway scores on a per-sample basis.
+Gene set variation analysis (GSVA) constitutes a starting point to build pathway-centric models of biology.
+Rather than contextualizing some results you have received from another analysis like DGE, GSVA is designed to provide an estimate of pathway variation for each of the samples in an experiment.
+This means that GSVA scores will depend on the samples included in the dataset when you run GSVA; if you added more samples and reran GSVA, you would expect the scores to change [@Hanzelmann2013].
```
jaclyn-taroni, line 19:
```suggestion
This means that GSVA scores will depend on the samples included in the dataset when you run GSVA; if you added more samples and ran GSVA again, you would expect the scores to change [@Hanzelmann2013].
```

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+This chunk of code will read both TSV files and add them as data frames to your environment.
+
+We stored our file paths as objects named `metadata_file` and `data_file` in [this previous step](#check-out-our-file-structure).
+
+```{r}
+# Read in metadata TSV file
+metadata <- readr::read_tsv(metadata_file)
+
+# Read in data TSV file
+df <- readr::read_tsv(data_file) %>%
```
jaclyn-taroni, line 220: Can we make this more descriptive than `df`? This comment will end up propagating throughout.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+all.equal(colnames(df), metadata$geo_accession)
+
+# Bring back the "Gene" column in preparation for mapping
+df <- df %>%
+  tibble::rownames_to_column("Gene")
+```
+
+### Import Gene Sets
+
+The function that we will use to run GSVA wants the gene sets to be in a list, rather than a tidy data frame that we used with `clusterProfiler` (although it does accept other formats).
```
jaclyn-taroni, line 242: I think this wording might come from training materials. In this context, there's no guarantee that someone will have looked at the examples that use `clusterProfiler`, so I would recommend rephrasing this to not make reference to the other examples at all.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+  dplyr::filter(!is.na(entrez_id)) %>%
+  # Make an `Ensembl` column to store the row names
+  tibble::rownames_to_column("Ensembl") %>%
+  # Now let's join the rest of the expression data
+  dplyr::inner_join(df, by = c("Ensembl" = "Gene"))
+```
+
+This `1:many mapping between keys and columns` message means that some Ensembl gene identifiers map to multiple Entrez IDs.
+In this case, it's also possible that a Entrez ID will map to multiple Ensembl IDs.
+For the purpose of performing GSEA later in this notebook, we keep only the first mapped IDs.
```
jaclyn-taroni, line 312:
```suggestion
For the purpose of performing GSVA later in this notebook, we keep only the first mapped IDs.
```

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+
+However, this is just one method of handling duplicate gene identifiers.
+See the [Gene Set Enrichment Analysis (GSEA) User guide](https://www.gsea-msigdb.org/gsea/doc/GSEAUserGuideFrame.html) for more information on other expression values that can be used, like the median expression value for example [@GSEA-user-guide].
+
+Now, let's implement our choose the max value method for all samples and Entrez IDs using tidyverse functions. 
+
+```{r}
+max_dup_df <- mapped_df %>%
+  # We won't be using Ensembl IDs moving forward, so we will drop this column
+  dplyr::select(-"Ensembl") %>%
```
jaclyn-taroni, line 364: I don't think the quotes are necessary. It's also possible that the `group_by()` would have resulted in this column being dropped anyway, but I haven't tested this!

```suggestion
  dplyr::select(-Ensembl) %>%
```

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+We can see `GSM187153` now has the `-0.007355525` value for Entrez ID `6013` like expected. 
+
+Looks like we were able to successfully get rid of the duplicate Entrez gene identifiers!
+
+Now let's combine our de-duplicated data with the rest of the mapped data!
+
+```{r}
+filtered_mapped_df <- mapped_df %>%
+  # We won't be using Ensembl IDs moving forward, so we will drop this column
+  dplyr::select(-"Ensembl") %>%
```
jaclyn-taroni, line 384: Come to think of it, why not move this to the chunk that ends at line 307 so you don't have to do it twice? I guess a downside is that you wouldn't get to _show_ that different Ensembl IDs mapped to the same Entrez ID in lines 350-351 but you could _tell_.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -0,0 +1,422 @@
...
+In other words, we will keep only the maximum expression value found across the duplicate Entrez gene identifier instances for each sample or column in this case.
+
+Let's take a look at the rows associated with the duplicated Entrez IDs and see how this will play out. 
+
+```{r}
+mapped_df %>%
+  dplyr::filter(entrez_id %in% dup_entrez_ids)
+```
+
+As an example using the strategy we described, for `GSM187153`'s data in the first column, `-0.007355525` is larger than `-0.136553967` so moving forward, Entrez gene `6013` will have ``-0.007355525` value and the `-0.136553967` would be dropped from the dataset.
```
jaclyn-taroni, line 354:
```suggestion
As an example using the strategy we described, for `GSM187153`'s data in the first column, `-0.007355525` is larger than `-0.136553967` so moving forward, Entrez gene `6013` will have `-0.007355525` value and the `-0.136553967` would be dropped from the dataset. 
```

jaclyn-taroni (approved):
👍  I had some comments but nothing major such that I need to take another look.

PR #353, opened 2020-11-12
WIP: WGCNA second draft

cansavvy:
## Analysis Purpose

It's a spin off issue from #306 and the discussion on #346 and a second draft from #350. 

Now that WGCNA is an advanced topic we can expand some things.

## Pull Request Stage

This is a **Draft PR** - needs review of big concepts and outline

## Strategy

After the first draft #350 and we made some decisions and changes, this is another draft before we go to refined PRs. 

I've tried to incorporate the bigger and smaller changes mentioned by @jashapiro here: https://github.com/AlexsLemonade/refinebio-examples/pull/350#issuecomment-725630945
As well as add a section where we dive into what module is most differentially expressed between the treatment groups. 

I also added a bit more guidance about some of the arguments for `blockwiseModules()` step. 
 
## Concerns/Questions for reviewers:

The main changes that need to be reviewed (and are tagged as such) are `Run WGCNA!` section and beyond. The rest hasn't changed as much, but I do have questions about whether the `Determine parameters for WGCNA` should be added to.

- Do we still want another plot like a heatmap showing a summary or do we think this is fine?
- My sina plot of the most differentially expressed module eigengene is a nice sanity check, but do we want something else/more?
- There are clustering parameters that I'm not sure if its worth giving guidance about or even what guidance to give. Section 3 of this doc is the only place I can find more description about choosing these parameters but I still don't find it that helpful. https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/BranchCutting/Supplement.pdf
See what you think, do we let it go? 
- A lot of the WGCNA tutorials have this kind of graph along with the soft threshold picking graph. I didn't include it originally because [its unclear to me from their docs](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/FemaleLiver-02-networkConstr-auto.pdf) what more you are really gaining from it that you can't conclude more easily from the R^2 power graph. Let me know if you disagree. 
Screen Shot 2020-11-13 at 12 26 52 PM

### Content checks
* [ ] All `{{BLANKS}}` have been replaced with the correct content.
* [ ] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable)

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [ ] Comments and documentation are up to date.
[...]

PR #360, opened 2020-11-18
WGCNA Part 2: Running WGCNA

cansavvy:
## Analysis Purpose

This is part one of the WGCNA example covering the actually running of WGCNA with `blockwiseModules()`. 

## Pull Request Stage

It's draft PR was reviewed in #353 

## Strategy

This second PR hasn't changed too much from the draft. This PR covers the picking a `power` argument, using that argument to run WGCNA, and saving the initial object to RDS. 

The main thing that needs review here is parameter discussions. Where should more info be added/reduced? 

There are more parameters discussions regarding clustering that we definitely could include here. Problem is the tutorials don't give a ton of concrete advice on how to diagnosis if a parameter needs tweaking or how to tweak it, so I don't have much to tell the users. Because of that, I've largely stuck with the defaults that they claim should be fine for most purposes. 

Some of these clustering parameters I'm referring to include `cutHeight` and `deepSplit`.
Section 3 of [this doc](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/BranchCutting/Supplement.pdf) is the only place I can find more description about choosing these parameters but I still don't find it that helpful. Note this paper is referring to a function that `blockwiseModules()` uses called `DynamicTreeCut` but you can give arguments directly to `cutHeight` and `deepSplit`.

You may want to take a brief look at `?blockwiseModules()` to see the whole list of parameters to see if anything to you seems worth discussing (I believe the things I've highlighted are whats most important). 

I think my feeling about these other clustering parameters are if the results look wonky, tweaking them probably(?) won't do much to help, so honestly switching to another gene co-expression method is probably a better bet (or switching datasets when possible). 

### Content checks

### Formatting Checks

### Add datasets to S3

* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar). -- done in #358 
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds. -- done in #358


cansavvy:
What's the meaning of the Scale-free Topology model fitting with unsigned R^2 that the tutorial suggests: https://github.com/AlexsLemonade/refinebio-examples/pull/358#discussion_r526996006

This is something we need to figure out. `-sign(slope) * SFT.R.sq` Why is that the calculation for model fitting and what does it mean?

jashapiro:
> What's the meaning of the Scale-free Topology model fitting with unsigned R^2 that the tutorial suggests: [#358 (comment)](https://github.com/AlexsLemonade/refinebio-examples/pull/358#discussion_r526996006)
> 
> This is something we need to figure out. `-sign(slope) * SFT.R.sq` Why is that the calculation for model fitting and what does it mean?

I don't think we need to figure it out... I think it is fine to just say it is the measure they recommend.

But as to what it means? My vague sense is that the model they use for splitting the network assumes it is "scale free". Scale free networks tend to have nice mathematical properties, so they want to use one, if they can get at network that matches fairly closely to this "ideal". 

The main question I was having was why and how there could be a negative slope for this fit with a large R^2. If that never happens, why would I need to multiply by the sign of the slope rather than just plotting R^2.  As I said first though, I don't think we need to figure it out for this. Just say this is the recommended measure and leave it at that.

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -320,6 +323,115 @@ normalized_counts <- assay(dds_norm) %>%
...
+```
+
+Now, let's plot the model fitting by the `power` soft threshold so we can decide on a soft-threshold for power. 
+
+```{r}
+ggplot(sft_df, aes(x = Power, y = model_fit, label = Power)) +
+  # Plot the points
+  geom_point() +
+  # We'll put the Power labels slightly above the data points
+  geom_label(nudge_y = 0.1) +
```
jashapiro, line 355: Super minor, but since this is a pretty clean plot, we probably don't need the boxes that `geom_label()` adds, so maybe just `geom_text()`?

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -320,6 +323,115 @@ normalized_counts <- assay(dds_norm) %>%
...
+sft_df %>%
+  ggplot() +
+  geom_label(aes(x = Power, y = max.k.), label = sft_df$Power) +
+  # We can add more sensible labels for our axis
+  xlab("Soft Threshold (power)") +
+  ylab("Scale Free Topology Model Fit, signed R^2") +
+  ggtitle("Scale independence") +
+  # This adds some nicer aesthetics to our plot
+  theme_classic()
+```
```
jashapiro, line 389: I don't think we need this plot?

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -320,6 +323,115 @@ normalized_counts <- assay(dds_norm) %>%
...
+
+Using this plot we can decide on a power parameter. 
+WGCNA's authors recommend using a `power` that has an signed $R^2$ above `0.80`, otherwise they warn your results may be too noisy to be meaningful. 
+
+If you have multiple power values with signed $R^2$ above `0.80`, then picking the one at an inflection point, in other words where the $R^2$ values seem to have reached their saturation [@Zhang2005].
+You want to a `power` that gives you a big enough $R^2$ but is not excessively large.
+
+So using the plot above, going with a power soft-threshold of `16`!
+
+If you find you have all very low $R^2$ values this may be because there are too many genes with low values that are cluttering up the calculations.
```
jashapiro, line 376: By "low values" do you mean low expression values?

jashapiro (approved):
What is here looks good; I only had minor comments about plot formatting and a bit of text. Also one plot to remove, I think.

I do have one major comment on what is _not_ here though, which is that we will probably want to add an outlier check before we get to this point. While `blockwiseModules()` does drop samples with lots of NAs, it doesn't seem to do the kind of outlier removal discussed in https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/FemaleLiver-01-dataInput.pdf

Since that section would not be part of the code added here though, I think it make sense to merge this in and add that as a separate issue/PR.

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
cansavvy: Consider it gone!

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
cansavvy: Yes. Low expression. Will clarify.

cansavvy:
> to add an outlier check before we get to this point.

I've been working on that in part 3's branch.

Issue #295, opened 2020-10-14
[Discussion] Improving the PR review process for this project


cbethell:
I am filing this issue to record some improvements that can be made to the PR review process for this project as discussed in a meeting between @jaclyn-taroni and I. 

This issue was prompted based on open PR #246, which is a PR that has been open across multiple sprints and now has 100+ comments, which is not ideal for a 2nd reviewer's experience (who may also have high-level comments affecting the structure of what's in the PR which would not be an ideal for anyone involved).

The proposed solution is to follow a PR review structure like the following:

```
1. File a draft PR to get high-level review comments from members of the team, 
specifically but not limited to those who may not be so involved in the recent updates/additions
to the content in the project already.

An example of said review would include feedback on any conceptual or structural
aspect of the PR (the author may not update comments/documentation in this PR
but will provide adequate information for reviewers to follow along and understand
what the author is accomplishing/looking to accomplish within the PR).

Note: This draft PR can be closed once the feedback is incorporated and high-level
review is completed.

2. A new PR will then be filed, incorporating the feedback from the draft PR, to get a
more thorough review from two members of the team. This will be the nitty-gritty PR
that will ultimately be merged into master.

Note: The author may use their discretion here to decide whether or not the comments
on the second PR are becoming extensive and a new PR should be opened to address any
further feedback. In the case of this project, the author may then make their case on the PR
as to why it should be merged and a new PR opened, and upon approval from the reviewer, 
get the second (current) PR in a condition to merge and merge it into master.
Then, a subsequent PR can be opened to address the further suggested changes and continue
PR review.
```

This will hopefully decrease the amount of comments on the PR for an easier review experience, and also address higher-level comments early on in the process for a smoother experience overall.

We want to take a first attempt at this plan for one of the upcoming series of pathway analysis PRs (which will be after wrapping up the ortholog mapping PRs and some other `before-testing` issues), so feel free to leave comments on this issue below with any suggestions or concerns you may have!

jashapiro:
This sounds good! 

One other thought I had was that it may make sense to explicitly mark sections of text and or code as **DRAFT** or **OUTLINE** so that the reviewers can focus on one or a few sections at a time for review. With reference to #246 (just because it is there), this might have taken the form of getting the base code working in a first PR, but leaving the section on collapsing of multimapping for a separate PR. We kind of knew early on that was a section that was going to require discussion, but in retrospect we probably didn't need to delay the rest of the notebook while we mulled it over.

[...]

PR #361, opened 2020-11-18
Add words about Draft and Refined PRs to CONTRIBUTING.md

cansavvy:
### Purpose

#295 discusses the updated PR process we've been using for adding new analyses. 
Before we can close that issue, we should put that important info in the place it should be, in the CONTRIBUTING.md. 

Does this better reflect the process? 

### Issue addressed

#295


CONTRIBUTING.md
```diff
@@ -89,18 +91,36 @@ For development purposes, you can download all the datasets for the example note
...
+This process ensures that discussions around the big picture: conceptual decisions, what steps are included, which packages are used, are generally concluded before review moves on to the details and further polishing.
+Note that all the following steps describe PRs to `staging` branch only (see more [about the branch set up](#pull-requests)).
+
+### Draft PR: Big picture reviews  
 
 - On your new git branch, [set up the analysis file from the template](#setting-up-the-analysis-file).
-- Add a [link to the html file to `_navbar.html`](#add-new-analyses-to-the-navbar)
+- Get the basic steps for the analysis set up and create a draft PR for a big picture review (Not all descriptions need to be 100% word-smithed, but the general steps/outline should be reflected).
+- Try to highlight things that encapsulate main concepts, ready for review using a `**REVIEW**` tag.
+Or use a `**DRAFT**` tag to indicate it hasn't really been worked on much yet.
```
cbethell, line 106:
```suggestion
- Try to highlight things that encapsulate main concepts as ready for review using a `**REVIEW**` tag and/or use a `**DRAFT**` tag to indicate a section that hasn't really been worked on much yet.
```

I was a little confused by the wording/structure of these two sentences. What do you think of the above wording?

CONTRIBUTING.md
```diff
@@ -89,18 +91,36 @@ For development purposes, you can download all the datasets for the example note
...
+- After the general outline of the analysis has been agreed upon through a reviewing process, incorporate the major feedback from the draft PR process.
```
cbethell, line 107: What exactly do you mean here? From this sentence, I would think that I should incorporate the major feedback from the review on the actual draft PR and its contents, rather than incorporating the major feedback in subsequent PRs (which is fine but not necessary since it is the subsequent PRs that will be merged to staging).

```suggestion
- After the general outline of the analysis has been agreed upon through a reviewing process, incorporate the major feedback from the draft PR process in subsequent refined PRs.
```

Maybe adding some language like I did above would clarify this point a bit more (not sure we really need the word 'refined' here just yet).

CONTRIBUTING.md
```diff
@@ -89,18 +91,36 @@ For development purposes, you can download all the datasets for the example note
...
+- Keep the original draft PR open for easy reference.
+
+### Refined PRs: Detailed reviews  
+
+- Break up the steps of the analysis into the manageable review chunks on their own branches for detailed review (you may want to discuss what ths chunks should be on the Draft PR).
```
cbethell, line 112:
```suggestion
- Break up the steps of the analysis into manageable review chunks on their own branches for detailed review (you may want to discuss what the chunks should be on the Draft PR).
```

cbethell (approved):
LGTM @cansavvy!

Just left a few minor suggestions re some of the wording, but otherwise this looks good so I don't believe I'll need to look at this again (but feel free to mention me or re-request my review if you feel the need to)!

CONTRIBUTING.md
cansavvy: > incorporate the major feedback from the review on the actual draft PR

That's what I've been doing. I don't know(?) if it matters when it's incorporated as long as its incorporated before refined PRs.

cansavvy: But I'm all for trying out a more specific approach. If you have a strategy you recommend, that might be good to put here.

cbethell: > That's what I've been doing. I don't know(?) if it matters when it's incorporated as long as its incorporated before refined PRs.

I completely agree with you here. The main reason behind my original comment was to make sure the feedback from the draft PR review process is propagated to the appropriate subsequent PRs.
That being said, I am fine with keeping your original wording here and perhaps reiterating the incorporation of the feedback on the draft PR in subsequent PRs, in the `Refined PRs` section of this doc.

cansavvy: I'll try to switch the wording a bit to make it clear.

PR #225, opened 2020-09-22
Differential Expression for Microarray - more than 2 groups

cansavvy:
## Purpose:
Closes #219 

## Strategy
I used the 2 group DE as a template, but switched to the Robinson et al 2012 Medulloblastoma data.
Still using limma, there's two main chunks that are added as opposed to 2 group DE microarray: 
- A series of steps to remove the outlier and the too small, U group. 
- The contrast_matrix steps that are needed for when you have many groups. 

Additionally, I kept the EnhancedVolcano, but I needed to select just one of the group's data to plot. 

The output of this will be used in #206 pathway analysis. (partially why I picked this dataset). 

## Concerns/Questions for reviewers:
- I looked through documentation for topTable to make sure the contrasts.fit toptable() still is reporting `log2FC` I think it is? Can you also look at the docs and see if you agree? 
 
### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable)

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [ ] Spell checked any Rmd file or md file.
* [x] Comments and documentation are up to date.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar).
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds.
* [x] In the Docker container, [snakemake was run for rendering](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#how-to-re-render-the-notebooks).

PR #362, opened 2020-11-20
Pr 2 of 2: Add Microarray Pathway Analysis - GSVA example

### LINKED ISSUE #343: New Analysis Example: Microarray Pathway Analysis - GSVA

cbethell:
## Analysis Purpose

This PR closes #343.

## Pull Request Stage

The relevant draft PR is #352.

## Strategy

This PR iterates on open PR #359 by adding the GSVA and visualization steps to the microarray GSVA example analysis.

I copied over the steps from draft PR #352, incorporating the relevant feedback from there.

Therefore, this PR includes:

- The implementation of `gsva()` -- [per this comment](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524328942) on the relevant draft PR, the parameters `min.sz = 5` and `max.sz = 50` were changed to `min.sz = 15` and `max.sz = 500` in this PR. The reason being that these values are pretty low for this analysis (and the lower values were implemented for development purposes, not intended for the user-facing version of the notebook).

- Visualization of the GSVA scores -- [per this comment](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524295683), the heatmap was removed from the visualization section. Per [this comment](https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524324537), the violin plot was adjusted to look at differences between groups in the dataset (using the metadata to derive said groups), as opposed to differences between gene sets like the violin plot in the draft PR.

More specifically, based on the goal of the [associated paper](https://pubmed.ncbi.nlm.nih.gov/18565887/), which is to "identify molecular profiles specific for treatment resistance to the current standard of care of concomitant chemoradiotherapy with the alkylating agent temozolomide", the violin plot now displays the GSVA score differences between a combination of the `refinebio_disease` and `refinebio_treatment` groups. This was done by adding a column to the annotation data frame to combine and store the information from `refinebio_disease` and `refinebio_treatment`.

The current violin plot looks like:
GSE7696_gsva_violin_plot

**What do we think of the current plot? Does this plot appear to make sense and better suit the purpose of this example notebook?** _Note that this PR is marked as a draft PR for the purpose of receiving feedback on the general concept/preparation of the violin plot before receiving more in depth feedback._

**Update**: After a discussion with Jackie, it was determined that the current violin plot (shown above) should change as it is showing the scores for a combination of pathways, which is mostly telling us that the scores are normally distributed. For the purpose of looking at the differences between disease status and treatment groups in our dataset (as stated earlier, this is associated with the goal of the paper), we should pick a pathway that's significantly different between groups and plot that.
The next steps therefore are to:
1. Join the selected columns from the metadata with the GSVA scores
2. Test every single pathway for significance and correct for multiple hypothesis testing (using https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/gene-set-enrichment-analysis as an example for this step)
3. Determine the pathway with the highest significance and plot that pathway to look at the differences between the disease status and treatment groups

**Note that PR 1 of 2 ends at line 386 (at the end of the `Handling duplicate gene identifiers` section) so review for this PR should start after that (at around line 388).** (this will be updated once #359 is merged)

Besides the physical appearance of the violin plot, does the interpretation of the plot and the context/instructions leading up to the plot make sense? How about the preparation of the data frame for plotting?

Overall, are the instructions or comments in this PR clear and suffice?

### Content checks

### Formatting Checks

### Add datasets to S3


cbethell:
> **Update**: After a discussion with Jackie, it was determined that the current violin plot (shown above) should change as it is showing the scores for a combination of pathways, which is mostly telling us that the scores are normally distributed. For the purpose of looking at the differences between disease status and treatment groups in our dataset (as stated earlier, this is associated with the goal of the paper), we should pick a pathway that's significantly different between groups and plot that.
The next steps therefore are to:
> 1. Join the selected columns from the metadata with the GSVA scores
> 2. Test every single pathway for significance and correct for multiple hypothesis testing (using https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/gene-set-enrichment-analysis as an example for this step)
> 3. Determine the pathway with the highest significance and plot that pathway to look at the differences between the disease status and treatment groups

Slight change of plans in the [latest commit](https://github.com/AlexsLemonade/refinebio-examples/pull/362/commits/60d5cbda658057a2e206bb4b5668abb546b1e348) -- I had some obstacles wrangling the steps at https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/gene-set-enrichment-analysis for this example, so in the interest of time, I used the [GSVA docs](https://www.bioconductor.org/packages/release/bioc/vignettes/GSVA/inst/doc/GSVA.pdf) and what's in open the WGCNA Part 3 open PR #363 as examples for using `limma` for this purpose. Please let me know if you have any thoughts on whether or not this is a good decision in this use case and/or suggestions for using another method like the one at https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/gene-set-enrichment-analysis. cc:@jaclyn-taroni

That being said, I am not sure that the current plot is ideally what we want here, so we may want to go with another method..
GSE7696_gsva_violin_plot

jaclyn-taroni:
Is there only one sample each in the `recurrent gbm-radiotherapy` and `re-recurrent gbm-tmz/radiotherapy` group? If so, I'm not surprised you're having trouble with what is in that OpenPBTA module.

cbethell:
> recurrent gbm-radiotherapy

You are correct!

jaclyn-taroni:
Okay, then you are not going to be able to make these comparisons as is... we'll probably need to take a different approach with `limma`. I might actually recommend picking a different dataset that you can do a straightforward two group comparison (or even like 3+ groups where there are no interactions). The reason why I am making that recommendation is because the point of this example is to communicate what you can do with GSVA scores, not to dig into how to use `limma` for different experimental designs. (This is not an advanced usage example after all.)

jaclyn-taroni:
@cansavvy is going to take a look at switching in `GSE37382`, generally following the steps [in this OpenPBTA module](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/0a6e54370399e7df632059957186a631b381d23b/analyses/gene-set-enrichment-analysis) for testing for significance, and plotting a single significant pathway (violin, boxplot, sina, pick your own adventure) this afternoon. If this dataset does not work, she will report back!

cansavvy:
I think Northcott seems to work well. Here's the plot I made from the most significant pathway: 
GSE37382_gsva_HALLMARK_UNFOLDED_PROTEIN_RESPONSE_sina_plot

Second note is that I did not follow the steps in OpenPBTA for significance because they ended up a using a custom function that was rather long. In another example I already went down the rabbit hole of trying to run a ANOVA tukey in a "short" way and had determined limma is just much quicker to show and explain. 

Plus, we already had the [limma for multiple subgroups written elsewhere](https://github.com/AlexsLemonade/refinebio-examples/blob/staging/02-microarray/differential-expression_microarray_02_several-groups.Rmd) so it required only a few changes (it's a different subgroup dataset (Robinson 2012) so I just need to change a few things with the `makeContrasts()` and column name clean up bit.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+```{r}
+# Let's explore what the output of gsva() looks like
+head(gsva_results[, 1:10])
+```
+
+## Find the pathway with the most significant differences across treatment groups
+
+If we want to know which pathway is most differentially expressed across subgroups, we can pretty easily use GSVA scores with differential expression steps using the limma package. 
+
+### Create the design matrix
```
cansavvy, line 468: For reviewing purposes, I'll note that most of this is a fairly direct grab from https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_02_several-groups.html#44_Create_the_design_matrix and has been reviewed before. 

Only difference is the subgroups are slightly different.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -376,20 +394,21 @@ max_dup_df <- mapped_df %>%
...
-
+We can see `GSM917111` now has the `0.2294387` value for Entrez ID `6013` like expected. 
 Looks like we were able to successfully get rid of the duplicate Entrez gene identifiers!
 
 Now let's combine our de-duplicated data with the rest of the mapped data!
 
 ```{r}
 filtered_mapped_df <- mapped_df %>%
+  # We won't be using Ensembl IDs moving forward, so we will drop this column
+  dplyr::select(-Ensembl) %>%
```
cansavvy, line 405: You will see, I brought this back. Because we show this in a preview I don't want users think that this whichever Ensembl ID actually makes it into the data frame has anything to do with anything. 

I realize this means we drop it twice, but I think it makes the data frame less confusing for the preview.

cansavvy:
This is ready for a review whenever @jaclyn-taroni.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+
+```{r}
+head(stats_df)
+```
+
+For each gene, each group's fold change in expression, compared to the average of the other groups is reported.
+
+By default, results are ordered from largest `F` value to the smallest, which means your most differentially expressed genes across all groups should be toward the top.
+
+This means `HALLMARK_UNFOLDED_PROTEIN_RESPONSE` appears to be the pathway that contains the most significant distribution of scores across subgroups.
```
jaclyn-taroni, line 564: Just wanted to note that I did not review this section in detail because there were two reviewers for #225.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+
+```{r}
+gsva_results %>%
+  as.data.frame() %>%
+  tibble::rownames_to_column("pathway") %>%
+  readr::write_tsv(file.path(
+    results_dir,
+    "GSE37382_gsva_results.tsv"
+  ))
+```
```
jaclyn-taroni, line 637: Wondering if we want to save the `topTable()` output as well? I don't have strong feelings about that. Or maybe I don't like that idea at all 🤔

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+Let's make a plot for our most significant pathway, `HALLMARK_UNFOLDED_PROTEIN_RESPONSE`.
+
+### Sina plot
+
+First we need to get our GSVA scores for this pathway in an appropriate format for `ggplot2`.
+
+```{r}
+annotated_results_df <-
+  # Pull out only the results for this pathway
+  gsva_results["HALLMARK_UNFOLDED_PROTEIN_RESPONSE", ] %>%
```
jaclyn-taroni, line 577: I think a _more general_ approach where we are getting the `gsva_results` matrix (an assumption, could be a wide data frame) into a long format could be welcome here. I suspect we don't do a lot of wide -> long in these examples and it is something that comes up pretty often when you go to plot these types of data. This has come up before on the draft PR https://github.com/AlexsLemonade/refinebio-examples/pull/352#discussion_r524295683. (To be fair, I'm the one that brought it up each time.)

I anticipate that we'd want to break this up into two or three steps instead of doing it "all at once":

- Make long format
- Maybe filter to relevant pathway 
- Join metadata

(Filter before join makes sense to me generally but depends on how we're going to write about it and what we want to show!)

components/dictionary.txt
```diff
@@ -1,6 +1,7 @@
 actin
```
jaclyn-taroni, line 1: Some of these can probably come out now that we've switched the dataset but not a big deal.

components/references.bib
```diff
@@ -539,6 +539,19 @@ @article{McInnes2018-how-umap-works
   url    = {https://umap-learn.readthedocs.io/en/latest/how_umap_works.html#}
 }
 
+@article{Meng2019,
```
jaclyn-taroni, line 542: Surprised to see this added here? I _believe_ this is the paper that's about meta-analysis with QuSAGE..

jaclyn-taroni (commented):
This looks pretty good! As I state in an individual comment, I didn't dig into the `limma` parts because there were two reviewers for #225. I had a few other comments I think we should address before we merge. @cansavvy and @cbethell - I'd recommend touching base about who would like to wrap up this PR before you get started making any edits!

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+  # Maximum gene set size
+  max.sz = 500,
+  # Compute Gaussian-distributed scores
+  mx.diff = TRUE,
+  # Don't print out the progress bar
+  verbose = FALSE
+)
+```
+
+Note that the `gsva()` function documentation says we can use `kcdf = "Gaussian"` if we have expression values that are continuous such as log-CPMs, log-RPKMs or log-TPMs, but we would use `kcdf = "Poisson"` on integer counts.
```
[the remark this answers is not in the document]
cbethell: Gotcha, took it out of this example 👍.

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +424,222 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+See `?gsva` for more options.
+
+```{r}
+# Let's explore what the output of gsva() looks like
+head(gsva_results[, 1:10])
+```
+
+## Find the pathway with the most significant differences across treatment groups
+
+If we want to know which pathway is most differentially expressed across subgroups, we can pretty easily use GSVA scores with differential expression steps using the limma package.
```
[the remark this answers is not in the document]
cbethell: I added a general statement after this one re `mx.diff = TRUE` and what it means for the GSVA output in latest set of changes. Let me know what you think. Should it be more detailed?

components/references.bib
cbethell: Ah yes, when I looked up the package citation this was one of the sources that were presented (it was third on the list however). I removed this reference and updated it to use the actual R package reference 👍

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
cbethell: My thoughts on this were we probably want to stick to as close to GSVA as possible in this example, seeing as we have an existing differential expression example where the `topTable()` output is saved. Users in this particular situation will likely be more interested in the GSVA results, however, I can be convinced otherwise. 

I did not save the `topTable()` output in the last set of commits based on my thoughts above, but feel free to let me know if you feel more strongly about saving the output now.

cbethell:
@jaclyn-taroni, I addressed most of your comments above (see: [my comment here](https://github.com/AlexsLemonade/refinebio-examples/pull/362#discussion_r532802953) on one of your comments that I left unaddressed) so this is ready for another look! 👍

cbethell:
@jaclyn-taroni, _now_ I have addressed most of your comments above (see: [my comment here](https://github.com/AlexsLemonade/refinebio-examples/pull/362#discussion_r532802953) on one of your comments that I left unaddressed) so this is ready for another look! 👍

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +415,244 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+Let's explore what the output of gsva() looks like.
+
+```{r}
+head(gsva_results[, 1:10])
+```
+
+## Find differentially expressed pathways
+
+If we want to identify most differentially expressed pathways across subgroups, we can use functionality in the `limma` package to test the GSVA scores.
+Keep in mind that the `mx.diff = TRUE` argument that we supplied to the `gsva()` function in the previous section means the GSVA output scores should be normally distributed, which will be evident in our downstream analyses.
```
jaclyn-taroni, line 457:
```suggestion
This is one approach for working with GSVA scores.
The `mx.diff = TRUE` argument that we supplied to the `gsva()` function in the previous section means the GSVA output scores should be normally distributed, which allows downstream analyses that make this assumption about the distribution.
```
And I would cite the package _vignette_ here which comes out and says this nicely IMO!

02-microarray/pathway-analysis_microarray_03_gsva.Rmd
```diff
@@ -405,11 +415,244 @@ Note that if we had duplicate gene identifiers here, we would not be able to set
...
+
+This means `HALLMARK_UNFOLDED_PROTEIN_RESPONSE` appears to be the pathway that contains the most significant distribution of scores across subgroups. 
+
+## Visualizing Results
+
+Let's make a plot for our most significant pathway, `HALLMARK_UNFOLDED_PROTEIN_RESPONSE`.
+
+### Sina plot
+
+First we need to get our GSVA scores for this pathway in an appropriate format for `ggplot2`.
```
jaclyn-taroni, line 563: Can we "remind" folks what `gsva_results` looks like here by repeating what we have at line 451:


```r
head(gsva_results[, 1:10])
```

jaclyn-taroni: I would also maybe make a note about what you mean by "an appropriate format."

jaclyn-taroni (approved):
👍 looks good!  I had a few more comments that should be addressed, but no need for me to re-review.

PR #364, opened 2020-11-20
WGCNA Part 4: Warn about Outliers

cansavvy:
## Analysis Purpose

This is part four where I'm incorporating a outlier analysis/decision. 

## Pull Request Stage

<del>This is a **Refined PR** - needs review of details and polishing </del> Sort of, but actually this is a new concept/methods being added, so it could still use a big picture review first. 

The relevant draft PR for WGCNA did not include this add, but was reviewed in #353 

## Strategy

This  fourth (and I _think_) last PR covers a thing that should have been added originally, which is outlier identification and management. 

I'm largely mimicking the wgcna tutorial here: https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/FemaleLiver-01-dataInput.pdf

I don't love this method of looking at outliers. I feel like it's not very succinct and involves a lot of artful guessing. 
I'm also not sure that its helping the WGCNA turn out better. But I don't know how to know see if its "better" the nonsig heatmap doesn't necessarily look more normal. 

Take a look at the heatmaps. I think removing 7 supposed "outliers" is probably too much. I can always try again with a more lenient cutoff, or maybe this whole analysis strategy is just not great. ¯\\_(ツ)_/¯ Let me know what you think. 

### Content checks

### Formatting Checks

### Add datasets to S3

* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar). -- done in #358 
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds. -- done in #358


jashapiro:
I did some informal playing around with this, which I will summarize briefly here: The short version is that I don't think we should have to worry about outliers in this data set (at least not much).

In this notebook, you did the outlier analysis before normalization, but with RNAseq data, the standard distance measure is not likely to be very informative; we will want to transform the data first. So I moved the clustering steps to after `vst()` and looked at the tree: 

6ad8c661-3e7e-4cc1-b41a-252ca83c4a43

Nothing obviously outlying. 

For good measure, I made a PCA plot: 
b0c25572-a400-4961-9cc9-c7a9b4a1f496

Same story: nothing I would obviously worry about as an outlier. Some points clustered more than others, but nothing extreme. (Not having looked at the metadata, I'm inferring from all of these that there are technical replicates in this data set, but I doubt those are a problem.)

jashapiro:
Okay, I couldn't resist looking: definitely technical replicates: https://www.ncbi.nlm.nih.gov/sra?term=SRP133573 It appears there are generally 4 sequencing runs per sample, but they all come from the same library. Refine.bio does not collapse replicates like this, but I wonder if we should look into that? A question for another day, or maybe another issue.

cansavvy:
So your comments here make sense. 

1) Does this mean we want to drop outlier analysis completely and replace it with a warning about outliers? Or is it good to keep the null outlier analysis as a test?

2) Looking at the smaller picture for replicates, I probably change the DE test here to take those replicates into account. Seems dishonest not to. Probably at the beginning we should collapse replicates: https://www.rdocumentation.org/packages/DESeq2/versions/1.12.3/topics/collapseReplicates..
OR, if we don't want to deal with this now, I can switch datasets.

jashapiro:
I think my preference would be to comment about outliers but leave it as a reference to other docs (the WGCNA data cleaning doc, I would guess). I would make sure that comment is after transformation though; detecting outliers in count data probably shouldn't be done with clustering on the raw data.

If you do stick with this data set, I do think collapsing the replicates is worth doing. In my Friday night Slack comments, I made a note that this might actually be a topic worth its own page we could add after going live. I am envisioning a "common patterns and pitfalls" collection of example pages that might be a useful addition. I will be filing an issue on that later today, I expect.

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -335,6 +335,9 @@ For more information about these transformation methods, [see here](https://alex
 dds_norm <- vst(dds)
 ```
 
+At this point, if your data has any outliers, you should look into removing them as they can affect your WGCNA results.
```
jashapiro, line 338:
```suggestion
At this point, if your data set has any outlier samples, you should look into removing them as they can affect your WGCNA results. 
```

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -335,6 +335,9 @@ For more information about these transformation methods, [see here](https://alex
 dds_norm <- vst(dds)
 ```
 
+At this point, if your data has any outliers, you should look into removing them as they can affect your WGCNA results. 
+WGCNA's tutorial has [an example of exploring your data for outliers you can reference](https://horvath.genetics.ucla.edu/html/CoexpressionNetwork/Rpackages/WGCNA/Tutorials/FemaleLiver-01-dataInput.pdf).
+
```
jashapiro, line 340: Maybe worth adding a little final note here about why we are not doing this.
```suggestion
For this example data set, we will skip this step (there are no obvious outliers) and proceed.

```

jashapiro (approved):
A couple small edit suggestions. I know the document talks about outlier genes as well, but I expect those to be much less of a problem most of the time. 

Approving pending small edits.

PR #366, opened 2020-11-20
Fix WGCNA installation

jashapiro:
This PR is to fix the Docker build steps so WGCNA installs correctly. Mostly this just means changing the install order so WGCNA installs later, but I also added another try at raising errors more dramatically.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,9 +35,6 @@ RUN pip3 install \
 #### R packages
 ###############
 
-# We need to install impute so WGCNA will install in the next step
-RUN R -e "options(warn = 2); BiocManager::install('impute', update = FALSE)"
-
 # Commonly used R packages
 RUN Rscript -e  "install.packages( \
     c('cluster', \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -48,12 +45,11 @@ RUN Rscript -e  "install.packages( \
...
       'viridis', \
       'spelling', \
-      'styler', \
-      'WGCNA'))"
+      'styler'))"
 
 # Install bioconductor packages
 # org.Mm.eg.db and org.Dr.eg.db are required for gene mapping
-RUN R -e "options(warn = 2); BiocManager::install( \
+RUN Rscript -e "options(warn = 2); BiocManager::install( \
```
cansavvy, line 52: Does it not matter if impute is not installed before WGCNA? WGCNA needs it to install. That's why I put it there.

docker/Dockerfile
```diff
@@ -85,6 +82,9 @@ RUN Rscript -e  "install.packages( \
 # Need this package to make plots colorblind friendly
 RUN Rscript -e "remotes::install_github('clauswilke/colorblindr', ref = '1ac3d4d62dad047b68bb66c06cee927a4517d678', dependencies = TRUE)"
 
+# Install WGCNA
+RUN Rscript -e  "install.packages('WGCNA')"
```
cansavvy, line 86: Oh. Never mind. I see its down here now.

cansavvy (approved):
This seems good and like it should do the trick. 🚀  Thanks for doing the sleuthing and fixing! :shipit:

PR #377, opened 2020-11-24
Microarray ORA Restructure Instruction

cansavvy:
### Purpose

Some of the instruction in the ORA microarray example was either too long winded, too open ended or just didn't quite fit the "self-teach" tutorial that this is meant to be. 
We didn't have a ton of advice about "why/when to choose ORA". 

### Issue addressed

#306 - Do we feel this addresses all of #306 comments? 

Do we feel that these word swaps better fit what we are looking for based on #306? 

Here's the main changes: 
- I've added some tutorials for people who want to know more about hypergeometric tests but I deleted the long winded explanation of what hypergeometric tests are. 
- I've added in a few places explaining "why choose ORA"
- I've added why this dataset isn't exactly ideal for ORA, but instead would probably better fit GSEA but we're just using it as an example.


02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -111,9 +113,12 @@ If the concept of a "file path" is unfamiliar to you; we recommend taking a look
 
 # Using a different refine.bio dataset with this analysis?
 
-If you'd like to adapt an example analysis to use a different dataset from [refine.bio](https://www.refine.bio/), we recommend placing the files in the `data/` directory you created and changing the filenames and paths in the notebook to match these files (we've put comments to signify where you would need to change the code).
-We suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
-From here you can customize this analysis example to fit your own scientific questions and preferences.
+If you'd like to adapt an example analysis to use a different dataset from [refine.bio](https://www.refine.bio/), we recommend replacing the `dge_url` with a different file path to a gene list that you are interested in.
```
jashapiro, line 116: We should specify what this file will contain: it is not just a gene list in this case, but a file with all the genes, each with p values and logFC. Not just any file will work in this code example.

I might add something about substituting the actual gene and background lists, rather than this specific file, with its specific format.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -111,9 +113,12 @@ If the concept of a "file path" is unfamiliar to you; we recommend taking a look
...
-We suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
-From here you can customize this analysis example to fit your own scientific questions and preferences.
+If you'd like to adapt an example analysis to use a different dataset from [refine.bio](https://www.refine.bio/), we recommend replacing the `dge_url` with a different file path to a gene list that you are interested in. 
+
+If you have differential expression results or something with a gene-level ranking and a two-group comparison, we recommend considering [GSEA](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_02_gsea.html) for your pathway analysis questions. 
+
+As far as code, we still suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
+But, from here you can customize this ORA example to fit your own scientific questions and preferences.
```
jashapiro, line 121: Why did you change the wording here? I don't feel like we need to add the extra qualifications at this point: I'd go back to original wording: 
```suggestion
We suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
From here you can customize this analysis example to fit your own scientific questions and preferences.
```

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -111,9 +113,12 @@ If the concept of a "file path" is unfamiliar to you; we recommend taking a look
 
 # Using a different refine.bio dataset with this analysis?
 
-If you'd like to adapt an example analysis to use a different dataset from [refine.bio](https://www.refine.bio/), we recommend placing the files in the `data/` directory you created and changing the filenames and paths in the notebook to match these files (we've put comments to signify where you would need to change the code).
-We suggest saving plots and results to `plots/` and `results/` directories, respectively, as these are automatically created by the notebook.
-From here you can customize this analysis example to fit your own scientific questions and preferences.
+If you'd like to adapt an example analysis to use a different dataset from [refine.bio](https://www.refine.bio/), we recommend replacing the `dge_url` with a different file path to a gene list that you are interested in. 
+
+If you have differential expression results or something with a gene-level ranking and a two-group comparison, we recommend considering [GSEA](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_02_gsea.html) for your pathway analysis questions.
```
jashapiro, line 118: I think this sentence is likely to be missed here. I would move it to line 19, before you are getting into details about files and folders.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -170,11 +175,15 @@ library(magrittr)
 
 ## Import data
 
-We will read in the differential expression results we will download from online.
+For ORA, we only need a list of gene IDs as our input, so this example can work for any situations where you have gene list and want to know more about what biological pathways it shares genes with.
+
+For this example, we will read in the differential expression results we will download from online.
```
jashapiro, line 180:
```suggestion
For this example, we will read in results from a differential expression analysis that we have already performed.
Rather than reading from a local file, we will download the results table directly from a URL.
```

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -170,11 +175,15 @@ library(magrittr)
...
 These results are from a zebrafish microarray experiment we used for [differential expression analysis for two groups](https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_02_2-groups.html) using [`limma`](https://bioconductor.org/packages/release/bioc/html/limma.html) [@Ritchie2015].
 The table contains Ensembl gene IDs, log fold-changes for each group, and adjusted p-values (FDR in this case).
 We can identify differentially regulated genes by filtering these results and use this list as input to ORA.
 
+Generally, the dataset we are using here would be a good candidate for [GSEA](https://alexslemonade.github.io/refinebio-examples/02-microarray/pathway-analysis_microarray_02_gsea.html) since it has gene-level rankings, but here we are showing you how to use ORA since it can be useful in a number of other situations.
```
jashapiro, line 185: I don't feel like we need to belabor this? If we want to say it again, I would probably put it toward the end with conclusions, rather than at this stage where we are focused on getting the data arranged for analysis.

components/references.bib
```diff
@@ -639,6 +639,14 @@ @manual{Prabhakaran2016
   url    = {http://r-statistics.co/Complete-Ggplot2-Tutorial-Part1-With-R-Code.html}
 }
 
+@manual{Puthier2015,
+  title        = {Statistics for {B}ioinformatics - {P}racticals - {G}ene enrichment statistics},
+  author       = {{Denis Puthier and Jacques van Helden}},
```
jashapiro, line 644:
```suggestion
  author       = {Denis Puthier and Jacques {van Helden}},
```

jashapiro (commented):
I think the simplification parts are good. We are not going to do justice to the hypergeomtric test here.

As to the references to GSEA, my suggestions are mostly about moving those discussion points to places where they are more likely to be noticed, rather than in the middle of other points.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
cansavvy: Should I have it in both places?

02-microarray/pathway-analysis_microarray_01_ora.Rmd
cansavvy: I deleted this mention

02-microarray/pathway-analysis_microarray_01_ora.Rmd
cansavvy: Added some advice/caution about what needs to be changed.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
cansavvy: I left it in both places for now.

jashapiro: I don't think it actually makes much sense here. The file you are describing has gene level ranking, so this caution seems more likely to confuse than illuminate when put here.

cansavvy: I removed it in this instance but not the other one that I added in line 19.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -14,7 +14,11 @@ output:
 This example is one of pathway analysis module set, we recommend looking at the [pathway analysis table below](#how-to-choose-a-pathway-analysis) to help you determine which pathway analysis method is best suited for your purposes.
 
 This particular example analysis shows how you can use over-representation analysis (ORA) to determine if a set of genes (e.g., those differentially expressed using some cutoff) shares more or fewer genes with gene sets/pathways than we would expect at random.
```
jashapiro, line 16: Just noticed this little wording change that seems minor but matters to me.
```suggestion
This particular example analysis shows how you can use over-representation analysis (ORA) to determine if a set of genes (e.g., those differentially expressed using some cutoff) shares more or fewer genes with gene sets/pathways than we would expect by chance.
```

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -340,21 +347,21 @@ head(dge_annot_df)
 ## Over-representation Analysis (ORA)
 
 Over-representation testing using `clusterProfiler` is based on a hypergeometric test [@clusterProfiler-book].
```
jashapiro, line 349: Unless there is objection, I think it is worth noting here that the hypergeometric test is equivalent to a Fisher's exact test, which people may be more familiar with. 
```suggestion
Over-representation testing using `clusterProfiler` is based on a hypergeometric test (often referred to as Fisher's exact test) [@clusterProfiler-book].
```

jashapiro (approved):
I had a couple more little changes. One of which I feel bad about, but I think it might be more familiar to people, despite my aversions.

Issue #372, opened 2020-11-23
Create a template Rmd specifically for advanced topics/usage examples


jaclyn-taroni:
### Background

Context is #357 - please read the initial comment for that issue in detail!

### Problem

The advanced topics notebooks have a different audience and it stands to reason that those users will have a bit more experience with things like file paths and the different platforms. We can't assume that those users have gone through our other examples (e.g., we should still treat them as independent). We should try to strike a balance between making sure folks can get up and running with refine.bio data but give more information so they can understand whether an analysis is appropriate for their question.

### What potential "gotchas" do we know of?

I expect most gotchas will become apparent while we work our way through #357. Because we want to give users an idea about whether or not an advanced approach is right for them, I'm imagining a mostly blank section in the template that's about _why_ to use an analysis.

### What are the recommended next steps?

1. Add introductory material to our first advanced topic example, which is WGCNA (#357). 
2. Create a template Rmd specifically for our advanced topics examples.
3. Update CONTRIBUTING.md to reflect that there will be two templates to choose from.

PR #378, opened 2020-11-24
`Advanced Usage` introduction simplification

### LINKED ISSUE #357: [Discussion] How should instructions differ in `Advanced Topics` from the other material
### Background

Initially we had a discussion that sprung up on https://github.com/AlexsLemonade/refinebio-examples/pull/353#issuecomment-728277824 (the first draft for a first "advanced topic"). 
The question is how should the instruction in the "advanced topic" section differ from the "non-advanced" topic sections (microarray and rna-seq). 

### Questions for discussion

Before we write more Advanced topic examples we should have some discussions, post some thoughts about how this section's material may need to differ from the non-advanced topics. 

Some more specific questions (these can/should be broken into their own issues if we find that useful).

##### What material from our template example does or does not apply? 
Will a separate advanced topics template be helpful or will things be too varied for a template to really help? (We may not know this entirely until we try out the first few examples). 

##### Technology isn't a known 
We will not be able to assume a technology (rna-seq/microarray) so we may have to have tech-specific notes where applicable and definitely need to be clear when we are making tech specific moves. 

##### Less hand-holding steps?
We should probably be careful with this one, I would still want to ere on the side of giiving more info than less, but this is a very context specific question. My guess is there are probably _less_ times in these sections where we need to do a step like `head(df)` for example. I think prompts for users to "look at their data" can decrease. 

##### More interpretation? 
@jashapiro mentioned more "Why" and less "How". We should be able to trust users to do a tad more tailoring and troubleshooting at this point, but they may want more guidance about interpretation of results. 

Related to these questions: When something isn't in R, we should figure out a new format that is as parallel as possible #139 (consistency is good). 

### Is there a particular timeframe for this issue?

If people can post their thoughts/responses to this, this sprint and next sprint then it would be good to get this somewhat aired out before we venture into writing more Advanced topics examples.

jashapiro:
### Purpose

This PR is to try out some simplifiedsetup instructions for advanced use examples. The general idea is that if people are clicking on "advanced" then they might not need quite so detailed instructions for the introduction

### Issue addressed

Closes #357

The main question the reviewers should address is whether these instructions are sufficiently clear for someone who comes in with a bit more knowledge of the material; i.e. someone who has used refine.bio before to download data files and has done at least one R notebook analysis. 

Additionally, is there enough of an off-ramp for someone who comes in without the prior knowledge... is there enough detail about where to go for more detailed instructions?


cansavvy:
Would it make more sense to discuss these changes by creating a new template file specifically for advanced topics?

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -26,156 +26,79 @@ As with many clustering and network methods, there are some parameters that may
...
-
-Click the "Download Now" button on the right side of this screen.  
-
-<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/download-now.png" width=200>  
-
-Fill out the pop up window with your email and our Terms and Conditions:  
-
-<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/40e47f4d3f39283effbd9843a457168061be9680/template/screenshots/download-email.png" width=500>  
-
-We are going to use non-quantile normalized data for this analysis.
```
cansavvy, line 88: I'm worried about deleting this piece in particular. We need them to know whether or not they should download quantile normalized data and where to find that.

jashapiro: I do cover that on line 40, but it could be made more prominent.

cansavvy: I think screenshots help. I guess bigger question is just because someone's an "advanced topics" user, can we assume they know how to download data from refine.bio and know the refine.bio options more readily?

cansavvy: I guess I think even advanced topics users will appreciate screenshots. Though I do agree with cutting back on the file path hand holding.

jashapiro: That was my thought, or they could go to another example to get this information. 🤷🏼

jashapiro:
> Would it make more sense to discuss these changes by creating a new template file specifically for advanced topics?
> 
> I guess this is a good way to get the conversation started.

That is what #372 is for. We split up making the first example from a general template.

jashapiro:
Converting this to a draft pending incorporation of changes in #379

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -26,156 +26,82 @@ As with many clustering and network methods, there are some parameters that may
...
+If you need more detailed instructions about how to obtain the data files from refine.bio, please consult one of the earlier examples, such as this one about [clustering and heatmaps](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/clustering_rnaseq_01_heatmap.html).
 
-## Obtain the `.Rmd` file
 
-To run this example yourself, [download the `.Rmd` for this analysis by clicking this link](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential_expression_rnaseq_01_rnaseq.Rmd).
+## Directory structure and required files
 
-Clicking this link will most likely send this to your downloads folder on your computer. 
-Move this `.Rmd` file to where you would like this example and its files to be stored.
+To run this example yourself, [download the `.Rmd` for this analysis by clicking this link](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential_expression_rnaseq_01_rnaseq.Rmd) and moving the `.Rmd` file to your preferred analysis folder.
```
cansavvy, line 36:
```suggestion
To run this example yourself, [download the `.Rmd` for this analysis by clicking this link](https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential_expression_rnaseq_01_rnaseq.Rmd) and move the `.Rmd` file to your preferred analysis folder.
```

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
cansavvy: > go to another example to get this information. 🤷🏼

If we think they might do this, then we should probably just keep the screenshots here too. I think screenshots help decrease brain glucose usage for simple things.

04-advanced-topics/network-analysis_rnaseq_01_wgcna.Rmd
```diff
@@ -26,156 +26,82 @@ As with many clustering and network methods, there are some parameters that may
...
-if (!dir.exists("data")) {
-  dir.create("data")
-}
-
-# Define the file path to the plots directory
-plots_dir <- "plots" # Can replace with path to desired output plots directory
-
-# Create the plots folder if it doesn't exist
-if (!dir.exists(plots_dir)) {
-  dir.create(plots_dir)
```
cansavvy, line 60: We still will need to create these directories though. (Unless you've put this part somewhere else)

cansavvy (commented):
Couple other comments in the meantime. 

I still disagree with getting rid of the screenshots. I don't think "advanced users" will be upset by having more screenshots. But perhaps we can ask Deepa to weigh in on a discussion about what different kinds of users might expect or respond.

PR #381, opened 2020-11-30
WIP: ORA for RNA-seq (with WGCNA module genes)

cansavvy:
## Analysis Purpose

Add an "RNA-seq" version of ORA pathway analysis.

#344 

## Pull Request Stage

This is a **Draft PR** - needs review of big concepts and outline.

To help reviewers save time, I'm using slightly different tags to alert you to the few places that have changed in this module so far. 
`**IDENTICAL TO MICROARRAY EXAMPLE**` means that until a different tag shows up, all the info there is 99% the same as the microarray example. Any differences I put in parenthesis like `**IDENTICAL TO MICROARRAY EXAMPLE** (except microarray. --> rnaseq)`

`**REVIEW**` doesn't mean it's 100% polished, but it does mean that it differs from the microarray example. 

## Strategy

So far, not much has changed from the microarray example to the RNA-seq example besides using the WGCNA gene modules instead of a DGE results table. Note that for the purposes of this Draft PR, the descriptions are completely polished yet, but wanted to get some feedback on the larger steps and see if there's anything we want to change. 

## Concerns/Questions for reviewers:

What I want to know from this Draft PR is what do we want to change from microarray? Related to #223. 
- Do we want to do different plots, that seems to be the main thing we could play with, though for story purposes I do like what the clusterProfiler plots here illustrate. 
- For the background gene list should we use all the detected genes, or should we pull all possible genes from Ensembl as background genes -- this feels like too much, but let me know if you disagree. 

Another idea, if we are okay with what is here, and given the other `before going live` examples we want to get to and that steps here are fine, perhaps we leave these steps as is and file an issue about us coming back if we have inspiration about how to show something different with ORA later?

### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable) -- not set here. 

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [ ] Comments and documentation are up to date.
* [ ] All links have been checked and are properly formatted.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
[...]

PR #382, opened 2020-11-30
WGCNA: Switch plots to save as png (instead of PDF)

cansavvy:
### Purpose

I wasn't able to render the PDFs I saved in the ORA example (#381) so I needed to go back and make the heatmaps save to PNG instead. 

See this comment: https://github.com/AlexsLemonade/refinebio-examples/pull/381#discussion_r532788707


cbethell (approved):
LGTM 👍

PR #389, opened 2020-12-01
Publish internal changes (and footer)

cansavvy:
## Changes Being Published

I've separated out the internal changes from the original #383 (plus a PR that has basic changes to the footer since I wasn't sure where else that should fit). 

I got this branch together by checking out master, and cherry picking all these commits. 

## List of Commits/PR's Included

- Updates to CONTRIBUTING #326, #333
- Adding basic footer #307
- Edits and Adds to PR and issue templates: #319, #334, #341, #342, and #337

## Link to html Preview

For reviewing purposes, you can take a browse through htmlpreview: http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/gh-pages-stages/01-getting-started/getting-started.html

PR #392, opened 2020-12-01
Remove getting started zip file

cansavvy:
### Purpose

A zip file I used for previewing html changes snuck into staging at some point. Need it gone. 
See https://github.com/AlexsLemonade/refinebio-examples/pull/389#issuecomment-736692025


jaclyn-taroni (approved):
Get outta here zip file!

PR #394, opened 2020-12-01
ORA RNA-seq: Part 1 - The Set Up

cansavvy:
## Analysis Purpose

Add an "RNA-seq" version of ORA pathway analysis. This PR adds the "Set Up" part 2 will add the actual ORA and plot making. 

#344 

## Pull Request Stage

This is a **Refined PR** - needs review of big concepts and outline. Draft Review occurred here: #381 not much changed. 

To help reviewers save time, I'm still using slightly different tags to alert you to the few places that have changed in this module so far. 

After your initial review, I will make a commit that deletes all these tags. (Unless you don't want the tags in the first place, then I can just delete them).

`**IDENTICAL TO MICROARRAY EXAMPLE**` means that until a different tag shows up, all the info there is 99% the same as the microarray example. Any differences I put in parenthesis like `**IDENTICAL TO MICROARRAY EXAMPLE** (except microarray. --> rnaseq)`

`**REVIEW**` doesn't mean it's 100% polished, but it does mean that it differs from the microarray example. 

## Strategy

Main changes from the Microarray are: 
- We're using a WGCNA module gene list instead of DGE results. 
- Conversion now is not gene symbols but entrez IDs
- Not danio rerio -> human 

Most everything else is pretty identical to microarray ORA. 

## Concerns/Questions for reviewers:

I think wording and descriptions of the WGCNA analysis are the main focus here. 
Do they make sense but aren't going off the trail too much? 

### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable) -- not set here. 

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [x] Comments and documentation are up to date.
* [x] All links have been checked and are properly formatted.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar).
* [x] Any not yet added packages needed for this analysis have been added to the Dockerfile and it successfully builds.

Issue #289, opened 2020-10-14
Editing: Someone needs to look at the GitHub Pages version of all examples and file issues


jaclyn-taroni:
The HTML versions of examples that can be browsed via the GitHub Pages will be different from development (`html_document` vs. `html_notebook`). As a result, there may be somethings like printing out large data frames that do not work particularly well in that context. We should have someone go through the entire site and make notes.

jashapiro:
I'm putting this here, as it seems the most reasonable place, rather than as a new issue: 

We may want to consider making blocks with `library()` calls include `message = FALSE` to avoid chatty outputs cluttering the page. (DESeq2 👀 )

We could also do this on a global basis, with `knitr::opts_chunk$set(message = FALSE)` early on in the doc, but it probably makes sense to keep this on a case by case basis.

jashapiro:
One particular issue that came up as I started review with some of these thoughts in mind: how do we feel about the output of `readr::read_tsv()`? Do we want to suppress this in general?

Original comment:
https://github.com/AlexsLemonade/refinebio-examples/pull/409#issuecomment-738246954

jaclyn-taroni:
@jashapiro are we good to close this (and #290)? Last GSVA PR is imminent (and tracked via #371 and #349 arguably).

jashapiro:
I think we can close it. I'm sure we will catch things later, but individual issues seem to be the way to go now.

PR #395, opened 2020-12-01
ORA RNA-seq: Part 2 - Run ORA and get results!

cansavvy:
## Analysis Purpose

Add an "RNA-seq" version of ORA pathway analysis. This PR adds the the actual ORA actual ORA and plot making after part 1 added the basic set up and gene ID conversion. 

#344 

## Pull Request Stage

This is a **Refined PR** - needs review of big concepts and outline. Draft Review occurred here: #381 not much changed. 

To help reviewers save time, I'm still using slightly different tags to alert you to the few places that have changed in this module so far. 

After your initial review, I will make a commit that deletes all these tags. (Unless you don't want the tags in the first place, then I can just delete them).

`**IDENTICAL TO MICROARRAY EXAMPLE**` means that until a different tag shows up, all the info there is 99% the same as the microarray example. Any differences I put in parenthesis like `**IDENTICAL TO MICROARRAY EXAMPLE** (except microarray. --> rnaseq)`

`**REVIEW**` doesn't mean it's 100% polished, but it does mean that it differs from the microarray example. 

## Strategy

Main changes from the Microarray are: 
- We're using a WGCNA module gene list instead of DGE results. 
- Conversion now is not gene symbols but entrez IDs
- Not danio rerio -> human 

Most of the steps themselves pretty identical to microarray ORA, but the wording is what we need to focus on here. 

The wording about RNA-seq and this datasets' results are the main differences here that need to be reviewed. 

### Content checks
* [x] Seed is set (if applicable) -- not set here. 

### Formatting Checks

### Add datasets to S3


cansavvy:
If you don't want to wait for the checks on #394 to finish, everything in the `Over-representation Analysis (ORA)` section and beyong is what needs to be reviewed.

jashapiro:
> If you don't want to wait for the checks on #394 to finish, everything in the `Over-representation Analysis (ORA)` section and beyong is what needs to be reviewed.

I can do that, but might be a bit of delay. Meanwhile a neat trick: if you change the PR target to `cansavvy/part1-ora-rnaseq`, it will automatically retarget to `staging` when #394 goes in. And then I will only see the changes that are really part of this PR in the diff. Not critical, but a neat trick.

cansavvy:
> a neat trick: if you change the PR target to `cansavvy/part1-ora-rnaseq`, it will automatically retarget to `staging` when #394 goes in. And then I will only see the changes that are really part of this PR in the diff. Not critical, but a neat trick.

Wow. Yeah. I guess that makes sense! Well too late now. #394 is merged!

jashapiro (commented):
This looks good. I have a few very minor comments and clarifications, but nothing too dramatic.

cansavvy:
Used your wording/punctuation suggestions, I apparently forgot how to use apostrophe's yesterday. 😬  See if things look good to you.

jashapiro:
> Used your wording/punctuation suggestions, I apparently forgot how to use apostrophe's yesterday. 😬 See if things look good to you.

apostrophes?

03-rnaseq/pathway-analysis_rnaseq_01_ora.html
```diff
@@ -4007,7 +4007,7 @@ <h2><span class="header-section-number">4.2</span> Import data</h2>
 <span id="cb31-3"><a href="#cb31-3"></a><span class="co"># desired gene list results</span></span>
 <span id="cb31-4"><a href="#cb31-4"></a>gene_module_df &lt;-<span class="st"> </span>readr<span class="op">::</span><span class="kw">read_tsv</span>(gene_module_url)</span></code></pre></div>
 <pre><code>## 
-## ── Column specification ──────────────────────────────────────────────────────────────────────────────
+## ── Column specification ────────────────────────────────────────────────────────────────────────────────────────────────────
```
jashapiro: Commenting on this change because it reminded me of something that we should look at, which is whether we can make the text output width a bit narrower. When rendered, these lines often end up wrapped, which is not ideal. We could partially solve this by making sure output blocks do not wrap in the html, but a better solution would be to actually control the width, if possible.

jashapiro (approved):
Looks good.

03-rnaseq/pathway-analysis_rnaseq_01_ora.html
cansavvy: I think this is pertinent to #289

PR #398, opened 2020-12-02
Add message = FALSE to mute chatty blocks

### LINKED ISSUE #370: Check compatibility with magrittr 2.0
### Background
Magrittr 2.0 was just released, and while it overall seems great (if soon to be obsolete), it does some changes that _could_ be incompatible with current usage: https://www.tidyverse.org/blog/2020/11/magrittr-2-0-is-here/

### Problem

My guess is that we will be fine, but as users who do not use the docker image (most!) may end up with magrittr 2.0, we should check compatibility. The most likely place things might arise is in the timing of evaluation in pipes: magrittr2 is lazy, which can mean that steps that are done for side effects may be skipped. 

### What potential "gotchas" do we know of?

It's a docker issue, so who knows!

More generally, the "checking if things changed" is always a bit of a challenge, as there will almost certainly be minor changes that need sorting through when doing a diff between html versions.

### What are the recommended next steps?

I think the easiest next step is to make an image based on R 4.0.3 (or whatever the now-current version of R is) and check that our pages build with similar results. Using the current R will also result in using the current CRAN and hence will get magrittr2.

If there are any problems that arise, we should aim to fix them in the least disruptive way we can, which will likely be by  

Note, I don't think we should unfreeze the `refinebio-examples` docker image for this; we should stick with the current image of 4.0.2, at least for now. Instead, we should create an alternative dockerfile/image.

### Is there a particular timeframe for this issue?

Probably should be done before going live...

A slightly later issue might be to make this a general check we can run to see if we have code that won't work as R, CRAN, *& Bioconductor change and evolve.

jashapiro:
### Purpose

Some libraries, such as DEseq2, are quite chatty when loaded, and this adds a lot of visual clutter to our rendered pages, without adding much information for the user. This PR seeks to cut that down by adding `message=FALSE` to code blocks with `library()` calls. In our current template, there is one of these blocks per page. 

### Issue addressed

closes #370

I think this is overall pretty straightforward; the changes in .Rmd files are small, and I don't think I missed any. The rendered files change more, of course, but these seem to all be the expected changes.

I also did change the template file, so future pages should get this change by default.

Occasionally, there might be messages in this block that are useful to users, so it might be nice if they showed up when users run the notebook on their own. This could be done by selectively adding `message=FALSE` in preprocessing before rendering (as we do with the bibliography) to chunks with some kind of special tag, but I think that is well out of scope for now.


cansavvy (approved):
This seems good to me! Although now there's a conflict to resolve.

Issue #390, opened 2020-12-01
Publish microarray GSVA to master after intro is complete


cansavvy:
Originally this was on one issue -- #380. This is reminder to publish GSVA to master after #371 is complete. 

This will involve checking out master branch, making a new branch from that and cherry picking the GSVA relevant commits to that (as described in CONTRIBUTING.md).

PR #401, opened 2020-12-02
Add GSVA package to Dockerfile

cbethell:
### Purpose

The microarray GSVA example was merged into staging without updating the Dockerfile to install the `GSVA` package.

### Issue addressed

No particular Github issue is being addressed here however, this is related to/blocking #390 

No gotchas, tested that the Dockerfile builds locally 👍 

Reviewer(s) may also want to test that this builds locally (although Github actions should tell us if it does not, if I'm not mistaken).


jashapiro (approved):
Looks like it should work just fine! Assuming checks pass, good to go.

Issue #399, opened 2020-12-02
Control the width of output blocks


jashapiro:
Sometimes we end up rendering output chunks that are wider than ideal. We should look into whether we can change that, perhaps setting the width for rendering at 60-70 characters. This is too narrow for most usage, but is more generally compatible with display on an html page.

> Commenting on this change because it reminded me of something that we should look at, which is whether we can make the text output width a bit narrower. When rendered, these lines often end up wrapped, which is not ideal. We could partially solve this by making sure output blocks do not wrap in the html, but a better solution would be to actually control the width, if possible.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/refinebio-examples/pull/395#r534232468_

jashapiro:
Turns out, this is a bit uglier to deal with than we might like: 

The main solution is to set `options(width=60)` or another value, but this has to be done in the document, which, ugh. 

The main points are discussed here:
https://bookdown.org/yihui/rmarkdown-cookbook/text-width.html
from a tweet by Yihue Xie https://twitter.com/xieyihui/status/1334158473102307331?s=20

Some extra hacks for display are here: 
https://github.com/hadley/adv-r/blob/master/common.R
via Hadley Wickham: https://twitter.com/hadleywickham/status/1334163848753000449?s=20

jashapiro:
Closed (mostly?) by #402

There may be some blocks/functions that do not respect the `width` parameter but we may have to find those as part of #289

PR #402, opened 2020-12-02
Add rendering options 

jashapiro:
### Purpose

The original impetus for this change was #399, trying to add an option to adjust the width of the output in rendered chunks. However, it turns out there is no easy option to do this from the render call itself. The only way to accomplish this is to add a call to  `options()` within the notebook itself. 

I didn't want to modify all the notebooks, and I didn't want a random block confusing people, so I decided to take a similar approach we took for adding a bibliography. This PR then modifies the rendering script to add a chunk immediately *after* the header which sources a file provided by a command line option. This chunk is set with `include=FALSE` so that it does not show up in the rendered document, but its effects will.

The current version of the sourced file only includes the `options(width = 70)` call, but it can now be easily modified with more changes and hacks if we want to improve other particular rendering as may be required. See https://github.com/hadley/adv-r/blob/master/common.R for some inspiration from Hadley Wickham.

### Issue addressed

#399

Because the rendering script deletes the temp file, you may not see the effects of this script from a casual glance (the added chunk is not saved anywhere). If you want to see what is happening for testing, you can comment out the final line:
`file.remove(tmp_file)` so you can see what the script has actually done.

To ease review, I did not include the actual rerendered files yet, except for `/02-microarray/clustering_microarray_01_heatmap.html`, which you can verify does shorten the output length. 

Before merging, all of the other rerendered files should be added.

~~This option should probably also be documented in `CONTRIBUTING.md`~~ Done!


cansavvy (approved):
Seems like it works as expected. And the other changes seem fine too. 🚀

PR #405, opened 2020-12-03
PR 1 of 2: Add RNA-seq Pathway Analysis - GSEA example

cbethell:
## Analysis Purpose

This PR addresses issue #368.

## Pull Request Stage

The relevant draft PR is #397

## Strategy

I mostly copied and pasted the first half of the code from the draft PR (stopping right before the step implementing the `GSEA()` function) and refined the context/user instructions along the way.

I updated the pathway analysis introduction to be relevant to the RNA-seq modules.
The specific example introduction, although copied from the microarray, I _believe_ still applies to this GSEA example. However, please correct me if I am wrong! 

Per [this comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/397#discussion_r535257352), I included some context around the log2 fold change values we are using in this example analysis.

I also include a step to set the seed in this PR as I noticed the top 3 pathways would change with each run (although `seed = TRUE` is provided to the `GSEA()` function and this won't be evident until the upcoming PR 2 of 2).

Overall, the context/user instructions should receive an extra close look (especially around the log2 fold change values and the use of all gene sets rather than hallmark gene sets).

### Content checks

### Formatting Checks

### Add datasets to S3


jaclyn-taroni:
Having looked at exactly zero documentation or code, I would expect the argument to `seed` to be an integer?

cbethell:
> Having looked at exactly zero documentation or code, I would expect the argument to `seed` to be an integer?

Of course! That would explain why it would not be working as expected. The docs refer to the `seed` argument as a logical argument so I assumed that it would be looking for a TRUE or FALSE statement. I'll try supplying it with an integer instead!

jaclyn-taroni:
Worth a try, but I'm not particularly hopeful from the documentation!

cbethell:
> Worth a try, but I'm not particularly hopeful from the documentation!

Unfortunately that didn't work.

This is even more interesting because I've re-run `pathway-analysis_microarray_02_gsea.Rmd` and gotten the same results without explicitly adding the set seed step, but supplying the `seed = TRUE` argument to the `GSEA()` function. I'm not sure if using the hallmark gene sets in the microarray case plays a role here (I didn't think that it would).

cbethell:
> Having looked at exactly zero documentation or code, I would expect the argument to `seed` to be an integer?

I will note here that I did some googling and a user encountered the same issue here and also solved it by setting the seed before the function and supplying the argument `seed = TRUE` within the function. See reference [here](https://www.biostars.org/p/367774/).

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+
+```{r}
+# Read in the contents of your differential expression results file
+# `dge_url` can be replaced with a file path to a TSV file with your
+# desired gene list results
+dge_df <- readr::read_tsv(dge_url)
+```
+
+`read_tsv()` can read TSV files online and doesn't necessarily require you download the file first. 
+Let's take a look at what these contrast results from the differential expression analysis look like.
```
cansavvy, line 201: Are there multiple contrasts here? I think it's just the one comparison right?

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+The annotation package `org.Hs.eg.db` contains information for different identifiers [@Carlson2020-human].
+`org.Hs.eg.db` is specific to _Homo sapiens_ -- this is what the `Hs` in the package name is referencing.
+
+We can see what types of IDs are available to us in an annotation package with `keytypes()`.
+
+```{r}
+keytypes(org.Hs.eg.db)
+```
+
+Even though we'll use this package to convert from Ensembl gene IDs (`ENSEMBL`) to gene symbols (`SYMBOL`), we could just as easily use it to convert from an Ensembl transcript ID (`ENSEMBLTRANS`) to Entrez IDs (`ENTREZID`).
```
cansavvy, line 261: Could make this comment more general since there's nothing in particular about transcripts that might be useful to the user. 
```suggestion
Even though we'll use this package to convert from Ensembl gene IDs (`ENSEMBL`) to gene symbols (`SYMBOL`), we could just as easily use it to convert to and from any of these `keytypes()` listed above. 
```

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+
+Take a look at our other gene identifier conversion examples for examples with different species and gene ID types:
```
cansavvy, line 263: I think we can just have one spot that we refer people to the gene ID conversion examples and I think after we run the gene conversion example is better. So I'd just move this sentence and replace what's at 293. 

(I know this is copied over from how we have it written in microarray -- I don't feel strongly enough about this to make an issue, but it is something I want to switch if we do some edits too those notebooks later). So IDK.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+
+```{r}
+any(duplicated(dge_mapped_df$gene_symbol))
+```
+
+Looks like we do have duplicated gene symbols.
+Let's find out which gene symbols have been duplicated.
+
+```{r}
+dup_gene_symbols <- dge_mapped_df %>%
```
cansavvy, line 316: We could probably trim out some of these previews to be a bit more succinct. But I don't feel strongly about this either.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+  dplyr::distinct(gene_symbol, .keep_all = TRUE)
+```
+
+Note that the log2 fold change values we use here have been transformed by `DESeq2` to account for genes with low counts or highly variable counts.
+See the [`DESeq2` package vignette](https://bioconductor.org/packages/release/bioc/vignettes/DESeq2/inst/doc/DESeq2.html) for more information on how `DESeq2` handles the log2 fold change values with the `lfcShrink()` function.
+
+Let's check to see that we removed the duplicate gene symbols and kept the rows with the higher absolute value of the log2 fold change.
+
+```{r}
+filtered_dge_mapped_df %>%
```
cansavvy, line 355: I think this could be shortened to just `any(duplicated())` type deal. We probably don't need to see this data previewed again.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,408 @@
...
+GSEA relies on genes' rankings on the basis of this statistic.
+Retaining the instance of the gene symbol with the higher absolute value means that we will retain the value that is likely to be more highly- or lowly-ranked or, put another way, the values less likely to be towards the middle of the ranked gene list.
+We should keep this decision in mind when interpreting our results.
+For example, if the duplicate identifiers tended to be enriched in a particular gene set, we may get an optimistic view of how perturbed that gene set is by preferentially selecting values that have a higher absolute value.
+
+```{r}
+filtered_dge_mapped_df <- dge_mapped_df %>%
+  # Sort so that the highest absolute values of the log2 fold change are at the top
+  dplyr::arrange(dplyr::desc(abs(log2FoldChange))) %>%
+  # Filter out the duplicated rows using `dplyr::distinct()`-- this will keep
```
cansavvy, line 343: I think this very long comment should be just a quick sentence before we do this action and then it can be shortened here. Some people don't read comments so this will make it more likely they don't miss it plus then we can avoid this multiline comment which makes this code chunk look bigger than it is.

cansavvy (commented):
I have some comments, but they are all minor, and you can see I don't feel strongly about some of them.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
cbethell: Ah yes good point 👍

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
cbethell: I agree, I took away the preview as we will see the duplicate gene symbols with their associated values in the next step.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -0,0 +1,403 @@
...
+    C4: computational gene sets
+    C5: GO gene sets
+    C6: oncogenic signatures
+    C7: immunologic signatures
+
+The data we're interested in here comes from human samples, so we can obtain just the gene sets (from all of the gene set collections listed above) relevant to _Homo sapiens_ by specifying `species = "Homo sapiens"`.
+
+```{r}
+hs_gene_sets <- msigdbr(
+  species = "Homo sapiens" # Replace with species name relevant to your data
```
jaclyn-taroni, line 230: I would not use all the gene sets here. It's not something I think we want to encourage. We'd rather folks select gene sets based on their question of interest. I do wonder if including _all_ of them might be contributing to the weirdness in #408. KEGG is pretty commonly used.

cbethell:
Closing this PR as updates and comments from here have been propagated over to #419

PR #407, opened 2020-12-03
Link citations in render

jashapiro:
This PR adds a very small change to the render script so that citations get hyperlinks to their positions in the bibliography. 

Separately, it also fixes citations for Sonja Hänzelmann so the umlaut is correctly rendered.

I didn't rerender anything, to keep the changes more modular, but I did check that it works...


cansavvy (approved):
Looks like good changes to me!

Issue #406, opened 2020-12-03
Enforce code width limit


jashapiro:
### Background/Problem

Related to #289 

Occasionally we have lines of code that get a bit long and don't render well in the html versions of notebooks. 
A normal code width limit of ~80-100 characters for a bare notebook is probably too generous at times for html display, especially if users try to narrow their windows.

This seems to occur in our code most often in comments (perhaps because these are not affected by styling), especially when trailing comments occur on a line. 

### What potential "gotchas" do we know of?

I don't think this is something [`styler`](https://styler.r-lib.org/index.html) can handle automatically, but if it can, that might be the ideal case. I can't find a line-width argument in its options, though it may be buried somewhere. 

`formatR` was an alternative I looked at that does handle this somewhat, but it seems to not play well with pipes, and is otherwise aggressive.

### What are the recommended next steps?

I am compiling a list of wide lines (>70 characters, as tha tis what I used for output in #402) based on the current staging branch, but maintaining this standard in the future will require vigilance, and it should be added to CONTRIBUTING.md at a minimum. 

Automated checking would be a nice feature if we can figure out a method.

I am going to add to this issue with separate comments for each file. 

### Is there a particular timeframe for this issue?

Before going live

jashapiro:
- [ ] https://github.com/AlexsLemonade/refinebio-examples/blob/9acefe54dfc0abe0da0b8f5f3c55c4bbb5bc1384/02-microarray/clustering_microarray_01_heatmap.Rmd#L199-L200
- [ ] https://github.com/AlexsLemonade/refinebio-examples/blob/9acefe54dfc0abe0da0b8f5f3c55c4bbb5bc1384/02-microarray/clustering_microarray_01_heatmap.Rmd#L233-L233
- [ ] https://github.com/AlexsLemonade/refinebio-examples/blob/9acefe54dfc0abe0da0b8f5f3c55c4bbb5bc1384/02-microarray/clustering_microarray_01_heatmap.Rmd#L247-L249
- [ ] https://github.com/AlexsLemonade/refinebio-examples/blob/9acefe54dfc0abe0da0b8f5f3c55c4bbb5bc1384/02-microarray/clustering_microarray_01_heatmap.Rmd#L272-L272
[...]

Issue #290, opened 2020-10-14
Editing: Pass through for content


jaclyn-taroni:
When reviewing examples one by one, it can be hard to keep an eye on the big picture. Specifically - keep an eye on things like the audience and assumptions about what that audience will know or is interested in. Once all the before going live examples go in, we should have at least one person go through the product as a whole and make notes about changes that should be made.

cansavvy:
From a conversation I had with @jaclyn-taroni, just wanted to mention that this issue should ideally be completed by someone with "more distance" from this project. Someone who isn't as familiar with all the material. Aka not @cbethell or myself.

jashapiro:
Closing this with the same logic as https://github.com/AlexsLemonade/refinebio-examples/issues/289#issuecomment-749110544. We'll catch any remaining changes with individual issues as they come up.

Issue #296, opened 2020-10-14
Further refine wording about: Why use microarray data?


cansavvy:
In the microarray section header, we have an "advantages of microarray" section. Currently we have a few listed bu tas @jashapiro noted here: https://github.com/AlexsLemonade/refinebio-examples/pull/232#discussion_r504189680 , our main point to our users about microarray is that it's useful, still relevant even thought RNA-seq exists, and it there's a lot of it out there that might prove to be helpful for their particular scientific questions and interests. 

I've struggled a bit with getting these points across eloquently in 02-microarray/00-intro-to-microarray.Rmd but I think this is something we should come back to because it's an important point to make.

jashapiro:
Closed by #409

PR #409, opened 2020-12-03
Editing/polish of microarray heatmap notebook

jashapiro:
### Purpose

This PR is born out of #289 and #290, as I read through the heatmap example for content and display issues. I did this based mostly on what was currently rendered on github pages as a baseline for ease of reading, though I do not think that is far behind what is in `staging` in this case.  All of the changes here though are of course based off `staging`.  

There were no major content changes I would suggest, so I mostly made small edits here. (Also a couple in the intro)

Many of the are streamlining of code comments, which is largely to address #406. 

Are the code comments still sufficiently clear and informative?

To the extent that some of those changes are in common template material, they should be translated to those sections in a later PR or PRs, but this seemed a good place to test them out.


jashapiro:
One other thought was with the use of `read_tsv()`; do we want to show the output of the column specification every time? I can make an argument either way, so I did not change it here, but I could see doing so. We do show the table later, so maybe that is sufficient?

cansavvy:
> One other thought was with the use of `read_tsv()`; do we want to show the output of the column specification every time? I can make an argument either way, so I did not change it here, but I could see doing so. We do show the table later, so maybe that is sufficient?

I've seen some training participants see the output from `read_tsv()` and think its an error. So if we hide those messages from them, I think we are increasing the chances that when the user tries to run `read_tsv()` on their own they will be surprised by the output and think they've done something wrong. 

I think hiding some stuff is okay so we make it pretty, but I think we want them to see the normal output for other things so when they move to another chunk or notebook they aren't shocked about all the output we've hidden from them.

jashapiro:
> I've seen some training participants see the output from `read_tsv()` and think its an error. So if we hide those messages from them, I think we are increasing the chances that when the user tries to run `read_tsv()` on their own they will be surprised by the output and think they've done something wrong.

This was the main reason I had not to remove it. I was thinking about suppressing the output not with `message = FALSE` but with `read_tsv(file, col_types = cols())` with a comment that including the second line is to avoid wordy output.

jaclyn-taroni (approved):
👍 LGTM. The inline nature of the `Replace ...` comments were from user feedback (#99; you may have already been aware of this issue), but it doesn't seem like there's a way to do both.

jashapiro:
Added a bit more commentary on why microarrays are awesome wrt refine.bio to address #296 . Thoughts?

jaclyn-taroni (approved):
Most recent changes wrt #296 LGTM!

Issue #299, opened 2020-10-14
Improve approach to getting the data for Ortholog Mapping examples


cbethell:
After some review comments and discussion on the microarray ortholog mapping example notebook update PR #246, there is an outstanding question of how to get the data from HGNC for the Ortholog Mapping example notebooks.

In that PR, we direct users to download the data file directly from the FTP server link: ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/

However, as noted in one of @jashapiro's [comments on there](https://github.com/AlexsLemonade/refinebio-examples/pull/246/#discussion_r504672688), clicking this link on a Mac in Safari opens the link in the Finder and this is not the ideal behavior (so we made a note in that PR as a placeholder but would need to find a more robust solution).

Additionally, once a user clicks a file to download, Safari does another less than ideal thing (noted in [another one of Josh's comments](https://github.com/AlexsLemonade/refinebio-examples/pull/246/#discussion_r504692367)) where it automatically decompresses the file so the file name after download would be different across users as those using Safari may have `human_zebrafish_hcop_fifteen_column.txt` while other users will have `human_zebrafish_hcop_fifteen_column.txt.gz` (therefore forcing us to add another note to the notebook in that PR).

To summarize, the general aspects to think about here are how do we handle getting the data while considering:
1. Safari users who cannot access the ftp link
2. Safari users whose downloaded file will be decompressed instead of compressed (.gz)
3. Other users for machines that we have not yet tested (like Windows)

cansavvy:
I suppose for our users we don't necessarily want to have them construct a `wget` command? Or a `download.file()` in R?

cansavvy:
i've tried a download.file() solution. I think its better? Will file a draft PR to see what people think. Also found out that Safari's problems can be circumvented by changing the ftp to http if you just want to see the file itself. (You'll see what I mean in the draft PR).

PR #411, opened 2020-12-04
Try different strategy for ortholog file download

cansavvy:
### Purpose

I read #299 and gathered the main problems with the ortholog file download were all the Safari caveats:

>To summarize, the general aspects to think about here are how do we handle getting the data while considering:
> Safari users who cannot access the ftp link
Safari users whose downloaded file will be decompressed instead of compressed (.gz)
Other users for machines that we have not yet tested (like Windows)

### Issue addressed

#299 

### Strategy

1) I made this download by `download.file()` which should work no matter the browser someone may be using.
2) I found a thing on Apple that says that if you want to view ftp using Safari, you can change it to an http. So for looking at the files available, I steer the user to that file page but a different link (with http) if they are using Safari. 
3) I trimmed out the GUI instructions since I think this should work no matter what. (I wasn't able to get the GUI to work for me for some reason anyway -- the files won't download). 

I tried think link on Safari and it works. 

Additionally the data file got changed to the wrong dataset that was mouse? This got messed up when I did #298 apparently. It still was running like that but not really, everything was returning 0 rows and no errors. I changed it back to the correct zebrafish so it was actually working. (This is a tad scary and when we are taking a look at polishing the other dataset examples, we should make sure nothing else like this slipped through) 

Do we feel like this is less confusing to follow and more flexible for people who might be using different web browsers?

### DIFF
--- a/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
+++ b/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
@@ -192,7 +192,7 @@ zebrafish_hgnc_file <- file.path(
 )
 ```
 
-Using the file path we just declared, we can use the `destfile` argument to download the file to this directory and use this file name. 
+Using the file path we just declared, we can use the `destfile` argument to download the file we need to this directory and use this file name. 
 
 We are downloading this orthology predictions file from the [HGNC database](https://www.genenames.org/).
 If you are looking for a different species, see the [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](http://ftp.ebi.ac.uk/pub/databases/genenames/hcop/) and find the file name of the species you are looking for.
--- a/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
+++ b/02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
@@ -198,7 +198,7 @@ Using the file path we just declared, we can use the `destfile` argument to down
 download.file(
   paste0(
     "ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/",
-    # Can replace with the file name of the species conversion you are looking for
+    # Replace with the file name of the species conversion you are looking for
     "human_zebrafish_hcop_fifteen_column.txt.gz"
   ),
   # The file will be saved to this location and with this name
@@ -206,8 +206,8 @@ download.file(
 )
 ```
 
-If you are looking for a different species, see the [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/) and find the file name of the species you are looking for
-If you are using Safari as your web browser see the directory using [this link](http://ftp.ebi.ac.uk/pub/databases/genenames/hcop/).
+If you are looking for a different species, see the [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/) and find the file name of the species you are looking for.
+**If you are using Safari** as your web browser see the directory using [this link](http://ftp.ebi.ac.uk/pub/databases/genenames/hcop/).
 This is where the files that reflect the data provided via the [HGNC database](https://www.genenames.org/) are maintained.
 If you are using a different dataset, in the last chunk you can replace `zebrafish` in `human_zebrafish_hcop_fifteen_column.txt.gz` with the name of the species you have data for (if you see it listed in the directory).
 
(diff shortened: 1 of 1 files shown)


02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -122,13 +122,13 @@ This is handy to do because if we want to switch the dataset (see next section f
 
 ```{r}
 # Define the file path to the data directory
-data_dir <- file.path("data", "GSE13490") # Replace with accession number which will be the name of the folder the files will be in
+data_dir <- file.path("data", "GSE71270") # Replace with accession number which will be the name of the folder the files will be in
```
jashapiro, line 125: Are these accession number changes correct?

cansavvy: Yes. And that's the scary part. I accidentally changed these for #298 and this didn't throw an error since this analysis doesn't have an particular step that was affected by it being a different species.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -178,36 +178,49 @@ The [HGNC Comparison of Orthology Predictions (HCOP)](https://www.genenames.org/
...
-For this notebook, you will want to download the file named `human_zebrafish_hcop_fifteen_column.txt.gz`.
-If you are using a different dataset, you can replace `zebrafish` in `human_zebrafish_hcop_fifteen_column.txt.gz` with the name of the species you have data for, and click on that file to download.
+Using the file path we just declared, we can use the `destfile` argument to download the file to this directory and use this file name. 
 
-<img src="https://github.com/AlexsLemonade/refinebio-examples/raw/07499927a22ed7aa045245e40d692f8e49d7ba52/template/screenshots/download-hgnc-zebrafish-file.png" width=400>
+```{r}
+download.file(
+  paste0(
+    "ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/",
+    # Can replace with the file name of the species conversion you are looking for
```
jashapiro, line 201: Without looking, I don't think people will be able to guess what the correct file name is. Perhaps restore lines 191-192 for context?

Also, comment width:
```suggestion
    # Replace with the file name of the species conversion you are looking for
```

cansavvy: I changed that but didn't push the commit yet.  (the first thing, not the comment thing)

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -178,36 +178,49 @@ The [HGNC Comparison of Orthology Predictions (HCOP)](https://www.genenames.org/
...
+    "human_zebrafish_hcop_fifteen_column.txt.gz"
+  ),
+  # The file will be saved to this location and with this name
+  destfile = zebrafish_hgnc_file
+)
+```
 
-Next, move the `human_zebrafish_hcop_fifteen_column.txt.gz` file into your `data/` folder.
+If you are looking for a different species, see the [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/) and find the file name of the species you are looking for
+If you are using Safari as your web browser see the directory using [this link](http://ftp.ebi.ac.uk/pub/databases/genenames/hcop/).
```
jashapiro, line 210: The `http` link should work in any browser, so let's just use that one. And from my previous comment, ets move this above the `download.file()` chunk to give people context _first_.

jashapiro (approved):
LGTM! Wondering if we should apply the same method to the ORA file download? That might be less confusing than replacing a URL with a file name here: https://github.com/AlexsLemonade/refinebio-examples/blob/cb39d7c06d4edebcf7b7dae91eaf5bbffb30369d/02-microarray/pathway-analysis_microarray_01_ora.Rmd#L191-L204

cansavvy:
> LGTM! Wondering if we should apply the same method to the ORA file download? That might be less confusing than replacing a URL with a file name here:
> 
> https://github.com/AlexsLemonade/refinebio-examples/blob/cb39d7c06d4edebcf7b7dae91eaf5bbffb30369d/02-microarray/pathway-analysis_microarray_01_ora.Rmd#L191-L204

Well, for sure I'm going to file a follow up PR for the RNA-seq ortholog example. 

I can see your point about the downloading the file being less confusing there too though. I'll file a separate issue about that. Seems reasonable. (GSEA examples have the same kind of URL read in step so this would probably apply to them too).

PR #413, opened 2020-12-04
Use same download.file strategy for ortholog RNA-seq example

### LINKED ISSUE #299: Improve approach to getting the data for Ortholog Mapping examples

cansavvy:
### Purpose

This is a direct copy over from #411 's changes but I changed zebrafish -> mouse of course. 

### Issue addressed

Closes #299 

### Strategy

Also copied from #411: 

> 1) I made this download by `download.file()` which should work no matter the browser someone may be using.
> 2) I found a thing on Apple that says that if you want to view ftp using Safari, you can change it to an http. So for looking at the files available, I steer the user to that file page but a different link (with http) if they are using Safari. 
> 3) I trimmed out the GUI instructions since I think this should work no matter what. (I wasn't able to get the GUI to work for me for some reason anyway -- the files won't download). 

> I tried think link on Safari and it works. 

Changes look okay?

### DIFF
--- a/03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
+++ b/03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
@@ -187,38 +187,51 @@ The [HGNC database](https://www.genenames.org/) currently contains over 39,000 p
 
 The [HGNC Comparison of Orthology Predictions (HCOP)](https://www.genenames.org/tools/hcop/) is a search tool that combines orthology predictions for a specified human gene, or set of human genes from a variety of sources, including Ensembl Compara, HGNC, and NCBI Gene Orthology [@Wright2005].
 In general, an orthology prediction where most of the databases concur would be considered the reliable, and we will use this to prioritize mapping in cases where there is more than one possible ortholog for a gene.
-HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes [@hcop-help].
+HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes, including mouse, which we will use in this notebook [@hcop-help].
 
-First, we need to download the file from the server holding the HGNC data.
-Go to this [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/).
+We can automatically download the human mouse file we need for this example using `download.file()` command. 
+For this notebook, we want to download the file named `human_mouse_hcop_fifteen_column.txt.gz`.
 
-This is where the files that reflect the data provided via the [HGNC database](https://www.genenames.org/) are maintained.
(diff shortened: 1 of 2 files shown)


03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
```diff
@@ -187,38 +187,51 @@ The [HGNC database](https://www.genenames.org/) currently contains over 39,000 p
...
 In general, an orthology prediction where most of the databases concur would be considered the reliable, and we will use this to prioritize mapping in cases where there is more than one possible ortholog for a gene.
-HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes [@hcop-help].
+HCOP was originally designed to show orthology predictions between human and mouse, but has been expanded to include data from 18 genomes, including mouse, which we will use in this notebook [@hcop-help].
 
-First, we need to download the file from the server holding the HGNC data.
-Go to this [directory page of the HGNC Comparison of Orthology Predictions (HCOP) files](ftp://ftp.ebi.ac.uk/pub/databases/genenames/hcop/).
+We can automatically download the human mouse file we need for this example using `download.file()` command.
```
jashapiro, line 192: Oh, I meant to make this comment in the other one but it slipped my mind... it isn't really automatic. (port back to the other file as well, please?)

```suggestion
We can download the human mouse file we need for this example using `download.file()` command. 
```

jashapiro (approved):
Looks good, less one or two "automatically"s.

03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
cansavvy: Yes, I will add this change to the other file too.

PR #414, opened 2020-12-04
Carry over common  comment changes

jashapiro:
I made a number of changes in #409 to shorten comments to keep things within width limits discussed in #406 , some of which are repeated across many files. This PR transfers the common changes across the repo, including to the template file.

This does not address every change needed for #406, so looking at individual files is still required, but it gets a big chunk of the easy changes done!


cansavvy (approved):
Looks good to me!! 🚀

jashapiro:
The check here isn't happening because the last commit was by the bot... I think it is good to merge.

PR #418, opened 2020-12-07
Try out a different download strategy for ORA

cansavvy:
### Purpose

As a result of #411, we thought it'd be worth while to take another whack at other places we download files for use in analyses examples. So here I've tried a similar type of strategy with ORA. 

### Issue addressed
#413 

It's not shorter, but I think it is more intuitive for users. How do we feel about this?


02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -193,18 +193,44 @@ First we will assign the URL to its own variable called, `dge_url`.
...
-# desired gene list results
-dge_df <- readr::read_tsv(dge_url)
+# Read in the contents of the differential expression results file
+dge_df <- readr::read_tsv(dge_results_file)
 ```
 
+Note you can also supply `dge_url` directly to the `readr::read_tsv()` function and read in the file without downloading it. 
 `read_tsv()` can read TSV files online and doesn't necessarily require you download the file first.
-Let's take a look at what these contrast results from the differential expression analysis look like.
+Let's take a look at what these results from the differential expression analysis look like.
```
jashapiro, line 233: Separate paragraph?
```suggestion

Let's take a look at what these results from the differential expression analysis look like.
```

02-microarray/pathway-analysis_microarray_01_ora.Rmd
```diff
@@ -193,18 +193,44 @@ First we will assign the URL to its own variable called, `dge_url`.
...
-# Read in the contents of your differential expression results file
-# `dge_url` can be replaced with a file path to a TSV file with your
-# desired gene list results
-dge_df <- readr::read_tsv(dge_url)
+# Read in the contents of the differential expression results file
+dge_df <- readr::read_tsv(dge_results_file)
 ```
 
+Note you can also supply `dge_url` directly to the `readr::read_tsv()` function and read in the file without downloading it. 
 `read_tsv()` can read TSV files online and doesn't necessarily require you download the file first.
```
jashapiro, line 232: These two sentences are a bit redundant, so I am consolidating and providing a bit more specificity. If we want to provide this hint at all?
```suggestion
Note that `read_tsv()` can also read TSV files directly from a URL and doesn't necessarily require you download the file first.
If you wanted to use that feature, you could replace the call above with `readr::read_tsv(dge_url)` and skip the download steps.
```

jashapiro (commented):
Looks good. I added some clarification and a bit of space. I'm not 100% sold on the need to describe reading from a URL here, but I'm not opposed to it either. I just worry it will add unnecessary distraction and confusion.

02-microarray/pathway-analysis_microarray_01_ora.Rmd
cansavvy: I like the hint just in case they knew about this and are confused why we even bother downloading first. But maybe we should just assume they won't know.

jashapiro (approved):
🚀

PR #419, opened 2020-12-07
PR 1 of 2: Add RNA-seq Pathway Analysis - GSEA example (take 2)

cbethell:
## Analysis Purpose

This PR addresses issue #368.

## Pull Request Stage

This is a **Refined PR** - needs review of details and polishing

The relevant draft PR is #397

## Strategy

This PR has generally the same strategy as the first take at PR 1 of this example series, with exception of the use of the new differential expression results file which is generated as a result of the updates in PR #416 and the move back to using hallmark gene sets.

The strategy from take one #405 is as follows:

I mostly copied and pasted the first half of the code from the draft PR (stopping right before the step implementing the `GSEA()` function) and refined the context/user instructions along the way.

I updated the pathway analysis introduction to be relevant to the RNA-seq modules.
The specific example introduction, although copied from the microarray, I _believe_ still applies to this GSEA example. However, please correct me if I am wrong! 

Per [this comment on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/397#discussion_r535257352), I included some context around the log2 fold change values we are using in this example analysis.

I also include a step to set the seed in this PR as I noticed the top 3 pathways would change with each run (although `seed = TRUE` is provided to the `GSEA()` function and this won't be evident until the upcoming PR 2 of 2).

## Concerns/Questions for reviewers:

Overall, the context/user instructions should receive an extra close look and that suggestions have been propagated over from #405.
That said, as I am writing this I am realizing that we may want switch over to using KEGG gene sets in this PR based on [this comment on PR #405](https://github.com/AlexsLemonade/refinebio-examples/pull/405#discussion_r535801263). Is this something that we would still want to do although the results/plots are looking much better now?

### Content checks
* [x] All `{{BLANKS}}` have been replaced with the correct content.
* [x] [Sources are cited](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#citing-sources-in-text)
* [x] Seed is set (if applicable)

### Formatting Checks
* [x] Removed any manual numbering of sections.
* [x] Removed any instances of chunk naming.
* [x] Comments and documentation are up to date.
* [x] All links have been checked and are properly formatted.

### Add datasets to S3
* [x] Added [data and metadata files to S3](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#adding-datasets-to-the-S3-bucket).

### Docker/Snakemake rendering components
[...]

PR #416, opened 2020-12-07
RNA-seq DGE dataset switch

cansavvy:
### Purpose

The previous dataset for RNA-seq dge was found to be more and more wonky as we've looked into. 

### Issue addressed

Step 1 of the plan described here on GSEA Draft PR: https://github.com/AlexsLemonade/refinebio-examples/pull/397#issuecomment-739408396

> 1. Swap in a new dataset in `03-rnaseq/differential-expression_rnaseq_01` – maybe [SRP123625](https://www.refine.bio/experiments/SRP123625/translatome-analysis-of-the-ribosomal-protein-l10-r98s-mutation-reveals-altered-serine-metabolism-in-acute-lymphoblastic-leukemia-supplementaryrna-seq) for straightforward experimental design + _some metadata_ reasons?

#### Gotchas the reviewer should know about

Honestly things look good! 

Besides things directly related to switching the dataset I tried to slip into two (kind of three) other changes 😬 

- I did some wording updates and got rid of the two plot steps that was used before (It seems less necessary now that we have more significant results). 
- I added a few more sentences to the Purposes section since I was already here. It probably still could use more but it's a least a little less sad. 
- I switched the filtering to come before the DESeq2 object. It is much more clear. (I will make an issue about this for us to address this elsewhere). 

## Remaining concerns and questions

- Does my wording about the new dataset make sense? 
- Anything else of concern you notice?

PR #420, opened 2020-12-07
PR 2 of 2: Add RNA-seq Pathway Analysis - GSEA example (take 2)

cbethell:
## Analysis Purpose

This PR closes issue #368.

## Pull Request Stage

The relevant draft PR is #397

## Strategy

This PR has generally the same strategy as the first take at PR 2 of this example series, with exception of the use of the new differential expression results file which is generated as a result of the updates in PR #416 and the move back to using hallmark gene sets.

The strategy from take one #408 is as follows:

This PR iterates on open PR #419 by adding the GSEA and visualization steps to the RNA-seq GSEA example notebook.

I copied over the remaining steps from draft PR #397.
As [noted on the draft PR](https://github.com/AlexsLemonade/refinebio-examples/pull/397#issuecomment-740080645), I added a note re the warning that GSEA produces.

As [noted on closed PR 1 of 2](https://github.com/AlexsLemonade/refinebio-examples/pull/405#issuecomment-738134363), a step to set the seed was added at the beginning of the example notebook for the purpose of reproducing the GSEA results.

Also worth noting is the shift from using `top_n()` to using `slice_min()` and `slice_max()`.
I noticed that `top_n(n = -3)` did not take into account directionality, meaning that both positive and negative values were returned.
According to the documentation at `?top_n` and the documentation [here](https://rdrr.io/cran/dplyr/man/top_n.html), `top_n()` has been superseded by `slice_min()`/`slice_max()`.

Review for this PR should begin at line 393 (the "Run GSEA using the GSEA() function" section) as this is where PR 1 of 2 #419 ends.

Taken from the first take at this PR: I would encourage the reviewer to re-run this locally to ensure that the results are indeed reproducible with the new strategy (setting the seed and providing the logical argument to the GSEA() function).
Also, the strategy change from top_n() to slice_min()/slice_max() could likely receive a close look (as well as the visualizations).

I will also note that I propagated the suggestions over from the first take at this PR #408, but was not sure if [this particular suggestion re a plot description](https://github.com/AlexsLemonade/refinebio-examples/pull/408#discussion_r535529768) still applies. 

More specifically, do we still want to leave the sentence
`Notice how the genes that are in the gene set, indicated by the black bars, tend to be on the left side of the graph indicating that they have positive gene-level scores` out?
This sentence was originally after the `most_positive_nes_plot` but was removed after it no longer accurately described the strange plots in that PR. Now that we have a better looking plot, do we want to add this sentence back?

### Content checks

### Formatting Checks

### Add datasets to S3

* [x] Added the `.html` [link to the navigation bar](https://github.com/AlexsLemonade/refinebio-examples/blob/master/CONTRIBUTING.md#add-new-analyses-to-the-navbar). (in PR 1 of 2)


03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -392,11 +392,148 @@ head(lfc_vector)
...
+head(gsea_results@result)
+```
+
+Significance is assessed by permutating the gene labels of the pre-ranked gene list and recomputing the ES of the gene set for the permutated data, which generates a null distribution for the ES.
+The ES for each gene set is then normalized to account for the size of the set, resulting in a normalized enrichment score (NES), and an FDR (false discovery rate) value is calculated to account for multiple hypothesis testing.
+
+Looks like we have gene sets returned as significant at FDR of `0.05`, according to the `p.adjust` column in the table above.
+
+The information we're most likely interested in is in the `results` slot.
+Let's convert this into a data frame that we can use for further analysis and write to file.
```
jaclyn-taroni, line 435: I would maybe note what happens if you don't have significant results (e.g., this slot is empty!). I believe if you want _all_ results returned you need to set the `pvalueCutoff = 1`.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -392,11 +392,148 @@ head(lfc_vector)
...
+
+```{r}
+gsea_result_df <- data.frame(gsea_results@result)
+```
+
+## Visualizing results
+
+We can visualize GSEA results for individual pathways or gene sets using `enrichplot::gseaplot()`.
+Let's take a look at 2 different pathways -- one with a highly positive NES and one with a highly negative NES -- to get more insight into how ES are calculated.
+Note that if we didn't have any significant pathways, our visualizations below would show up blank as nothing would have met our `pvalueCutoff` above.
```
jaclyn-taroni, line 445: I think this is only true for the dot plots? You are able to pull out pathways from `gsea_results` that don't meet the `pvalueCutoff` using `enrichplot::gseaplot()`.

jaclyn-taroni (approved):
👍  Two comments about adding some text or subtracting it! No need for me to re-review.

PR #422, opened 2020-12-08
Use download.file for the three other notebooks

### LINKED ISSUE #412: For URL read in steps, change to `download.file()` for less confusion
### Background

For the ortholog examples, we download a file from online and I'm currently working on revamping those steps to be a little more succinct and use `download.file()`. 

We could try a similar strategy for the ORA and GSEA examples (both microarray and RNA-seq versions) perhaps that would be more clear than reading in from a URL directly?

https://github.com/AlexsLemonade/refinebio-examples/pull/411#pullrequestreview-545102321

### Problem

Reading in from a URL is probably not super clear to users. Maybe download first and then read in is better for explanation purposes? 

- [ ] ORA microarray 
- [ ] GSEA microarray 
- [ ] ORA RNA-seq
- [ ] GSEA RNA-seq

cansavvy:
### Purpose

As a result of #411, we thought it'd be worth while to take another whack at other places we download files for use in analyses examples. So here I've tried a similar type of strategy with ORA. 

We tried this out first on #418 with microarray ORA. 

In this PR here I'm using the same wording and steps but I attempt to carry this over to the other three notebooks that could use this change and are listed in #412: 

- microarray GSEA 
- rnaseq ORA
- rnaseq GSEA 

I also caught one comment that was out of date in the microarray ORA that we missed in #418.
### Issue addressed
Should be able to close #412  

Everything runs fine. Main thing to look for is that all comments are updated for the specific notebook they are in.


02-microarray/pathway-analysis_microarray_02_gsea.Rmd
```diff
@@ -145,33 +145,60 @@ library(org.Dr.eg.db)
...
+  dge_url,
+  # The file will be saved to this location and with this name
+  destfile = dge_results_file
+)
+```
+
+Now let's double check that the results file is in the right place.
+
+```{r}
+# Check if the organism orthology file file is in the `data` directory
```
cbethell, line 185:
```suggestion
# Check if the file exists
```

I believe this should also be changed here if I'm not mistaken.

03-rnaseq/pathway-analysis_rnaseq_02_gsea.Rmd
```diff
@@ -187,17 +187,44 @@ First we will assign the URL to its own variable called, `dge_url`.
 dge_url <- "https://refinebio-examples.s3.us-east-2.amazonaws.com/03-rnaseq/results/SRP123625/SRP123625_differential_expression_results.tsv"
 ```
```
cbethell, line 188: It won't allow me to comment a little further up but looks like the `##Import data` header on line 175 of this notebook still needs to be changed to `## Download data file`

cbethell (approved):
LGTM once the two small comments I left below are addressed! 👍

PR #425, opened 2020-12-10
Move filtering to before DESeq2 object creation 

### LINKED ISSUE #417: RNA-seq examples updates: Move filtering before DESeqDataSet creation
### Background

If we do counts filtering before we make a DESeqDataSet we can use tidyverse and its more clear. 

See https://github.com/AlexsLemonade/refinebio-examples/pull/416

### Problem

You can make this filtering way more clear and tidyverse if we do the filtering before we make the DESeqDataset creation. 
Filtering the ddset itself is not super good looking or clear. @jashapiro mentioned something about this somewhere but I don't think I ever made it into an issue. 

### What are the recommended next steps?

Here's what that step can look like if we move it to right before we do a DESeq2 thing: 
```
# Define a minimum counts cutoff and filter the data to include
# only rows (genes) that have total counts above the cutoff
filtered_expression_df <- expression_df %>%
  dplyr::filter(rowSums(.) >= 10)
```

This is relevant to all the places we do a counts filter in RNA-seq. 

From a quick `Define a minimum counts cutoff` search in project, it looks like these are the examples that could use this change: 

- [ ] clustering_rnaseq_01_heatmap 
- [x] differential-expression_rnaseq_01.html
- [ ] dimension-reduction_rnaseq_01_pca.html
- [ ] dimension-reduction_rnaseq_02_umap.html

cansavvy:
### Purpose

At some point jashapiro pointed out that doing the filter step before we set up the DESeq2 object makes the filtering step much more intuitive to follow! So #417 tracked that this should be done elsewhere. 

the differential expression example was already addressed in this respect in #416 

So this PR I've made this change in three other places (which is I believe the rest of the places we need it): 

- Clustering RNA-seq
- PCA RNA-seq 
- UMAP RNA-seq

Also took this opportunity to rename `df` -> `expression_df`. 
### Issue addressed

Should close #417

Everything runs fine. 

- Put the rounding steps underneath the same header -- should it have its own header or be moved underneath a different header in the subsequent step?
- Did I miss any comment updates? I feel like that's the main thing to look out for here.


cansavvy:
Okay, I believe the merge conflicts are resolved correctly. @jashapiro

03-rnaseq/clustering_rnaseq_01_heatmap.Rmd
```diff
@@ -232,47 +232,50 @@ Now let's ensure that the metadata and data are in the same sample order.
...
 # Check if this is in the same order
-all.equal(colnames(df), metadata$refinebio_accession_code)
+all.equal(colnames(expression_df), metadata$refinebio_accession_code)
 ```
 
 Now we are going to use a combination of functions from the `DESeq2` and `pheatmap` packages to look at how are samples and genes are clustering.
 
+## Define a minimum counts cutoff
+
+We want to filter out the genes that have not been expressed or that have low expression counts because we want to remove any possible noise from our data before we normalize the data and create our heatmap.
```
jashapiro, line 246: We definitely aren't getting rid of all the noise! 
```suggestion
We want to filter out the genes that have not been expressed or that have low expression counts since these genes are likely to add noise rather than useful signal to our analysis.
```

03-rnaseq/dimension-reduction_rnaseq_02_umap.Rmd
```diff
@@ -273,6 +264,26 @@ metadata <- metadata %>%
   )
 ```
 
+## Define a minimum counts cutoff
+
+We want to filter out the genes that have not been expressed or that have low expression counts because we want to remove any possible noise from our data before we normalize the data and create our heatmap.
```
jashapiro, line 269:
```suggestion
We want to filter out the genes that have not been expressed or that have low expression counts since these genes are likely to add noise rather than useful signal to our analysis.
```

03-rnaseq/clustering_rnaseq_01_heatmap.Rmd
```diff
@@ -232,47 +232,50 @@ Now let's ensure that the metadata and data are in the same sample order.
...
-df <- df %>%
-  # Mutate numeric variables to be integers
-  dplyr::mutate_if(is.numeric, round)
-
 # Create a `DESeqDataSet` object
 dds <- DESeqDataSetFromMatrix(
-  countData = df, # This is the data.frame with the counts values for all replicates in our dataset
+  countData = filtered_expression_df, # This is the data.frame with the counts values for all replicates in our dataset
   colData = metadata, # This is the data.frame with the annotation data for the replicates in the counts data.frame
   design = ~1 # Here we are not specifying a model -- Replace with an appropriate design variable for your analysis
```
jashapiro, line 275: I guess I haven't gotten to these yet, but can you shorten up these comments to match the dimension reduction notebooks?

03-rnaseq/clustering_rnaseq_01_heatmap.Rmd
```diff
@@ -232,47 +232,50 @@ Now let's ensure that the metadata and data are in the same sample order.
...
+# Define a minimum counts cutoff and filter the data to include
+# only rows (genes) that have total counts above the cutoff
+filtered_expression_df <- expression_df %>%
+  dplyr::filter(rowSums(.) >= 10)
+```
+
+We also need our counts to be rounded before we can use them with the `DESeqDataSetFromMatrix()` function.
+
+```{r}
+# The `DESeqDataSetFromMatrix()` function needs the values to be converted to integers
```
jashapiro, line 260: Line width, mostly.
```suggestion
# The `DESeqDataSetFromMatrix()` function needs the values to be integers
```

03-rnaseq/dimension-reduction_rnaseq_01_pca.Rmd
```diff
@@ -256,6 +248,27 @@ metadata <- metadata %>%
...
+# Define a minimum counts cutoff and filter the data to include
+# only rows (genes) that have total counts above the cutoff
+filtered_expression_df <- expression_df %>%
+  dplyr::filter(rowSums(.) >= 10)
+```
+
+We also need our counts to be rounded before we can use them with the `DESeqDataSetFromMatrix()` function.
+
+```{r}
+# The `DESeqDataSetFromMatrix()` function needs the values to be converted to integers
```
jashapiro, line 268:
```suggestion
# The `DESeqDataSetFromMatrix()` function needs the values to be integers
```

jashapiro (commented):
Looks good. A few little comment shortening changes suggested.

jaclyn-taroni (approved):
I went through and looked at whether all of Josh's comments were addressed. They were, so I am approving!

PR #427, opened 2020-12-14
Changes to differential exp microarray notebook

jashapiro:
This is the start of polishing the differential expression notebooks.

In this section I am adding changes to the microarray 2 group sample. Noteable changes include some additional explanation of design matrixes as well as recoding a factor. I could do this some other ways, but I thought this one was nice enough. 

One thing that is a bit of a note there is that when I make the factor, I don't explicitly set the control as the first factor, because it happens to be alphabetically first, so things work out. I could make that explicit.

I am submitting this as a draft because I think it will make sense to transfer these changes (or similar) to the RNAseq notebook and multi-group notebooks, but I didn't want to do this before getting a bit of feedback. 

Is the design matrix discussion helpful and clear? Are there other changes that I should make as well?


jashapiro:
some spellcheck fixes as I encountered them...

cansavvy:
I think these changes look good! But I do think we should put some sort of warning about setting the factor etc for when someone will change the dataset:

Since I feel like this is a thing that if you don't know about it, would definitely sneak right by you.

02-microarray/differential-expression_microarray_01_2-groups.Rmd
```diff
@@ -238,43 +238,63 @@ all.equal(colnames(df), metadata$geo_accession)
...
+# These renaming steps will not be the same (or might not be needed at all)
+# with a different dataset
 metadata <- metadata %>%
-  dplyr::rename("genotype" = `genotype/variation`) # This step will not be the same (or might not be needed at all) with a different dataset
+  # rename the column
+  dplyr::rename("genotype" = `genotype/variation`) %>%
+  # change the name of the CREB genotype (and make genotype a factor)
+  dplyr::mutate(
+    genotype = forcats::fct_recode(genotype,
+      CREB = "overexpressing the human CREB"
```
cansavvy, line 261: Might want a comment here to explain this particular line since its the "meat" of this step.

jashapiro:
I finished a pass through the microarray examples, so I think this is ready to review. The main changes are a bit of extra explanation in places and some code reformatting to simplify a few things (like avoiding dealing with `$` in model matrices).

I will transfer changes to RNAseq notebooks in a separate PR.

There are a number of small changes from spelling errors that were caught in an earlier pass, as well. 

The main files to look at can be viewed in rendered form at: 
https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/jashapiro/diff-polish/02-microarray/differential-expression_microarray_01_2-groups.html

and 

https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/jashapiro/diff-polish/02-microarray/differential-expression_microarray_02_several-groups.html

02-microarray/differential-expression_microarray_01_2-groups.Rmd
```diff
@@ -180,10 +180,6 @@ if (!("EnhancedVolcano" %in% installed.packages())) {
   # Install this package if it isn't installed yet
   BiocManager::install("EnhancedVolcano", update = FALSE)
 }
-if (!("apeglm" %in% installed.packages())) {
```
cansavvy, line 183: It runs without this now?

cansavvy: Did you remove this from the Docker image and try?

02-microarray/differential-expression_microarray_01_2-groups.Rmd
```diff
@@ -238,43 +234,70 @@ all.equal(colnames(df), metadata$geo_accession)
...
+
 ## Perform differential expression
 
-After applying our data to linear model, in this example we apply empirical Bayes smoothing and Benjamini-Hochberg multiple testing correction.
+We will use the `lmFit()` function from the `limma` package to test each gene for differential expression between the two groups using a linear model.
+After fitting our data to the linear model, in this example we apply empirical Bayes smoothing with the `eBayes()` function.
+
+Here's a [nifty article and example](http://varianceexplained.org/r/empirical_bayes_baseball/) about what the empirical Bayes smoothing is for [@bayes-estimates].
+
+Because we are testing many different genes at once, we also want to perform some multiple test corrections, which we will do with the Benjamini-Hochberg method while making a table of results with `topTable()`.
```
cansavvy, line 285: Should this be right before `topTable()` and the `toptable()` bit be made into its own code chunk?

02-microarray/differential-expression_microarray_01_2-groups.Rmd
```diff
@@ -238,43 +234,70 @@ all.equal(colnames(df), metadata$geo_accession)
...
+When we look at this design matrix, we see that there is now a `genotypeCREB` column that defines the group for each sample: 0 for control samples and 1 for the CREB samples. 
+(The model will also fit an intercept for all samples, so we can see that here as well.)
+
 ## Perform differential expression
 
-After applying our data to linear model, in this example we apply empirical Bayes smoothing and Benjamini-Hochberg multiple testing correction.
+We will use the `lmFit()` function from the `limma` package to test each gene for differential expression between the two groups using a linear model.
+After fitting our data to the linear model, in this example we apply empirical Bayes smoothing with the `eBayes()` function.
+
+Here's a [nifty article and example](http://varianceexplained.org/r/empirical_bayes_baseball/) about what the empirical Bayes smoothing is for [@bayes-estimates].
```
cansavvy, line 283: Maybe same with this bit.

cansavvy (commented):
Most of this looks good. I just have a couple comment/questions.

02-microarray/differential-expression_microarray_01_2-groups.Rmd
jashapiro: I didn't remove it from the docker image, but on my local instance I do not have it, and everything runs fine. This may be needed for the RNAseq notebook (lfcShrink uses it, I think), but not here.

02-microarray/differential-expression_microarray_01_2-groups.Rmd
jashapiro: Yeah, that makes sense, and I think it is that way in the other notebook here.

02-microarray/differential-expression_microarray_01_2-groups.Rmd
jashapiro: You mean to separate the `lmFit` from the `eBayes`? I think these are okay to keep together, as the bayes estimate is improving the original fit. Or do you mean something else?

cansavvy: Just if we wanted to have the eBayes explanation right before we use the `eBayes()` function. But really either is fine.

02-microarray/differential-expression_microarray_01_2-groups.Rmd
cansavvy: EnhancedVolcano is what complained to me. But if you are not getting that complaint, then let's remove it. It is probably RNA-seq specific then!

jashapiro: Yes, `EnhancedVolcano` suggests `DESeq2`, which suggests `apeglm`, so it is probably only on the code paths related to RNAseq.

cansavvy (approved):
Looks good to me!

Issue #428, opened 2020-12-15
Remove rownames from printed tables


jashapiro:
### Background

With the paginated tables option in our notebooks, outputs of data frames look much nicer, but sometimes there are row names that are shown which because of spacing can make things a bit confusing. 

### Problem

Row names that are just numbers can get mushed up against the first column, making it hard to read or confusing. See the example below from https://alexslemonade.github.io/refinebio-examples/02-microarray/differential-expression_microarray_01_2-groups.html

<img width="500" alt="Screen Shot 2020-12-15 at 4 08 25 PM" src="https://user-images.githubusercontent.com/20483/102273076-c8afc380-3eef-11eb-9a0e-5176b6e8403c.png">

This can be avoided by adding `rownames.print = FALSE` to the chunk header: i.e. `{r rownames.print = FALSE}` 

Alternatively, we could fix it globally (but opaquely) by adding a chunk options block to `include.R`:
```
knitr::opts_chunk$set(
  rownames.print = FALSE
)
```
### What potential "gotchas" do we know of?

I _think_ this only happens when printing with `head()`, so the individual fixes may be sufficient.

I don't know if there are any cases where we print data frames and definitively **do** want to print the row names, but I can certainly imagine there might be, which makes me wary of the global change.

### What are the recommended next steps?

Someone should do a quick pass through all rendered files to find places where we are printing row names and don't need to. 

### Is there a particular timeframe for this issue?

Probably before live?

jashapiro:
Tagging as part of #289

cansavvy:
> Alternatively, we could fix it globally (but opaquely) by adding a chunk options block to `include.R`:
> 
> ```
> knitr::opts_chunk$set(
>   rownames.print = FALSE
> )
> ```

I like this option so I will implement it unless anyone has objections.

jashapiro:
I think my main concern is that it will make things look different for people running locally, in a few cases. I don't think it is a major concern, and it is certainly easier to implement than on a block-by-block basis.

Also, I don't *think* we have any places where we print data frames where the gene or sample names are still in the rownames, but we might. Hiding the rownames in those places would be bad.

jaclyn-taroni:
Time to search all the html files for `head()` I think...

cansavvy:
It looks like almost every file has multiple occurrences of `head()` function. So I'll just take this one file at a time.

jaclyn-taroni:
[...]

Issue #415, opened 2020-12-05
The "Getting familiar with `clusterProfiler`'s options" section of ORA examples describes `msigdbr` 


jaclyn-taroni:
This section is written as if the code chunks pertain to the `clusterProfiler` package. For example, the section header and first sentence:

```
## Getting familiar with `clusterProfiler`'s options

Let's take a look at what organisms the package supports.
```

It should be focused on the `msigdbr` package and the Molecular Signatures Database (MSigDB) more generally. I can address this at the same time as #354 and #371.

jaclyn-taroni:
This was an issue for the GSEA and ORA notebooks, which I've since polished! Closing now.

PR #430, opened 2020-12-16
Polish the microarray ORA notebook

jaclyn-taroni:
### Purpose

Polishing the microarray ORA notebook.

### Issue addressed

For the microarray ORA example, the following are being addressed: 

- Enforcing comment character length limit (#406)
- Updating the section that said it was about `clusterProfiler` but was really about `msigdbr` (#415)
- Stopping row names from being printed when using `head()` (#428)

Not a gotcha, but I would like the reviewers to direct most of their attention to the section that is now called 

> Getting familiar with MSigDB gene sets available via `msigdbr`

I'll propagate these changes to the rest of the pathway analysis notebooks as I polish them.


cansavvy (approved):
This all looks good to me!

PR #431, opened 2020-12-16
Part 1: Add rownames.print = FALSE where its helpful

cansavvy:
### Purpose

Some of the printed tables have row names printed out even when its not helpful. 

This is the first PR of 2 that involves me going through each notebook: 
1) Searching for `head`
2) Looking at that table in the html 
3) Adding `rownames.print = FALSE` if the row names are not helpful there. 

Notebooks evaluated and addressed here: All microarray notebooks except the pathway analyses ones. 
Next PR will be all RNA-seq notebooks except pathway analyses. 

### Issue addressed

First part of addressing #428 

- If a notebook has no changes, but was included in the list of the notebooks I looked at, that means I concluded that the row names were helpful or at lease not confusing to include. 

- Apparently design matrices don't listen to `rownames.print = FALSE`?? So I deleted those. They don't seem too bad anyway. 

- <del> I have not added the html versions of the re-rendered notebooks for ease of reviewing, but I have checked them and they look better. </del> I added them now since its probably not easy to review these otherwise I have realized.


jashapiro:
> * Apparently design matrices don't listen to `rownames.print = FALSE`?? So I deleted those. They don't seem too bad anyway.

That's because they are matrixes... Only data frames are affected (and only those that are printed as paged tables).

cansavvy:
Here's the html preview links of the notebooks that are changed. Probably the best way to review these: 

http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames/02-microarray/differential-expression_microarray_01_2-groups.html
http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames/02-microarray/differential-expression_microarray_02_several-groups.html
http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames/02-microarray/gene-id-annotation_microarray_01_ensembl.html

02-microarray/differential-expression_microarray_01_2-groups.html
```diff
@@ -3929,7 +3929,7 @@ <h2><span class="header-section-number">4.2</span> Import and set up data</h2>
 <div class="sourceCode" id="cb10"><pre class="sourceCode r"><code class="sourceCode r"><span id="cb10-1"><a href="#cb10-1"></a><span class="co"># Read in metadata TSV file</span></span>
 <span id="cb10-2"><a href="#cb10-2"></a>metadata &lt;-<span class="st"> </span>readr<span class="op">::</span><span class="kw">read_tsv</span>(metadata_file)</span></code></pre></div>
 <pre><code>## 
-## ── Column specification ──────────────────────────────────────────────
+## ── Column specification ──────────────────────────────────────────────────────────────────────────────────
```
jashapiro, line 3932: Why did it do this? It's a mystery, but should not affect final rendering.

jashapiro (approved):
Looks good! I am a bit surprised this showed up in so few places, but if it rears its head again elsewhere, we can catch it then.

02-microarray/differential-expression_microarray_01_2-groups.html
cansavvy: This is a mystery. I ran it with snakemake in the docker image, so I find this perplexing.

PR #433, opened 2020-12-16
Part 2: Add the rownames.print = FALSE and re-render

cansavvy:
### Purpose

Some of the printed tables have row names printed out even when its not helpful. 

This is the second PR of 2 that involves me going through each notebook: 
1) Searching for `head`
2) Looking at that table in the html 
3) Adding `rownames.print = FALSE` if the row names are not helpful there. 

In this PR I evaluated all the RNA-seq notebooks except pathway analyses and also the only current advanced-topics notebook we have: wgcna. 

The pathway-analysis notebooks @jaclyn-taroni is addressing in this respect as she is going through them for polishing. 

### Issue addressed

This #428 

If a notebook has no changes, but was included in the list of the notebooks I looked at, that means I concluded that the row names were helpful or at lease not confusing to include. 

### Html previews of the changed notebooks: 

- http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames-part2/03-rnaseq/differential-expression_rnaseq_01.html
- http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames-part2/03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.html
- http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames-part2/03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.html
- http://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/cansavvy/remove-rownames-part2/04-advanced-topics/network-analysis_rnaseq_01_wgcna.html


cansavvy:
This is a reminder that [these renderings mysteriously aren't listening to the line width](https://github.com/AlexsLemonade/refinebio-examples/pull/431#discussion_r544529121), so ignore those changes, but the htmls that make it into `gh-pages-stages` should be fine since github pages will re-render.

jashapiro (approved):
🚀 Looks good!

PR #434, opened 2020-12-16
Polish Microarray GSEA example

jaclyn-taroni:
### Purpose

Polishing microarray GSEA example.

### Issue addressed

For the microarray GSEA example, the following are being addressed:

- Enforcing comment character length limit (#406)
- Updating the section that said it was about `clusterProfiler` but was really about `msigdbr` (#415)
- Stopping row names from being printed for paged tables (#428)
- Adding the introductory paragraphs that are in every pathway analysis notebook (#354)

There is also a good amount of editing happening in this pull request in the spirit of #290. 

### Directions for the reviewers

I would recommend that the reviewers look at the whole document when reviewing, rather that just the diff, to assess the level of detail and description of methods in the writing.

One concern I have is that the "main" description of how GSEA works is under `Purpose of this analysis`. Ideally it would be _closer_ to the `GSEA()` business itself. _However_, I think we need some explanation of the method before we get down to removing duplicate identifiers. An idea: We could move the description of GSEA that's under `Purpose` to under the `Perform gene set enrichment analysis (GSEA)` header and then move the removal of duplicates (and the explanation) to under the `Determine our pre-ranked genes list` section.

#### Specific questions

- @cbethell - We set the seed in the RNA-seq example, but we didn't here. I added a `set.seed()` step here, right before we "need" it should someone be viewing this example and have jumped to the analysis code. Is this the correct thing to do?
- @jashapiro - you filed #410 and have looked at more examples than me recently! Is this an appropriate level of detail/description in your opinion?

PR #435, opened 2020-12-17
Polish Ensembl Gene ID conversions (bonus reference updates!)

jashapiro:
### Purpose

Polishing Ensembl gene id conversions... 
Plus some reference updates

### Issue addressed

General content and formatting related to #290 and #289. These changes are primarily in `02-microarray/gene-id-annotation_microarray_01_ensembl.Rmd` and `03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.Rmd`.

Rendered versions of these files can be viewed at: 

https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/jashapiro/polish-ensembl/02-microarray/gene-id-annotation_microarray_01_ensembl.html
and
https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/jashapiro/polish-ensembl/03-rnaseq/gene-id-annotation_rnaseq_01_ensembl.html

The biggest changes are a bit of extra explanation of where the `org.` packages come from, and their relationship to AnnotationDbi. Some of that was in the notebook, but not right where I thought the reader might want it. I also added explicit notes about where the `mapIds()` function comes from. 

Since I had to cite AnnotationDbi, I ended up getting a bit into citation updating, which makes this PR spread a bit wider than it probably ought to. See below about the trouble I caused myself.

I also removed all occurrences of `Refine.bio` and replaced them with `refine.bio`

I noticed that one of the references was labeled as `Igor2020`, but the author's last name is Dolgalev, so I updated it to `Dolgalev2020` and moved the reference. However, this will likely create conflicts with #434.

Similarly, I updated the AnnotationDbi reference to `Pages2020-package` (from `Carlson2020-package`) to reflect the current citation. 

I fully expect that I will have some conflicts to resolve before this merges, after #434 goes in.
This is a problem of my own making... I will be responsible for fixing it.


jashapiro:
I am retargeting this to #434 (as a kind of stacked PR) to allow me to work on conflicts. I am assuming based on past experience than when #434 is merged it will retarget to staging. Also changing to draft so it doesn't accidentally get merged...

jashapiro:
Fewer conflicts than expected! Still waiting in case others develop...

02-microarray/gene-id-annotation_microarray_01_ensembl.Rmd
```diff
@@ -283,27 +289,26 @@ head(mapped_df)
...
 
 We can see that our data frame has a new column `Symbol`.
-Let's get a summary of the gene symbols returned in the `Symbol` column of our mapped data frame.
+Let's get a summary of the gene symbols in the `Symbol` column of our mapped data frame.
 
 ```{r}
-# We can use the `summary()` function to get a better idea of the distribution of symbols in the `Symbol` column
-summary(mapped_df$Symbol)
+# Use the `summary()` function to show the distribution of Symbol values
+# We need to use `as.factor()` here to get the counts of unique values
```
cansavvy, line 296: Probably should add a short comment explaining the `maxsum` bit.

cansavvy (approved):
These all seem like good changes to me! 

> I noticed that one of the references was labeled as Igor2020, but the author's last name is Dolgalev, so I updated it to Dolgalev2020 and moved the reference. However, this will likely create conflicts with #434.
Similarly, I updated the AnnotationDbi reference to Pages2020-package (from Carlson2020-package) to reflect the current citation. 

I didn't check all the other notebooks that possibly use this reference, but is this something you search and replaced? This is the only thing I haven't checked up on.

02-microarray/gene-id-annotation_microarray_01_ensembl.Rmd
jashapiro: Done!

PR #436, opened 2020-12-17
Polishing Ortholog notebooks

jashapiro:
### Purpose

Polishing updates related to #289 and #290.

I am filing this as a stacked PR, so it expects #435 to go in first.

### Issue addressed

Most of the changes are small. I made a few changes to remove extraneous `group_by` steps, and added a bit of context that I thought would help in a few places. 

In updating an experiment description, I made some changes that I also wanted to transfer to other notebooks, so those are in here as well. 

The primary changes are to 02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd and 03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd

Rendered versions of thes files can be viewed at: 
https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/polish-orthologs/02-microarray/ortholog-mapping_microarray_01_ensembl.html
and 
https://htmlpreview.github.io/?https://github.com/AlexsLemonade/refinebio-examples/blob/polish-orthologs/03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.html

In the zebrafish example, I noticed that there are some places where the same zebrafish/human mapping is listed twice with different "support". I suspect that this is because the "symbol" is not distinct, and there may be more than one human ensembl ID with the same symbol. I did not dive into it too deeply, except to add some extra filtering steps to catch those duplicates when tabulating. 

There are a few places in the text where we refer to counts (rows in a table etc.), but these are dependent on the downloaded file, which seems to have changed since the last time those numbers were updated!

There are a few ways we could address this: one is by using inline R functions to keep the text up to date. The other is to explicitly note some of these numbers may not match, depending on the version of the file. I guess one other is to do both, so the numbers in the rendered notebook are always correct, but if people are following along at home they might get different numbers.


jaclyn-taroni:
Do we want to add `polish-orthologs` to the render workflow to see if https://github.com/AlexsLemonade/refinebio-examples/pull/436/commits/4617b73a69dfc155a3da4604d8ab9b4c601576d5 closes #439? 🔬

03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
```diff
@@ -305,49 +305,50 @@ We don't want to handle duplicate data, so let's remove those duplicates before
...
 ```
 
-Looks like we have mapped IDs!
+There are certainly a good number of places where we mapped multiple zebrafish Ensembl IDs to the same human symbol!
+We'll look at this in a bit.
 
-Now, let's get an idea of how many mouse Ensembl IDs we have that were not mapped to human Ensembl IDs.
+We can also see that there 19,126 mouse Ensembl IDs that did not map to a human Ensembl ID.
+These are the rows with a value of NA.
+This seems like a lot, but many of these may not be "real" genes, and we  do not expect everything to map neatly across species.
```
cansavvy, line 350:
```suggestion
This seems like a lot, but many of these may not be "real" genes, and we do not expect everything to map neatly across species.
```

cansavvy: What's a "real" gene anyway? I there a different way we can phrase this to be more specific/informative?

03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
```diff
@@ -305,49 +305,50 @@ We don't want to handle duplicate data, so let's remove those duplicates before
...
 
 ```{r}
-# We can use this `count()` function after `group_by()`to get a count of how many
+# We can use this `count()` function to get a tally of how many
 # `mouse_ensembl` IDs there are per `human_ensembl` IDs
 human_mouse_mapped_df %>%
-  dplyr::group_by(human_ensembl) %>%
-  dplyr::count() %>%
-  # Sort by highest `n` which would be the human Ensembl ID with the most mapped
+  # count the number of rows per human gene
```
cansavvy, line 338:
```suggestion
  # Count the number of rows per human gene
```
I'm just being silly at this point.

02-microarray/ortholog-mapping_microarray_01_ensembl.Rmd
```diff
@@ -395,43 +412,44 @@ Since multiple zebrafish Ensembl gene IDs map to the same human symbol, we may w
...
+As we noted earlier, an orthology prediction where more than one of the databases concur would be considered more reliable.
+Therefore, where we have multi-mapped zebrafish Ensembl gene IDs, we will retain the mappings with more than one database to support the assertion.
 
 Before we do, let's take a look how many multi-mapped genes there are in the data frame.
 
 ```{r}
 human_zebrafish_mapped_df %>%
-  # Group by human gene symbols
-  dplyr::group_by(human_symbol) %>%
+  # Remove the support column
```
cansavvy, line 434:
```suggestion
  # Remove the `support` column
```

cansavvy: Most places you've used backticks.

cansavvy (approved):
This looks good to me. My comments are so teeny tiny; it's silly.

03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
```diff
@@ -305,49 +305,50 @@ We don't want to handle duplicate data, so let's remove those duplicates before
...
+  dplyr::count(human_ensembl) %>%
+  # Sort by highest `n` which will be the human Ensembl ID with the most mapped
   # mouse Ensembl IDs
   dplyr::arrange(desc(n))
 ```
 
-Looks like we have mapped IDs!
+There are certainly a good number of places where we mapped multiple zebrafish Ensembl IDs to the same human symbol!
```
jashapiro, line 345:
```suggestion
There are certainly a good number of places where we mapped multiple mouse Ensembl IDs to the same human symbol!
```

03-rnaseq/ortholog-mapping_rnaseq_01_ensembl.Rmd
jashapiro: This is not a silly comment! I'm trying to convey that many of the things we annotate are not all that well supported, so the fact that we are missing 19K genes is more likely the result of spurious annotations than error. I think having a reference for this claim might be useful. I'll look for one.

cansavvy: I forgot I had one un-silly comment.

PR #437, opened 2020-12-18
Polish RNA-seq GSEA example

jaclyn-taroni:
### Purpose

Polishing the RNA-seq GSEA example, building off of #434 

### Issue addressed

For the RNA-seq GSEA example, the following are being addressed:

- Enforcing comment character length limit (#406)
- Updating the section that said it was about `clusterProfiler` but was really about `msigdbr` (#415)
- Stopping row names from being printed for paged tables (#428)
- Adding the introductory paragraphs that are in every pathway analysis notebook (#354)

Also a fair bit of editing, but that's largely the same as what was in #434.


components/references.bib
```diff
@@ -180,6 +180,13 @@ @manual{gene-id-annotation-rna-seq
   url    = {https://github.com/AlexsLemonade/refinebio-examples/blob/master/03-rnaseq/gene-id-convert_rnaseq_01_ensembl.html}
 }
 
+@website{gene-set-testing-rnaseq,
+    author = {Stephane Ballereau and Mark Dunning and Abbi Edwards and Oscar Rueda and Ashley Sawle},
+    title = {{RNA-seq analysis in R: Gene Set Testing for RNA-seq}},
```
jashapiro, line 185: Let the algorithm handle capitalization for most things, (keep override for RNA-seq and R)
```suggestion
    title = {{RNA-seq} analysis in {R}: Gene Set Testing for {RNA-seq}},
```

jashapiro (approved):
Looks good (though I admit I didn't read too carefully, as it did seem that most of these changes were familiar from the last set.)

One bibliography fix, because I am oddly stringent on that.

components/references.bib
jaclyn-taroni: Might be a good item for CONTRIBUTING.md (because I looked at it for this info)

jashapiro: It would be. Bibtex is strange though... it may be hard to summarize the rules.

jaclyn-taroni: I don't think we're going to successfully summarize all the rules, just the ones that come up in review.

PR #438, opened 2020-12-19
GHA: Slack us if Docker build or rendering fails

### LINKED ISSUE #315: If Github actions rendering fails, we should know about it
Related to: 
https://github.com/AlexsLemonade/refinebio-examples/issues/312#issuecomment-713026482

Github actions rendering with `build-all` failed three times without us knowing about it because of a weird docker build fail that didn't really stop the docker building. 

We don't want fails to happen quietly, so we need to have system that sends us an alert if the render fails. Spell check sends alerts when it fails but perhaps this is because its required? - I'm not sure how this works, need to look into it. 

### Proposed solution(s)

I saw an example for [sending emails](https://github.com/spandanpal22/sending-alert-github-actions/blob/master/.github/workflows/send.yml) but they seemed to required a lot of `SECRETS` and set up. 

A slack notification might be the way to go? But what channel?
https://github.com/marketplace/actions/slack-notify

jaclyn-taroni:
#### Purpose

Sometimes rendering or pushing to Dockerhub fails and we don't know about it. This adds steps to those two workflows per the instructions posted: https://github.com/AlexsLemonade/refinebio-examples/issues/315#issuecomment-717914841

#### Issue addressed

I'm going to say that this closes #315 but of course this is the kind of thing I couldn't test ahead of time 🤞 

#### Notes for reviewers

- How would we like to test this? Intentionally add something that won't build to the Dockerfile?
- The action we're using looks a-okay to me, but another pair of 👀  would be helpful (this is tied to the commit for the tag I'm using): https://github.com/ravsamhq/notify-slack-action/tree/d91b0d897737030128964a7d274248dda13a6e3f

### DIFF
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -4,7 +4,7 @@ name: Build, Render, and Push
 # events only for the master branch
 on:
   push:
-    branches: [ staging, master ]
+    branches: [  staging, master, jaclyn-taroni/render-fail-loudly ]
 
 jobs:
   # This workflow contains a single job called "build-all"
--- a/.github/workflows/docker-build-push.yml
+++ b/.github/workflows/docker-build-push.yml
@@ -4,7 +4,7 @@ name: Build, Render, and Push
 # events only for the master branch
 on:
   push:
-    branches: [  staging, master, jaclyn-taroni/render-fail-loudly ]
+    branches: [ staging, master ]
 
 jobs:
   # This workflow contains a single job called "build-all"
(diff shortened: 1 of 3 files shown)


jashapiro:
This looks fine to me. Will it report the step that failed? Probably not, but easy enough to go look, which we would want anyway. 

Here is how I would test it: 

Add `jaclyn-taroni/render-fail-loudly` to the branch list:
```
on:
  push:
    branches: [ staging, master, jaclyn-taroni/render-fail-loudly ]
```

This will make it run with every push to your current branch. 

Then create something that makes things fail. I suggest typos as the simplest option, but trying to load a missing library is another good option (an error we don't explicitly deal with).

If I see an error in Slack, and if I then see the error-inducing commit reverted, I will approve.

jashapiro (approved):
Emoji decisions aside, this seemed to work! 🚀

PR #450, opened 2021-01-21
Add a template for user reports or questions

jaclyn-taroni:
I am working on https://github.com/AlexsLemonade/refinebio-docs/issues/138 and I wanted to include a link to file an issue on GitHub as the preferred way to report a problem or ask a question. Now that we have different issue templates for our own usage, I thought it made sense to create one specifically for _users_ of refine.bio examples rather than those of us who do development!


cansavvy (approved):
Seems good to me! :+1:

Issue #376, opened 2020-11-23
Navbar doesn't play well with sorta narrow windows 


jashapiro:
The navbar we have doesn't play super well on narrow displays; it goes to two lines at some points, overlapping and cutting off content and the TOC:

![](https://user-images.githubusercontent.com/20483/96883000-e3187400-144d-11eb-9659-e5422c441405.png)

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/refinebio-examples/issues/323#issuecomment-714562044_

For reference, the normal navbar currently looks like this: 

<img width="1256" alt="Screen Shot 2020-11-23 at 5 09 29 PM" src="https://user-images.githubusercontent.com/20483/100020933-a8885b00-2dae-11eb-985c-7685d256cb8a.png">

In a very narrow window we get a "mobile" layout:

<img width="500" alt="Screen Shot 2020-11-23 at 5 10 08 PM" src="https://user-images.githubusercontent.com/20483/100020997-bf2eb200-2dae-11eb-8fd1-e28aa5f2b44a.png">

But we fail in between...

This will probably require a look from the dev team to help us sort out, and as such, it is likely to be low priority for now.

Issue #328, opened 2020-10-22
Footer.html can use more info and aesthetics help


cansavvy:
From #307 , this is where the footer is left off. It is an `footer.html` file that is automatically incorporated into all html files upon rendering with `render-notebook.R`. 

But it doesn't currently have much information in it and could probably use an aesthetics boost. 
However, how to do this is beyond my knowledge. We also may want to have further discussion about what information should be included. 

> This is what this looks like right now:
> 
> <img alt="Screen Shot 2020-10-22 at 11 10 39 AM" width="1215" src="https://user-images.githubusercontent.com/23458084/96892063-41962000-1457-11eb-9021-76625ae421aa.png">
> 
> Would it look better if it extended across the whole bottom? Also, do we want some of this information that's in refine.bio's footer?
> 
> <img alt="Screen Shot 2020-10-22 at 11 13 18 AM" width="1215" src="https://user-images.githubusercontent.com/23458084/96892453-a8b3d480-1457-11eb-9956-0a5b254f22e8.png">

Issue #323, opened 2020-10-22
Fix/add refine.bio logo to navbar


cansavvy:
@dvenprasad did some work with the CSS file on #303  and navbar to get refinebio-examples to better resemble refine.bio (#278). 

The last thing left is the logo wasn't able to be incorporated without misbehaving: 

From @dvenprasad 's comment on #303: https://github.com/AlexsLemonade/refinebio-examples/pull/303#issuecomment-714516285

> The only thing I wasn't able to do was get the logo to sit properly in the navbar. It cut out. And when I tried to make the navbar accommodate it, it just overlapped with the content, instead of pushing it down. So for now I've replaced the logo with text until someone else can take a pass at it.
> 
> <img alt="Screen Shot 2020-10-22 at 9 25 05 AM" width="1202" src="https://user-images.githubusercontent.com/15315514/96882517-62597800-144d-11eb-84a0-bc119e3dfcfc.png">

This is something we can revisit later. Perhaps have someone from the dev team take a look at it.

jashapiro:
I put some ideas for solution here: https://github.com/AlexsLemonade/refinebio-examples/pull/303#issuecomment-714528242

jashapiro:
The comment above also notes that there is a rendering glitch when the menu gets small but not too small: 

![](https://user-images.githubusercontent.com/20483/96883000-e3187400-144d-11eb-9659-e5422c441405.png)

It would be good to solve that related problem with this one, if possible.

PR #454, opened 2021-03-12
Header and Footer Updates

samuel-tate:
### Purpose

Aesthetic updates to navbar and footer!

<details>
<summary>screenshot!!</summary>
<p>

image

</p>
</details>

### Issue addressed

#323 
#328 

I'm only pushing the changes I made to `_navbar.html`/`footer.html`/`styles.css`

If you want to see the changes rendered to the actual html files you'll need to rebuild them manually.

I've been doing this by deleting `./01-getting-started/getting-started.html` then running the command:
```
$ docker run --mount type=bind,target=/home/rstudio,source=$PWD --rm ccdl/refinebio-examples snakemake --cores 4
```

You can then open getting-started.html in a browser to see the changes rendered.

### navbar concerns

It seems like the navbar is "on top" of everything - it's not actually like on the same level as all of the other html elements.

This means that the navbar can cover content on the page (most notably when it gets resized to 2 lines see #376).

Since I increased the height of the navbar to 60px to accommodate the logo image, it covers content when you use the TOC to navigate to a section.

<details>
<summary>example</summary>
<p>

image

</p>
</details>

It seems body has a custom style that sets `padding-top` to 51px to account for the navbar - I guess the navbar is only expected to have a height of 50px?

I'm thinking that we might be able to add custom js to dynamically set `body.padding-top = navbar.height + 1`.
This would also take care of #376 
LMK what you think.

### footer concerns

Right now I  have the footer all the way at the bottom of the screen - outside of `container-main` - so that the gradient can extend the whole length of the screen.

In order to do this, I added some `</div>`s above the footer's code in `footer.html` to get it out of the main container.
This seems kinda silly/hacky so lmk if there's a better way to do this.

Also, whatever TOC generator we are using is using js to add an extra div of a calculated height at the bottom of the main container so that when you use the TOC to navigate to a section, that section's title is at the top of the screen.

<details>
<summary>the div I'm talking about</summary>
<p>

image

</p>
</details>

This div is making the footer be placed way at the bottom of the screen.
I was thinking that we could use custom js to subtract the height of the footer when it calculates the height for that div.

We should also talk about what info we want to put into the footer!

### DIFF
--- a/components/_navbar.html
+++ b/components/_navbar.html
@@ -10,8 +10,7 @@
       </button>
       <div class = "logo">
         <a class="navbar-brand" href="#">
-          <!-- <img src = "../components/refine-bio-ex-logo.svg" alt ="refine.bio examples"> -->
-          refine.bio examples
+          <img src = "../components/refine-bio-ex-logo.svg" alt ="refine.bio examples" height="50">
         </a>
       </div>
 
--- a/components/footer.html
+++ b/components/footer.html
@@ -1,5 +1,14 @@
-<ul class="nav navbar-nav navbar-right">
-  <li><a href="https://share.hsforms.com/1pQ8pPB70TuG37sVjr-KtXA336z0">Give us feedback</a></li>
-  <li><a href="https://github.com/AlexsLemonade/refinebio-examples">GitHub</a></li>
-  <li><a href="https://www.refine.bio">refine.bio</a></li>
-</ul>
+<footer class="footer">
+  <div class="container">
+    <div class="col-xs-12 col-sm-4 col-md-3"></div>
+    <div class="footer-content col-xs-12 col-sm-8 col-md-9">
+      <p>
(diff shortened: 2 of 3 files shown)


jashapiro:
> It seems body has a custom style that sets `padding-top` to 51px to account for the navbar - I guess the navbar is only expected to have a height of 50px?
> 
> I'm thinking that we might be able to add custom js to dynamically set `body.padding-top = navbar.height + 1`.
> This would also take care of #376
> LMK what you think.

This is just to note that this fix should also be applied somehow to the hacky thing I did here to make a hidden anchor with some padding above it so links to it go to the right place. Note that this line appears in every notebook, not just the template that I am linking to directly. I think all the headers may have the same padding/margin applied by whatever the base style sheet is.

https://github.com/AlexsLemonade/refinebio-examples/blob/4a9cf238f1cab474f3347a906139ecfeedab1f1b/template/template_example.Rmd#L152

davidsmejia:
These changes are incorporated into https://github.com/AlexsLemonade/refinebio-examples/pull/457

Issue #452, opened 2021-03-04
Create Homepage for Refinebio Examples


jaclyn-taroni:
When @dvenprasad performed usability testing, she mentioned that the examples needed a home page. This is a placeholder issue intended to track that addition; it may be edited to be more detailed or it might be better to break things up into multiple issues.

Because there are some examples-related frontend issues in this dev sprint, we may want to get more details down this sprint or the next.

I'm going to assign @dvenprasad for the moment.

dvenprasad:
Copy: 
> Explore a variety of example analyses like clustering, differential expression, pathway analysis, etc., for [refine.bio](http://refine.bio) datasets.
> 
> How to use these examples
> 
> These examples are available as R notebooks (.Rmd) and can be modified for your [refine.bio](http://refine.bio) dataset.
> 
> - Set up: The examples are written in R and use a variety of libraries. Get started by installing the required tools and libraries.
> - Download and follow along an example analyses: You can use the refine.bio dataset that we use in analysis or you can use a different refine.bio dataset, relevant to your research interests.
> 
> 
> **RNA-seq Examples** 
> 
> Clustering
> 
> Diff expression - 
> 
> Dimension reduction
> 
> Pathways analysis
> 
> **Microarray Examples**
> 
> Clustering
> 
> Diff expression - 
> 
> Dimension reduction
> 
> Pathways analysis

@cansavvy for pathway, dimension reduction, and differential expression(microarray) which of the examples should we highlight? 

Also, here's a rough layout for the page. We can tweak it to whatever is feasible with the platform we are using. I have just used random plot images, we should replace them with appropriate ones (kudos to @cansavvy for showing the plots idea) 

![rbio-examples-homepage](https://user-images.githubusercontent.com/15315514/110388642-60c1d280-8031-11eb-957b-10b969ae8ffc.png)

cansavvy:
I think the ones you mentioned here are fine: https://github.com/AlexsLemonade/refinebio-frontend/issues/926#issuecomment-790870534

We can get you the png’s of those plots so they can be dropped right in there. We also may want some blurbs underneath each of those example titles so people know what they are getting into. (I can write those when you are ready).

jashapiro:
Probably better not to have the main image be a volcano plot with no significant results, lest we give people the sads.

cansavvy:
> Probably better not to have the main image be a volcano plot with no significant results, lest we give people the sads.

Oh yes 100.

[...]

PR #457, opened 2021-04-23
452 Fix Header + Footer + Add Homepage

davidsmejia:
### Purpose

* Add homepage to the examples site
* fix scaling of header/navbar
* adds footer/fix layout
* adds documentation for the homepage
* rerenders all the things
* fixes the fonts ~by adding the font file to be downloaded and renaming `google-analytics.html` -> `header-includes.html`~

### Issue addressed

#452 

* Imported the vendor dependencies from a cdn which I would consider "production ready".
* ~The homepage has to be rendered and committed since `snakemake --forceall` won't trigger the rule for the homepage.~

* ~Navbar couldnt be dynamically included - if it were the relative links would have to be updated to go from root (which would break local dev unless you spun up a server)~

Screen Shot 2021-07-02 at 9 38 54 AM
Screen Shot 2021-07-02 at 9 39 12 AM


davidsmejia:
Add link to refine.bio in how to use section

jashapiro:
> Navbar couldnt be dynamically included - if it were the relative links would have to be updated to go from root (which would break local dev unless you spun up a server)

I get this, but I worry that having two navbars we have to update down the line will be confusing. Can we somehow put `index.html` at `homepage/index.html` or something like that with a redirect? My other thought was some kind of js to rewrite the links, but that seems even more hackish?

jashapiro (commented):
Since this keeps popping up in my slack reminders, I am submitting a new comment as a "review". 

Overall, this looks great, but having two separate header locations to keep track of and keep updated strikes me as a problem we should solve. I had an idea in my earlier comment https://github.com/AlexsLemonade/refinebio-examples/pull/457#issuecomment-850465068, but I don't know how workable it is...

My other question is if we want to incorporate #454 into this PR (or merge that first?)

davidsmejia:
* add footer to homepage
* make the footer wider and at the bottom of the doc
* add github action to inject the header on deploy
* align the set up and other section on the homepage

davidsmejia:
@jashapiro I think I have addressed any feedback that you had but do you think its alright to have the homepage rendered on commit instead of the github action. I wasn't sure if they should be treated the same... also it didnt work out of the box.

scripts/render-homepage.R
```diff
@@ -0,0 +1,36 @@
...
+footer_index <- which(grepl(footer_comment, home_lines))
+footer_file <- normalizePath(file.path("components", "footer.html"))
+footer_lines <- readr::read_lines(footer_file)
+home_lines <- append(home_lines, footer_lines, footer_index)
+
+# replace any relative sources
+# Rmarkdown looks for an image locally (inside /components)
+# the homepage looks for the file on the server (inside /refinebio-examples/components)
+home_lines <- sub('href="../', 'href="/refinebio-examples/', home_lines)
+home_lines <- sub('src="../', 'src="/refinebio-examples/', home_lines)
```
jashapiro, line 33: Is there a reason not to keep these as relative links? Would this eliminate the need for a local server? 

```suggestion
home_lines <- sub('href="../', 'href="', home_lines)
home_lines <- sub('src="../', 'src="', home_lines)
```

Update: I tried it, and it seemed to mostly work, except for the link back to the home page, which I think would work if `index.html` were explicit in that link?

UPDATING_HOMEPAGE.md
```diff
@@ -0,0 +1,23 @@
...
+```shell
+$ docker run --mount type=bind,target=/home/rstudio,source=$PWD ccdl/refinebio-examples snakemake render_homepage
+```
+
+### Local Development
+
+Launching a simple server is helpful for previewing changes. In order to mirror how github-pages works it is recommended to run the server from the parent directory of the project.
+```shell
+python -m SimpleHTTPServer 8000
+```
```
jashapiro, line 23: Took me a second to get this to work since I changed my user default python...
~~~suggestion
```shell
python -m SimpleHTTPServer 8000
```
or
```shell
python3 -m http.server 8000
```
~~~

jashapiro (commented):
This looks good as an approach to getting thing inserted semi-dynamically. I'm fine with making the chances on commit rather than trying to set up actions to do it. Hopefully we are not updating all of this too often. 

I have a few comments about some `snakemake` changes that I think could make `--forceall` option work, but I'm not sure whether or not you tried that already. One other possibility is to make the navbar and footer files inputs to the snakemake rules, which should force rendering of everything when those change even without `forceall`, but I'm less sold on that in general.

One thing I ran into and I have no idea what is going on is that the navbar isn't getting styled properly (font only, it looks like?) in Safari on a mac for the internal pages. Looks fine with Brave, but not right with Safari, and the home page looks fine in either.

Home page:
Screen Shot 2021-07-06 at 4 01 54 PM
Internal pages:
Screen Shot 2021-07-06 at 4 02 01 PM

UPDATING_HOMEPAGE.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Making Changes
+
+You can edit `components/_homepage.html` and save to apply your changes to the homepage.
+
+Editing `components/_navbar.html` or `components/footer.html` will be incorporated into all pages including the homepage
+
+### Rendering
+
+The rule for updating the homepage will not be applied when calling `snakemake --forceall` so you must render and commit changes. The github action does not render these out on merges.
```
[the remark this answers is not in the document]
davidsmejia: This did seem to work I think I was missing the output on the render_homepage rule

davidsmejia:
> This looks good as an approach to getting thing inserted semi-dynamically. I'm fine with making the chances on commit rather than trying to set up actions to do it. Hopefully we are not updating all of this too often.
> 
> I have a few comments about some `snakemake` changes that I think could make `--forceall` option work, but I'm not sure whether or not you tried that already. One other possibility is to make the navbar and footer files inputs to the snakemake rules, which should force rendering of everything when those change even without `forceall`, but I'm less sold on that in general.
> 
> One thing I ran into and I have no idea what is going on is that the navbar isn't getting styled properly (font only, it looks like?) in Safari on a mac for the internal pages. Looks fine with Brave, but not right with Safari, and the home page looks fine in either.
> 
> Home page:
> Screen Shot 2021-07-06 at 4 01 54 PM
> Internal pages:
> Screen Shot 2021-07-06 at 4 02 01 PM

Looks like the google font was only being included with the changes I made for the homepage.. If it had ever appeared correctly before it was because it was cached in your browser. Good catch I made some changes to fix that and include the style sheet on the notebooks!

jashapiro:
Kinda minor, but I noticed in `refinebio-examples/03-rnaseq/pathway-analysis_rnaseq_01_ora.html` that one of the image files was not loaded because the AWS object was not public. Since the rendered html file embeds images, (Rmarkdown), this failed load is now baked in to the current rendered files. Rerendering should fix it, and while I am pretty confident that should be the only file affected, we might want to think about how to check for any other cases.

jashapiro (approved):
This all looks good to me! There is the one note about rerendering for a broken image that I made earlier, but otherwise, pretty! https://github.com/AlexsLemonade/refinebio-examples/pull/457#issuecomment-880211607

Issue #449, opened 2021-01-14
GitHub Actions: Check URLs in refine.bio frontend and docs when merging to master


jaclyn-taroni:
### Background

Borne out of discussion of #285. Before we made significant changes to the content and structure of examples, we created a release – [`v1.0`](https://github.com/AlexsLemonade/refinebio-examples/releases/tag/v1.0) – so the links in our documentation and the frontend of refine.bio continued to work. When we link to a specific example, we use HTML Preview tied to that tag. If we link to a specific topic area (e.g., differential expression), we usually link to the _directory_ on GitHub for that release because we used to organize examples by topic rather than by technology (e.g., https://github.com/AlexsLemonade/refinebio-examples/tree/v1.0/differential-expression). 

My understanding of the rationale for #285 was that we thought we needed to continue to have releases for things to just work ™️ in places where we link to examples. **That is not the case; we can use the GH pages URL instead.**

### Problem

Discussion around what's above revealed a _different problem_: if we link to specific examples as we plan to (see: https://github.com/AlexsLemonade/refinebio-frontend/issues/926 and https://github.com/AlexsLemonade/refinebio-docs/issues/138), things will break if the notebook names change. It seems plausible that notebook names will change if we do things like #242. 

We need to have a way to ensure that any changes that we make here and push to `master` (publish to GH pages) do not break links on the frontend and in our documentation.

### What are the recommended next steps?

@davidsmejia suggested that we write a GitHub action to check the links on the frontend and in the docs when we tried to merge to master. We'd want this to _fail_ if, after the proposed changes, the frontend and docs links would 404. 

Then you couldn't publish the changes until you fixed the frontend and the docs.

### Is there a particular timeframe for this issue?

After we address the following, we'll want to have this working:

- https://github.com/AlexsLemonade/refinebio-frontend/issues/906
- https://github.com/AlexsLemonade/refinebio-frontend/issues/926
- https://github.com/AlexsLemonade/refinebio-docs/issues/138

PR #458, opened 2021-07-12
449 - add github action to check frontend urls

davidsmejia:
### Purpose

* adds an Rscript that takes a single argument ( destination of a json file that lists urls ) and checks if the files and section id exists
* adds a github action that calls the above script on PR's merging into the `master` branch with a url pointing to refinebio-frontend production list of refinebio-examples' urls

For any subsequent PRs, their checks will fail and present a list of the missing files and ids that correspond to any would be broken link. In order to pass you can request changes on the refinebio-frontend and rerun the `Check Refinebio Frontend Urls` action after the updates are deployed to get a passing result.

### Issue addressed

#449 

The order of PRs that should be merged would be
* https://github.com/AlexsLemonade/refinebio-examples/pull/457
* https://github.com/AlexsLemonade/refinebio-frontend/pull/945
* this PR

I think logistically an improvement would be for the action to run both on `master` and `staging` prs to match against refinebio-frontend `master` or `dev` branch so staging so they could both be deployed at the same time.

### DIFF
--- a/.github/workflows/check-frontend-urls.yml
+++ b/.github/workflows/check-frontend-urls.yml
@@ -1,0 +1,37 @@
+name: Check Refinebio Frontend Urls
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "style and check"
+  check-frontend-urls:
+    runs-on: ubuntu-latest
+    container:
+      image: rocker/tidyverse:4.0.2
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: Rscript -e "install.packages(c('optparse', 'jsonlite'))"
+
+      - name: Run Script to check URLs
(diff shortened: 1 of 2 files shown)


.github/workflows/check-frontend-urls.yml
```diff
@@ -0,0 +1,37 @@
...
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "style and check"
```
davidsmejia, line 11:
```suggestion
  # This workflow contains a single job called "check-frontend-urls"
```

jashapiro (approved):
This looks fine to me as is, but I made some `R/tidyverse` suggestions _while I was there_.  You don't need to take them, but I thought you might be interested in some of the alternatives.

Your comment about checking the staging branch too seems to make sense to me, but I'm trying to think it through (warning, stream of consciousness to follow) I guess I am not fully sure how the workflow would go though? We would check against `dev`, merge to `staging`, then check against the frontend `master` when merging to `master`? Do we need that level of care, or can we get away with just what stands here? It seems to me that this issue will come up relatively rarely, but I guess I'm not sure. Maybe this is sufficient for now, and we can cross the later bridge when it comes up?

PR #461, opened 2021-07-15
Bump R base version

jashapiro:
To solve the perennial `matrixStats`/`MatrixGenerics` version incompatibility problem I am bumping the rocker/tidyverse  base image version to R 4.0.5.

This seemed to result in a rendering problem, which my guess was because there was a change in when the css file argument for `rmarkdown::render` is evaluated. To alleviate that I define the full path outside the function call now, and things seem to render fine. At least that is the case locally. We will know for sure after this is merged, I guess.


davidsmejia (approved):
LGTM

PR #468, opened 2021-07-16
449 check client urls

davidsmejia:
Same as before, and the time before that but this time includes the url fix. Waiting on last pr to be reverted so this can be just one squashed commit into stage.

### Purpose

* adds an Rscript that takes a single argument ( destination of a json file that lists urls ) and checks if the files and section id exists
* adds a github action that calls the above script on PR's merging into the `master` branch with a url pointing to refinebio-frontend production list of refinebio-examples' urls

For any subsequent PRs, their checks will fail and present a list of the missing files and ids that correspond to any would be broken link. In order to pass you can request changes on the refinebio-frontend and rerun the `Check Refinebio Frontend Urls` action after the updates are deployed to get a passing result.

### Issue addressed

#449 

The order of PRs that should be merged would be
* https://github.com/AlexsLemonade/refinebio-examples/pull/457
* https://github.com/AlexsLemonade/refinebio-frontend/pull/945
* this PR

I think logistically an improvement would be for the action to run both on `master` and `staging` prs to match against refinebio-frontend `master` or `dev` branch so staging so they could both be deployed at the same time.


davidsmejia:
After the revert i will merge in staging and then  merge this

jashapiro (approved):
This looks fine to me, I just resolved the conflict, rather than revert and remerge.

Issue #473, opened 2022-02-22
User issue: Download for RNA-seq differential expression Rmd file is 404ing


jaclyn-taroni:
### Which example or page does this problem or question pertain to?

This is a user-reported (via an email to the requests address) related to the RNA-seq differential expression example.

### Report the problem you encountered or ask a question

The HTML version / GitHub pages is working: https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential-expression_rnaseq_01.html

The problem is that the link to download the file is 404ing: https://alexslemonade.github.io/refinebio-examples/03-rnaseq/differential_expression_rnaseq_01_rnaseq.Rmd

I wonder if this applies to multiple notebooks, though. That's the first thing we should look at.

jashapiro:
A quick investigation says it is not _all_ pages... I haven't yet found another that fails.

jashapiro:
closed by #476

PR #474, opened 2022-02-28
Fix link to Rmd file

jashapiro:
### Purpose

Fix a broken link on the differential expression example page.

### Issue addressed

#473 

I don't think there is anything, but we will need to publish these changes before closing the issue.

I checked the link, but reviewers should confirm that it works for them!


jaclyn-taroni (approved):
👍🏻  thank you!

PR #477, opened 2022-03-01
Fix WGCNA url

jashapiro:
In #474, I fixed a broken link, but it was not the one actually requested by the user... They were asking about the link on the WGCNA page, which was coincidentally (erroneously) the same as the DE page, and pointed to the same broken place. So now it should point to the correct place, and work.


jaclyn-taroni (approved):
Ah, okay - now I get it!

